PR #10, opened 2024-04-17
Categorize gitignore items

szepeviktor:
These are viktor-invented categories.

[_source_](https://github.com/szepeviktor/debian-server-tools/blob/master/.gitignore#L15)


tractorjuice:
I now rename all the metadata provided by FireCrawl by just sticking fc_ infront of all of them, so I now know they were generated by FireCrawl.

PR #29, opened 2024-04-19
Fixes pdfs not found if .pdf is not present

nickscamara:
Pdfs that don't include the .pdf at the end of the url are not picked up by the extractor. This method uses 2 ways to check it:
1. The url first, that way is fast. If it doesn't work fallback to looking at the content type.
2. Fetch the page and look at the content type

Bottlenecks: If there are many pages to crawl, we will fetching it multiple multiple times.

Solution for now:
1. Lets keep this in the /scrape endpoint but not crawl for now (DONE)
2. Bundle it with the crawling system that way it fetches only one. (Do another PR for it)


rafaelsideguide:
@nickscamara I'm not being able to add you as a reviewer for some reason. Got a rate limit 429 testing on preview routes.

nickscamara:
All readyy!

PR #34, opened 2024-04-20
Implements the ability for the crawler to output all the links it found, without scraping

### LINKED ISSUE #33: Ability for the crawler to output all the links it found, without scraping

nickscamara:
closes #33


nickscamara:
This modifies the return statement to return a list of urls instead of Document. Think it is okay for now as long as the docs are clear about that. We could create another endpoint but it can get confusing and harder to maintain.

nickscamara:
- [ ] Update the docs before merging

PR #40, opened 2024-04-21
WIP: Cjp/making db auth optional <> Running project locally

calebpeffer:
Started to decouple the database from the rest of the repo to make self hosting easier. Still WIP, looking for feedback @nickscamara @rafaelsideguide @ericciarla


nickscamara:
Great start!

nickscamara:
created a withAuth generic function wrapper that we can wrap supabase required functions with it, that way it is  cleaner

So for billTeam for example, all we do is
export async function billTeam(team_id: string, credits: number) {
  return withAuth(supaBillTeam)(team_id, credits);
}
export async function supaBillTeam(team_id: string, credits: number) {
// ... rest of the implementation

import { AuthResponse } from "../../src/types";

export function withAuth<T extends AuthResponse, U extends any[]>(
  originalFunction: (...args: U) => Promise<T>
) {
  return async function (...args: U): Promise<T> {
    if (process.env.USE_DB_AUTHENTICATION === "false") {
      console.warn("WARNING - You're bypassing authentication");
      return { success: true } as T;
    } else {
      try {
        return await originalFunction(...args);
      } catch (error) {
        console.error("Error in withAuth function: ", error);
        return { success: false, error: error.message } as T;
      }
    }
  };
}

calebpeffer:
I just added starting contributors with local setup instructions. @ericciarla can you test it for tutorial/setup errors since you haven't yet installed the repo?

nickscamara:
@calebpeffer updated the docs too - looking good!

nickscamara (approved):
lgtm

PR #56, opened 2024-04-23
[Feat:mvp] Search Endpoint => serp api + firecrawl => 🔥 🔍 

nickscamara:
MVP for the /search endpoint. 

Introduces a serp + firecrawl experience. Allows user to search a query and get the top n results in google. **It will also return the n page contents in markdown along with the page metadata**. It is optimized for speed. Currently search is around ~10-12s depending on the query with benchmarks of 2s for the initial search itself - very comparable to other search apis. Still lots of room for improvements  ( n defined by limit/default 7 )

POST v0/search

```
{
 query: "what is a coconut?",
  searchOptions:{
    limit: 5 // max 20 for now
  }
}
```

Pending for release:
- [ ] proxing
- [ ] rate limits
- [ ] answ box + kg
- [ ] llm

### DIFF
--- a/apps/api/.env.example
+++ b/apps/api/.env.example
@@ -22,3 +22,4 @@ BULL_AUTH_KEY= #
 LOGTAIL_KEY= # Use if you're configuring basic logging with logtail
 PLAYWRIGHT_MICROSERVICE_URL=  # set if you'd like to run a playwright fallback
 LLAMAPARSE_API_KEY= #Set if you have a llamaparse key you'd like to use to parse pdfs
+SERPER_API_KEY= #Set if you have a serper key you'd like to use as a search api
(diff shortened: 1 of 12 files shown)


nickscamara:
Added:

### Time-Based Search (tbs) Parameter

You can also refine search results based on time. The tbs parameter allows you to filter results by various time ranges. For instance, to get results from the past hour, use `qdr:h`, or for results from the past year, use `qdr:y`. Here's how you can use the tbs parameter:
```python
from googlesearch import search
search("Google", tbs="qdr:h")  # Results from the past hour
search("Google", tbs="qdr:m")  # Results from the past month
search("Google", tbs="qdr:y")  # Results from the past year
```

nickscamara:
Added:

The 'filter' parameter: It helps refine search results based on different attributes like images, videos, or news. Using 'filter=1' will display image results only.

nickscamara:
Added support for serper too

nickscamara (commented):
good

PR #68, opened 2024-04-25
[Bugfix] JS-SDK: Remove dotenv and add tests

mdp:
I might be misunderstanding something here, but I suspect that using 'dotenv' in an NPM module could result in some unexpected modification of the users environment variables. For example, upon importing FirecrawlApp, I would be surprised if it loaded my .env file into the environment. I think this should be a decision left to the end-user.

I also added tests after removing dotenv, but it's in a second commit, so feel free to ignore and cherry pick a7be09e if you don't want the jest tests.


rafaelsideguide (approved):
Looks good! Awesome work with the tests!

rafaelsideguide:
- [ ] publish npm package

nickscamara:
All good

Issue #101, opened 2024-05-02
[Feat] Stream crawl results


rafaelsideguide:
User is able to get partial results on /crawl/status route

PR #65, opened 2024-04-25
feat: add timeout param to scrape and crawl

ezhil56x:
/claim #59

PR #120, opened 2024-05-04
Refactor of main web scraper + Partial data streaming

nickscamara:


nickscamara:
closes #101 and partially #65. #65 extracted part of the partial data from there + cleanup

apps/api/src/controllers/search.ts
```diff
@@ -69,7 +70,7 @@ export async function searchHelper(
     },
   });
 
-  const docs = await a.getDocuments(true);
+  const docs = await a.getDocuments(false);
```
nickscamara, line 73: Turned the cache off for search

PR #166, opened 2024-05-20
Allow override of API URL

mattjoyce:
Allows python sdk to be used with local installs.

Add optional parameter for the class, and defers to environment variable FIRECRAWL_API_URL if set.


nickscamara:
Great! Thank you! @mattjoyce

PR #170, opened 2024-05-21
update: wait until body attached in playwright-service

qyou:
Get the page content until the body is attached instead of hard coded timeout.


nickscamara:
Thank you! Removed the body attachment for now due to playwright recommending against it in some cases.

PR #175, opened 2024-05-22
Test/load testing

rafaelsideguide:


rafaelsideguide:
Closes #163 
- [x] /scrape
- [ ] /crawl

rafaelsideguide:
We still need to figure out an autoscaling strategy for /crawl. A solution is to:
- Create a cron service that checks every minute the active jobs queue
  - If higher than a max_active_jobs number:
    - runs `fly restart` on a previously paused worker machine
  - If zero for 5 minutes and machines running > min_machines_running:
    - runs `fly stop` on a autoscaling worker machine

PR #178, opened 2024-05-22
Fix FIRECRAWL_API_URL bug, also various PyLint fixes

mattjoyce:
Relates to 
https://github.com/mendableai/firecrawl/issues/177


rafaelsideguide (commented):
lgtm. Awesome job @mattjoyce !

apps/python-sdk/firecrawl/firecrawl.py
```diff
@@ -41,7 +73,8 @@ def scrape_url(self, url: str, params: Optional[Dict[str, Any]] = None) -> Any:
         response = requests.post(
             f'{self.api_url}/v0/scrape',
             headers=headers,
-            json=scrape_params
+            json=scrape_params,
+            timeout=(5,10)
```
nickscamara: Might not want to hard code the timeout for scrape.

nickscamara: We can prob remove that as the function itself already has a default timeout of 15s

apps/python-sdk/firecrawl/firecrawl.py
```diff
@@ -66,22 +112,38 @@ def search(self, query, params=None):
         response = requests.post(
             f'{self.api_url}/v0/search',
             headers=headers,
-            json=json_data
+            json=json_data,
+            timeout=(5,10)
```
nickscamara: Same here

PR #181, opened 2024-05-24
[Feat] Added proxy and media blocking support for Playwright

JakobStadlhuber:
Updated environment variables and application settings to include proxy configurations and media blocking option. The proxy settings allow users to use a proxy service, while the media blocking is an optional feature that can help save bandwidth. Changes have been made in the .env.example, docker-compose.yaml, and main.py files.


rafaelsideguide:
lgtm! Awesome job @JakobStadlhuber !

PR #185, opened 2024-05-24
Improvements to the blocklist regex

nickscamara:
It was blocking websites such as wix.com sometimes due to the x.com check.

This check aims to fix that.

- [x] Add some tests around this


nickscamara:
@rafaelsideguide can you test this one when you get a chance?

PR #213, opened 2024-05-30
Script to check local vs published versions of sdk packages

mattjoyce:
@rafaelsideguide 

This pull request adds a new script to check if the version of a local package has been incremented compared to the published version on PyPI (for Python packages) or npm (for JavaScript packages). The script supports both Python and JavaScript packages and can be executed from the command line or as part of a github action.

**Script Overview:**
The script retrieves the current version of the package from the local filesystem.
For Python packages, it fetches the version from __init__.py.
For JavaScript packages, it fetches the version from package.json.
It then retrieves the latest published version from PyPI (for Python) or npm (for JavaScript).
The script compares the local version with the published version and prints true if the local version is greater (indicating an increment), otherwise it prints false.

**Usage:**

For Python Packages:

```
python check_version.py python /path/to/package package_name
```

For JavaScript Packages:

```
python check_version.py js /path/to/package package_name
```

### DIFF
--- a/.github/scripts/check_version_has_incremented.py
+++ b/.github/scripts/check_version_has_incremented.py
@@ -1,0 +1,88 @@
+"""
+checks local verions against published verions.
+
+Usage:
+
+python .github\scripts\check_version_has_incremented.py js .\apps\js-sdk\firecrawl @mendable/firecrawl-js 
+Local version: 0.0.22
+Published version: 0.0.21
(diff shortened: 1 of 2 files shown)


nickscamara:
@rafaelsideguide can you add this script check to the main ci?

rafaelsideguide (approved):

PR #229, opened 2024-06-03
Use @ instead of # for default BULL_AUTH_KEY. Hash mark is reserved for URI fragments.

rombru:
Using # in a URL doesn't always work.
It prevents the page "/admin/#/queues" from loading correctly when self hosting.
Hash mark is reserved for URI fragments.

Source: https://en.wikipedia.org/wiki/URI_fragment


nickscamara:
Thank you @rombru!

Issue #214, opened 2024-05-30
[Feat] Webhook when page is processed


JakobStadlhuber:
**Problem Description**
I find it better if its also possible to get notified when a scrape, crawl process of a page is finished.

**Proposed Feature**
a webhook that is fired everytime a page is completely processed and scraped in the api

**Alternatives Considered**
subscribe to redis

**Implementation Suggestions**
fire in the `convertUrlsToDocuments` the webhook

**Use Case**
an application can react to new scraped pages

JakobStadlhuber:
Should I take a look into it or do you have an existing flow/solution for subscribing to changes without pulling the status api?

nickscamara:
@JakobStadlhuber We have webhooks support for /crawl right now but not /scrape. Have you tried that? We also need to improve the docs around it. I think there is only an explanation of it in the playground get code if I am not mistaking.

JakobStadlhuber:
I see, thanks for mentioning. But I think its not configurable at the moment isn't it?

nickscamara:
Take a look under the account page on the dashboard. There should be a webhook url that you can set there to receive a ping when the crawl job is done. That would be awesome if you can improve the functionality further + have better docs/readme for it! Thank you!

JakobStadlhuber:
Thank you for the clarification. In the self-hosted version is no directly ENV Variable to configure the webhook, probably this [https://github.com/mendableai/firecrawl/blob/260e31c68b880595a76821373b42cb600c9d5dc7/apps/api/src/services/webhook.ts#L4](https://github.com/mendableai/firecrawl/blob/260e31c68b880595a76821373b42cb600c9d5dc7/apps/api/src/services/webhook.ts#L4 ) is the file were optionally ENV Variables can be passed.

nickscamara:
Yes, I believe webhooks are coupled with the database right now. Feel free to pr a decoupled version of it so we can get this working for self host!

JakobStadlhuber:
okay thanks for the info, I will see when I have time. Thank you :)

PR #234, opened 2024-06-04
Add support for Self-Hosted Webhook URL Usage and added project_id into the webhook payload

JakobStadlhuber:
This commit introduces the capability of using a Self-Hosted Webhook URL. The application now checks for a self-hosted URL before querying the database for the webhook settings. If a Self-Hosted Webhook URL is set in the environment variables, it will be used directly, diminishing unnecessary database queries.

### DIFF
--- a/apps/api/src/services/webhook.ts
+++ b/apps/api/src/services/webhook.ts
@@ -2,11 +2,15 @@ import { supabase_service } from "./supabase";
 
 export const callWebhook = async (teamId: string, data: any) => {
   try {
-  const { data: webhooksData, error } = await supabase_service
-    .from('webhooks')
-    .select('url')
-    .eq('team_id', teamId)
-    .limit(1);
+  const selfHostedUrl = process.env.SELF_HOSTED_WEBHOOK_URL;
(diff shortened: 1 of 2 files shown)


JakobStadlhuber:
#214

rafaelsideguide (approved):
lgtm

apps/api/src/services/webhook.ts
```diff
@@ -28,13 +35,14 @@ export const callWebhook = async (teamId: string, data: any) => {
...
 
-  await fetch(webhooksData[0].url, {
+  await fetch(webhookUrl, {
     method: 'POST',
     headers: {
       'Content-Type': 'application/json',
     },
     body: JSON.stringify({
       success: data.success,
+      project_id: data.project_id,
```
nickscamara, line 45: What is this project_id again?

nickscamara: Besides this question, everything else looks great

nickscamara: Nevermind

nickscamara:
lgtm!

PR #235, opened 2024-06-04
[Feat] Added custom scraping for google-drive pdf usecase

rafaelsideguide:
Closes #89

PR #238, opened 2024-06-04
[Feat] Improved the scrape for gdrive pdfs

rafaelsideguide:
able to get more content than before #235


rafaelsideguide:
don't know if it's able to get a screenshot from first scraping attempt

nickscamara:
@rafaelsideguide looking at the tests, seems to be breaking something else

PR #251, opened 2024-06-07
Adds a nolink attribute to scraper pageOptions

mattjoyce:
Removes HREF and src, keep titles and alts if provided.

This is my first time coding TS, but works for me.
It just adds a couple of turndown rules to replace the elements.

Here's the test I used.
```
url = "https://www.humblebundle.com"

app = FirecrawlApp(api_url="http://localhost:3002")

params={"pageOptions": {
        "includeHtml": False,
        "noLinks": True,
        "waitFor": 123,        
        "onlyMainContent": True,}}
scraped_data = app.scrape_url(url,params=params)
```

### DIFF
--- a/apps/api/src/lib/entities.ts
+++ b/apps/api/src/lib/entities.ts
@@ -13,6 +13,7 @@ export interface Progress {
 export type PageOptions = {
   onlyMainContent?: boolean;
   includeHtml?: boolean;
+  noLinks?: boolean;
   fallback?: boolean;
   fetchPageContent?: boolean;
   waitFor?: number;
--- a/apps/api/src/lib/html-to-markdown.ts
+++ b/apps/api/src/lib/html-to-markdown.ts
@@ -1,44 +1,72 @@
-
-export function parseMarkdown(html: string) {
-  var TurndownService = require("turndown");
-  var turndownPluginGfm = require('joplin-turndown-plugin-gfm')
-
+export function parseMarkdown(html: string, removeLinks: boolean = false) {
+  var TurndownService = require('turndown');
+  var turndownPluginGfm = require('joplin-turndown-plugin-gfm');
 
   const turndownService = new TurndownService();
-  turndownService.addRule("inlineLink", {
-    filter: function (node, options) {
-      return (
-        options.linkStyle === "inlined" &&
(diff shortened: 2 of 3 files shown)


nickscamara:
Will check this out tomorrow! Thanks @mattjoyce!

nickscamara:
I think it would be great to add some tests here @mattjoyce as well. @rafaelsideguide any thoughts here? we could prob do something similar like instead of a boolean no links have an array of tags to exclude. like excludeTags: `[img, a]` idk.

rafaelsideguide:
I'm leaning towards @nickscamara's suggestion about the `pageOptions.removeTags: string[]`. It seems like a way to give us more control over which HTML elements we want to exclude. But just to clarify, this isn't exactly what @mattjoyce implemented. His update specifically targets the `href` and `src` **attributes**, not the tags themselves.

```md
with `noLinks: false`
[Talk to us](https://calendly.com/d/cj83-ngq-knk/meet-firecrawl)

with `noLinks: true`
[Talk to us]()
```

I'm not sure if we should implement a flexible way for removing html **attributes**... what you think?

Also, diving a bit deeper—if we introduce `pageOptions.removeTags`, do we still need the `onlyMainContent` option? I'm a bit hesitant to drop it since it removes 52 HTML tags.

mattjoyce:
I'm a bit out of my depth with TS, but happy to look at tests for this - will poke about later this week.

With regards to removing a general way to remove tags, won't pageOptions {}  become utterly bloated if you don;t provide macros for groups of tags, like onlyMainContent does?

rafaelsideguide:
Hey @mattjoyce! I just merged #275, which essentially encapsulates this PR, as it gives users the capability to remove images through HTML tags, classes, and Ids.

mattjoyce:
@rafaelsideguide fair enough, that has more utility and wider appeal.
Looking at #275, will it handle class and ID?   I'm somewhat familiar with soup, I think this will only address tags.
If so, would you be open to something like this :

```
const pageOptions = {
  removeTags: 'nav, footer',     
  removeClasses: 'sidebar, ad',   
  removeAttributes: 'href, src'    
};
```

mattjoyce:
Close?

PR #252, opened 2024-06-07
Fix 208 py sdk interval poll name

mattjoyce:
Just a minor patch to align param name between the two sdk.


nickscamara:
Looks great! Thanks @mattjoyce

PR #253, opened 2024-06-08
Python sdk improve response handling

mattjoyce:
I noticed not all the functions were using the same approach.
Minor tweaks.
Type hints.


nickscamara:
lgtm

PR #255, opened 2024-06-10
Cjp/image added to extract

calebpeffer:
When the useScreenshot parameter in pageOptions is set on LLM extract, the LLM will use the screenshot in addition to the textual content.

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -82,7 +82,7 @@
     "moment": "^2.29.4",
     "mongoose": "^8.0.3",
     "natural": "^6.3.0",
-    "openai": "^4.28.4",
+    "openai": "^4.49.1",
     "pdf-parse": "^1.1.1",
     "pos": "^0.4.2",
     "posthog-node": "^4.0.1",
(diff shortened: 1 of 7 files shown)


ericciarla:
I tested this out with a few urls that I am using for the docs converter. My first impressions is it does not work the way want it to. I may not be specific enough in the prompt or have an incorrect schema but I am getting results like this:

llm_extraction": {
      "titles": [
        {
          "tab_titles": "navigation tabs"
        }
      ]
    }
    
    
Here is what I am testing:

https://developers.dialpad.com/docs
https://developer.mightycall.com/docs/getting-started

With this request body:

{
      "url": "https://developers.dialpad.com/docs",
      "pageOptions": {
        "screenshot":true 
      },
      "extractorOptions": {
        "mode": "llm-extraction",
        "extractionPrompt": "Based on the information on the page, extract the titles of the navigation tabs from left to right. Like Introduction, API reference, etc",
        "extractionSchema": {
          "type": "object",
          "properties": {
                    "titles": {
                              "type": "array",
                              "items": {
                                        "type": "object",
                                        "properties": {
                                                  "tab_titles": {
                                                            "type": "string"
                                                  }
                                        }
                              },
                              "properties": {}
                    }
          }
        }
      }
}

I know this could be a game changer if we get it right! Any thoughts @calebpeffer?

calebpeffer:
Closing because it was unclear if this was an improvement

PR #262, opened 2024-06-11
Only fetch webhook from db if self host webhook not set and using db auth

### LINKED ISSUE #261: [BUG][SELF-HOST] Crawl requests generate Supabase error.
On a self-host system, using latest main branch (10/June/2024).
Make requ4est for crawl via Python SDK.
The following error is show in Docker API-1 console.

> api-1                 | Attempted to access Supabase client when it's not configured.
> api-1                 | Error logging crawl job:
> api-1                 |  Error: Supabase client is not configured.
> api-1                 |     at Proxy.<anonymous> (/app/dist/src/services/supabase.js:38:23)
> api-1                 |     at logCrawl (/app/dist/src/services/logging/crawl_log.js:9:14)
> api-1                 |     at crawlController (/app/dist/src/controllers/crawl.js:87:40)
> api-1                 |     at process.processTicksAndRejections (node:internal/process/task_queues:95:5)

I think it is caused by recent changes to **apps\api\src\services\webhook.ts**, specifically an attempt to get the webhook url from supabase.

If I'm right then I think there are a few options :
1. Simply test for USE_DB_AUTHENTICATION as well, if it's false - no webhook
2. Implement a new env SELF_HOSTED and use that to skip various stuff
3. Bake Superbase in to the self-host Docker setup  (have not got this going yet)

I think #1 is the quickest workaround, but architecturally it would be nice to have local drop in alternatives for all the non-local services.

nickscamara:
fixes #261

### DIFF
--- a/apps/api/src/services/webhook.ts
+++ b/apps/api/src/services/webhook.ts
@@ -3,15 +3,17 @@ import { supabase_service } from "./supabase";
 export const callWebhook = async (teamId: string, data: any) => {
   try {
     const selfHostedUrl = process.env.SELF_HOSTED_WEBHOOK_URL;
+    const useDbAuthentication = process.env.USE_DB_AUTHENTICATION === 'true';
     let webhookUrl = selfHostedUrl;
 
-    if (!selfHostedUrl) {
+    // Only fetch the webhook URL from the database if the self-hosted webhook URL is not set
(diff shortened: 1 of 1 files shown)


rafaelsideguide:
lgtm

Issue #205, opened 2024-05-29
[Feat] Pass back raw PDF so users can implement their own parsing methods


calebpeffer:
**Problem Description**
For the sake of flexibility and robustness, it would be good if we passed back raw pdfs we grab as objects so people can parse them with their own methods. 

**Proposed Feature**

Allow people to access raw pdf bytes in return, giving them the freedom to use their own parsing/conversion methods if they wish. We should also probably add an optional param to toggle this.

Customers have asked for this.

**Additional Context**

Reach out to BS via slack once completed @calebpeffer

rafaelsideguide:
@calebpeffer what do you think about adding a `parsePDF` parameter with a default value of true? If a user sets `"parsePDF": false` in their request, we would return the raw PDF bytes instead.

PR #271, opened 2024-06-12
[Feat] Added parsePDF option to pageOptions

rafaelsideguide:
User can decide if they are going to let us take care of the pdf parsing or they are going to parse the pdf by themselves


rafaelsideguide:
Closes #205

PR #274, opened 2024-06-13
Clusters support to scale our API to # of CPUs running

nickscamara:
Currently our node app instance is only taking advantage of 1 core of our machines, leading to various inefficiencies. This approach allows us to run # of worker processes in parallel (multiple instances of our server in a single machine) which helps boost efficiency and performance.


nickscamara:
No need anymore for performance-8x machines, 2x-4x should be more than enough.

PR #277, opened 2024-06-13
Make maxDepth relative to the entered url

ericciarla:
Making the maxDepth param relative to the entered url depth for crawling.

Why? Trying to avoid confusion on how deep the crawler goes when entering non base urls


nickscamara:
Lovely! @ericciarla could you add a test for it? Thanks!

ericciarla:
@nickscamara I added a test and made a few modifications.

Now relative to the entered url (BaseURL or not)
maxDepth of 0 just scrapes the page
maxDepth of 1 scrapes the url and everything after enteredURL + one /
maxDepth of 2 scrapes the url and everything after enteredURL + one or two /
etc

rafaelsideguide (changes requested):
@ericciarla can you add some tests for different values of maxDepth? like 0, 2, and 3 for example. Another important test is to verify if it's working for urls with a bunch of `/` (like `https://docs.dify.ai/v/zh-han`)

PR #280, opened 2024-06-14
Add rendering service to improve scalability

Ehsan-U:
**Improvements**:

- Removed Playwright from the `playwright-service`
- The Browser instance (Playwright) can now run and scale independently, making it compatible with microservices architecture.
- Support for ad blocking and stealth to counter anti-bot measures.

**Changes**:

- Renamed `playwright-service` to `rendering-service`
- `rendering-service` interacts with the `browser` instance (Playwright) to retrieve rendered HTML content.
- Removed Playwright, resulting in a smaller Docker image for `rendering-service` (from 270MB to 70MB)


rafaelsideguide:
@Ehsan-U Hey! Thanks for the PR. I noticed it switches from Playwright to the 'httpx' package. Could we adjust it to let users choose between them with a `.env` variable? For example, Playwright might be preferred for its broader browser support. It'd be great to keep options open. Thanks for considering this!

Ehsan-U:
@rafaelsideguide
Hey! Thanks for reviewing the PR!

The proposed solution allows to control **Playwright** with a REST API, and **httpx** is simply a client library to communicate with Playwright. Using the API, we can perform almost all necessary actions. This setup allows Playwright instances to run and scale independently than rest of the application.

PR #285, opened 2024-06-14
Feat/max depth relative

ericciarla:


ericciarla:
I am going to remove a couple of the e2e tests I added for the new relative maxDepth PR. It keeps failing and  instead I added a bunch of unit tests

PR #290, opened 2024-06-16
Various change to improve ability to debugs issues

mattjoyce:
It's time consuming to track issues across the services, but useful to be able to.
This series of patches implements FIRECRAWL_LOGGING_LEVEL in the root Docker compose and into the services.

I think it would be good to pick a contemporary logger for the TS components.
I would be happy to help implement it, if you pick one (winston, loglevel, pino)


mattjoyce:
Will resubmit.

PR #296, opened 2024-06-18
Added local host support for the javascript SDK

NeevJewalkar:
This pull request adds a feature to the Node.JS SDK so that it now supports local instances of Firecrawl, rather than hardcoding a single api url.


nickscamara:
Thank you @NeevJewalkar! Just merged!

PR #301, opened 2024-06-18
[Bug] Fixed includeHTML to use cleanedHtml as response

### LINKED ISSUE #291: [Feat] Make it so extract main content also cleans out fluff from the html?
I realized that extracting the main content doesn't change the included HTML when includeHTML is checked.

The question is, would it make sense for the HTML to have the main content extracted?

rafaelsideguide:
Closes #291 

Tested with `requests.http`:
```
POST http://localhost:3002/v0/scrape HTTP/1.1
Authorization: Bearer fc-<YOUR-KEY-HERE>
content-type: application/json

{
    "url":"mendable.ai",
    "pageOptions": {
      "includeHtml": true,
      "onlyMainContent": true
    }
}
```

the `html` should **not** contain "introducing" (from Firecrawl banner)


nickscamara:
lgtm

PR #313, opened 2024-06-24
fix multi-word search term issue: /search (w/o Serp)

snippet:
I tried to use /search without Serp in my self-hosted app, and there was the result of the request:
`
curl -X POST http://localhost:3002/v0/search \
    -d '{
      "query": "What is Mendable?"
    }'
`
Screenshot 2024-06-24 at 1 46 45 PM

This error only happens when the search term is a sentence.

In my fix, I pass the **term** without URLencode as this is apparently not as necessary when it is sent by params via axios.

so there are no problems with special characters after this change:
`
curl -X POST http://localhost:3002/v0/search \
    -d '{
      "query": "Como funciona validação cruzada em LLMs?"
    }'
`
image


nickscamara:
Thank you @snippet! Valeu! :)

PR #326, opened 2024-06-26
[Feat] Added implementation for saving docs on supabase

rafaelsideguide:
~- TODO: remove the comments on `log_job.ts` before deploying to prod~

### DIFF
--- a/apps/api/src/controllers/crawl-status.ts
+++ b/apps/api/src/controllers/crawl-status.ts
@@ -40,7 +40,7 @@ export async function crawlStatusController(req: Request, res: Response) {
       current_url,
       current_step,
       total,
-      data,
+      data: data ? data : null,
       partial_data: jobStatus == 'completed' ? [] : partialDocs,
     });
   } catch (error) {
--- a/apps/api/src/controllers/crawl-status.ts
+++ b/apps/api/src/controllers/crawl-status.ts
@@ -3,6 +3,7 @@ import { authenticateUser } from "./auth";
 import { RateLimiterMode } from "../../src/types";
 import { addWebScraperJob } from "../../src/services/queue-jobs";
 import { getWebScraperQueue } from "../../src/services/queue-service";
+import { supabaseGetJobById } from "../../src/lib/supabase-jobs";
 
 export async function crawlStatusController(req: Request, res: Response) {
   try {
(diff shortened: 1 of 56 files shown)


apps/api/src/controllers/crawl-status.ts
```diff
@@ -20,15 +21,27 @@ export async function crawlStatusController(req: Request, res: Response) {
...
-      current_step: current_step,
-      total: total,
-      data: job.returnvalue,
-      partial_data: partialDocs ?? [],
+      current,
+      current_url,
+      current_step,
+      total,
+      data: data ? data : null,
+      partial_data: jobStatus == 'completed' ? [] : partialDocs,
```
nickscamara, line 44: We lose the partial_data when docs are completed. We need partial data to be there as people rely on it.

nickscamara:
When running with crawlerOptions.returnOnlyUrls, save job throws this error
```
Failed to update job status: Error: canceling statement due to statement timeout
    at saveJob (/Users/nicolascamara/Documents/GitHub/firecrawl/apps/api/src/main/runWebScraper.ts:120:24)
```

nickscamara:
- [ ] need index on job id on firecrawl_jobs, it is a bit slow

rafaelsideguide:
Index is done. Pagination will be implemented on a v1.

nickscamara:
@rafaelsideguide what about the other issues above? partial data and returnOnlyUrls?

mogery:
@nickscamara Added code here that uses Supabase as backing for jobs that are in progress when the server crashes/restarts. On exit, jobs that are active are force unlocked and saved to the DB with `retry: true`. On startup, DB entries with `retry: true` are loaded back into the queue with status `"waiting"`, so the workers will pick them back up.

We are currently not saving scrape progress or `partialDocs`, but it would be a logical next step. Unfortunately it would require some pretty big adjustments to the WebScraperDataProvider.

rafaelsideguide:
I did a load testing (1753 requests) while deploying on staging server and it ended up with 306 stuck jobs in active status. This implementation is better than the one I did few months ago bc back then I had about 30% of jobs stuck, whereas now we're getting 17%, but it seems we're still missing something...

apps/api/src/controllers/crawl-status.ts
nickscamara: Added to the docs that it will become empty when the jobs complete. It should be fine.

nickscamara (approved):
lgtm

PR #332, opened 2024-06-28
Adds pageOptions.includeRawHtml and new extraction mode "llm-extraction-from-raw-html"

ericciarla:
Adds pageOptions.includeRawHtml and new extraction mode "llm-extraction-from-raw-html"

This is fixed extraction for the mintlify converter project.

Need to add tests!

### DIFF
--- a/apps/api/src/controllers/scrape.ts
+++ b/apps/api/src/controllers/scrape.ts
@@ -58,11 +58,18 @@ export async function scrapeHelper(
   }
 
   // make sure doc.content is not empty
-  const filteredDocs = docs.filter(
+  let filteredDocs = docs.filter(
     (doc: { content?: string }) => doc.content && doc.content.trim().length > 0
   );
   if (filteredDocs.length === 0) {
     return { success: true, error: "No page found", returnCode: 200, data: docs[0] };
+  }
+
+  // Remove rawHtml if pageOptions.rawHtml is false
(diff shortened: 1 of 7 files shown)


ericciarla:
Tests added should be good to go!

apps/api/src/controllers/scrape.ts
```diff
@@ -58,19 +58,26 @@ export async function scrapeHelper(
...
-  const filteredDocs = docs.filter(
+  let filteredDocs = docs.filter(
     (doc: { content?: string }) => doc.content && doc.content.trim().length > 0
   );
   if (filteredDocs.length === 0) {
     return { success: true, error: "No page found", returnCode: 200, data: docs[0] };
   }
 
+  // Remove rawHtml if pageOptions.rawHtml is false
+  if (!pageOptions.includeRawHtml) {
```
nickscamara: what is the reason for checking this here?

ericciarla:
TODO: also include a check for the mode there, cause otherwise it will always attempt to delete the keys

rafaelsideguide (changes requested):
Hey @ericciarla the tests are not passing locally either. I think some change might be causing the jobs not to complete.. do you know if this is this running llm extract for every scrape?

nickscamara (approved):

PR #337, opened 2024-06-30
[PROPOSAL] (docker-compose) regroup envs vars between services 

Sanix-Darker:
## WHAT

a small proposal to have a more sync docker-compose for envs vars between the api and the service


rafaelsideguide (approved):
lgtm

PR #338, opened 2024-06-30
[PROPOSAL] (deps): making sure all deps are always up to date

Sanix-Darker:
## WHAT

In other to be up to date regarding dependencies across apps.


rafaelsideguide (approved):
lgtm

PR #343, opened 2024-07-01
Adds support for `pip install firecrawl`

nickscamara:
Currently to install firecrawl via pip one needs to do `pip install firecrawl-py`.

It seems like `pip install firecrawl` would be better, so I have added support for that too. Currently it introduces a bash script to perform the publish that way we ensure we publish to the 2 different packages. In the future I think we should transition to `pip install firecrawl` - following this https://github.com/simonw/pypi-rename

Notes:

1. To use this you need to first change the version in __init__.py and then `chmod +x build_and_publish.sh`and then `./build_and_publish.sh`
2. Update your .pypirc to contain the pypip credentials as follows:

[distutils]
index-servers =
    pypi
    pypi-test

[pypi]
repository = https://upload.pypi.org/legacy/
username = __token__
password = 

[pypi-test]
repository = https://test.pypi.org/legacy/
username = __token__
password =


nickscamara:
Todo:
- [ ] Figure out if this might remove the dependency graph stuff from this repo.

nickscamara:
@rafaelsideguide we might need to make changes to the publish ci in order to publish this

Todo:

 Adjust the CI

rafaelsideguide:
ci changes on #344

PR #374, opened 2024-07-07
Only check Supabase if configured to when cancelling job

StefanTerdell:
Should fix https://github.com/mendableai/firecrawl/issues/372 :)


nickscamara:
@StefanTerdell Good catch! Thanks for the pr!

PR #375, opened 2024-07-07
Self-hosting quality of life fixes

StefanTerdell:
1. Fixes missing env var mapping in docker-compose.yml
2. Allows for connection to local machine from docker-compose network
3. Rudimentary jobId templating for webhook URL (self-hosted only of course)

### DIFF
--- a/apps/api/src/services/webhook.ts
+++ b/apps/api/src/services/webhook.ts
@@ -2,7 +2,7 @@ import { supabase_service } from "./supabase";
 
 export const callWebhook = async (teamId: string, jobId: string,data: any) => {
   try {
-    const selfHostedUrl = process.env.SELF_HOSTED_WEBHOOK_URL;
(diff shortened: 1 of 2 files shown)


apps/api/src/services/webhook.ts
```diff
@@ -2,7 +2,7 @@ import { supabase_service } from "./supabase";
 
 export const callWebhook = async (teamId: string, jobId: string,data: any) => {
   try {
-    const selfHostedUrl = process.env.SELF_HOSTED_WEBHOOK_URL;
+    const selfHostedUrl = process.env.SELF_HOSTED_WEBHOOK_URL?.replace("{{JOB_ID}}", jobId);
```
rafaelsideguide, line 5: @JakobStadlhuber what you think?

JakobStadlhuber: lgtm

Issue #387, opened 2024-07-09
[Improv] JS SDK checkCrawlStatus returning incomplete response


jhoseph88:
**Describe the Bug**
In the [API docs](https://docs.firecrawl.dev/api-reference/endpoint/status#get-crawl-status), the `/crawl/status/{jobId}` advertises several parameters, `current`, `total`, `current_step`, and `current_url`, not present in response via the node sdk. I am seeing only `data`, `partial_data`, `status`, and `success` in the result for `checkCrawlStatus`. Would exposing this via [JobStatusResponse](https://github.com/mendableai/firecrawl/blob/main/apps/js-sdk/firecrawl/types/index.d.ts#L95) be easy enough?

**To Reproduce**
Steps to reproduce the issue:
1. Initialize an `app` object with `new FirecrawlApp`
2. Run `app.crawlUrl(...`
3. Run await `app.checkCrawlStatus(jobId);` with `jobId` as the id of the job returned in the above crawl request
4. Observe fields `current`, `total`, `current_step`, and `current_url` are missing from the response

**Expected Behavior**
`current`, `total`, `current_step`, and `current_url` present in the response

rafaelsideguide:
Hey @jhoseph88 !

Thanks for pointing out the mismatch in the API docs and the node SDK responses. It looks like we need to update the JobStatusResponse to include the missing fields.

Would you be up for helping us fix this? Your assistance would greatly benefit everyone using the SDK. If you could also update the SDK's end-to-end tests to align with these changes, that would be perfect.

Let me know if you’re interested, or if there’s any other way I can assist.

jhoseph88:
@rafaelsideguide Pardon my delay! Absolutely--I've submitted https://github.com/mendableai/firecrawl/pull/391. I've had to mark these missing fields as optional because they periodically end up as `undefined` in my local environment. I'll also point the following tests are failing for me locally:

* `should throw error for invalid API key on scrape`
* `should throw error for invalid API key on crawl`
* `should handle idempotency key for crawl`
* `should throw error for invalid API key on search`

This is likely due to my configuration.

PR #391, opened 2024-07-11
[Feat] Pass along current, total, current_step, and current_url in js sdk

jhoseph88:
issue #387 

Ideally, I think these would be marked as required fields, but due to their instability in my local testing (often resolving to `undefined`), I've marked them all as optional, which I think makes sense for backward compatibility.


nickscamara:
Awesome! Thank you @jhoseph88. Will take a look at it very soon.

nickscamara:
@rafaelsideguide can you get this to the finish line? Thank you! :)

rafaelsideguide:
all tests passing locally. merging.

Issue #421, opened 2024-07-17
[Feat] Return all links on a page in the output


calebpeffer:
Potentia customer asked if they can have all the links returned to a page as a list. Seems like an easy lift that wouldn't cause breaking changes. 

@nickscamara @rafaelsideguide potential customer needs this by @EOW

calebpeffer:
Yes, it is intended - you're correct but I'd imagine the latency introduced is small compared to other sources of delay. Increasing the speed of /scrape is something we should discuss next week.

PR #423, opened 2024-07-17
Caleb: Return a list of links on a page by default

calebpeffer:
For issue #421. I just extract the links from the HTML then return them as the linksOnPageParameter.

I thought about making this optional, but I couldn't see any downside to just including them. Might adding an optional parameter just add complexity?

Would love your take @rafaelsideguide @nickscamara 

Either way, the customer needs this by EOW lets get it merged :)

### DIFF
--- a/apps/api/src/scraper/WebScraper/__tests__/single_url.test.ts
+++ b/apps/api/src/scraper/WebScraper/__tests__/single_url.test.ts
@@ -33,5 +33,5 @@ it('should return a list of links on the mendable.ai page', async () => {
   expect(result.linksOnPage).toBeDefined();
   expect(Array.isArray(result.linksOnPage)).toBe(true);
   expect(result.linksOnPage.length).toBeGreaterThan(0);
-  expect(result.linksOnPage).toContain('https://www.mendable.ai/blog')
+  expect(result.linksOnPage).toContain('https://mendable.ai/blog')
 }, 10000);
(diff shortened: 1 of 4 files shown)


nickscamara (commented):
This looks great @calebpeffer! Could you a few tests for it? Thank you :)

calebpeffer:
> This looks great @calebpeffer! Could you a few tests for it? Thank you :)

Sure! What type of tests would be best here? Whats the expectation for a commit like this? Unit, e2e, both? Don't really know where the testing is at since I haven't contributed in a while

apps/api/src/scraper/WebScraper/single_url.ts
```diff
@@ -109,6 +109,38 @@ function getScrapingFallbackOrder(
   return scrapersInOrder as (typeof baseScrapers)[number][];
 }
 
+function extractLinks(html: string, baseUrl: string): string[] {
```
rafaelsideguide: @calebpeffer could you move this function a extractLinks file inside api/src/scraper/WebScraper/utils?

apps/api/src/scraper/WebScraper/__tests__/single_url.test.ts
```diff
@@ -22,3 +23,14 @@ describe('scrapSingleUrl', () => {
...
+it('should return a list of links on the mendable.ai page', async () => {
+  const url = 'https://mendable.ai';
+  const pageOptions: PageOptions = { includeHtml: true };
+
+  const result = await scrapSingleUrl(url, pageOptions);
+
+  // Check if the result contains a list of links
+  expect(result.linksOnPage).toBeDefined();
+  expect(Array.isArray(result.linksOnPage)).toBe(true);
+  expect(result.linksOnPage.length).toBeGreaterThan(0);
```
rafaelsideguide, line 35: what about expect(result.linksOnPage).toContain('https://www.mendable.ai/blog') ?

rafaelsideguide (changes requested):

apps/api/src/scraper/WebScraper/single_url.ts
calebpeffer: @calebpeffer add docs too lol

PR #436, opened 2024-07-18
fix(WebScraper): infinite regex leading to fly.io instance hangs

mogery:
After SSH-ing into a frozen Fly machine with CPU utilization on 100%, I noticed that the process using the most CPU was Node itself.

Screenshot 2024-07-18 at 19 19 28

After putting Node into debugging mode, attaching the monitor, and pausing the process, it was stuck on a regex in `replacePaths`.

Screenshot 2024-07-18 at 19 21 34

I dumped the document's content and I was able to reproduce the issue locally. I consulted with @rafaelsideguide to replace the problematic regex with something that still worked the same, and I tested the new regex locally.


nickscamara (approved):
lgtm

PR #438, opened 2024-07-19
[Feat] Added rate limit singleton for redis

### LINKED ISSUE #419: Tests: Rate limit tests pass even when Redis is not set
Seems like all the rate limiter tests pass when `REDIS_RATE_LIMIT_URL` is not set in the .env

rafaelsideguide:
Closes #419


nickscamara:
@rafaelsideguide should I still merge this?

rafaelsideguide:
@nickscamara it checks if there's a redis instance. I think it's worth it.

PR #449, opened 2024-07-23
Added regex for links in sitemap

rafaelsideguide:
malformed sitemaps (example `lunit.io/en/`, get's `/` as sitemap url which causes server error)

### DIFF
--- a/apps/api/src/scraper/WebScraper/crawler.ts
+++ b/apps/api/src/scraper/WebScraper/crawler.ts
@@ -64,15 +64,7 @@ export class WebCrawler {
   private filterLinks(sitemapLinks: string[], limit: number, maxDepth: number): string[] {
     return sitemapLinks
       .filter((link) => {
-        
-        // if link is not a complete url, add the base url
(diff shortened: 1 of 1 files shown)


nickscamara:
@mogery can you double check this one?

apps/api/src/scraper/WebScraper/crawler.ts
```diff
@@ -64,6 +64,14 @@ export class WebCrawler {
...
       .filter((link) => {
+        
+        // if link is not a complete url, add the base url
+        link = link.trim();
+        const isCompleteUrl = new RegExp('^(?:[a-z+]+:)?//', 'i');
+        if (!isCompleteUrl.test(link)){
+          link = this.baseUrl + link;
+        }
+
         const url = new URL(link);
```
mogery:
```suggestion
        const url = new URL(link.trim(), this.baseUrl);
```

This is easier, faster, and does not rely on the RegEx engine

mogery (changes requested):

PR #516, opened 2024-08-09
Ensuring USE_DB_AUTHENTICATION is true in single URL scraper.

kevinswiber:
When `USE_DB_AUTHENTICATION` is set to `false` in an environment variable, a boolean check for `process.env.USE_DB_AUTHENTICATION` will always return `true`, as `process.env` values are strings.  This PR fixes it.

Bug introduced in 01fab6e036f44406204cc7ade38116654120a19f.


rafaelsideguide (commented):
Makes sense. `USE_DB_AUTH = "false"` evaluates to `true`. @nickscamara I think we should find a better way to define the scraping strategies. Maybe directly in `.env` file?

rafaelsideguide (approved):

apps/api/src/scraper/WebScraper/single_url.ts
```diff
@@ -85,18 +89,18 @@ function getScrapingFallbackOrder(
...
+    useScrapingBee ? "scrapingBeeLoad" : undefined,
+    useDatabaseAuth ? undefined : "playwright",
     "fetch",
   ].filter(Boolean);
 
   if (isWaitPresent || isScreenshotPresent || isHeadersPresent) {
     defaultOrder = [
       "fire-engine",
-      process.env.USE_DB_AUTHENTICATION ? undefined : "playwright",
+      useDatabaseAuth ? undefined : "playwright",
```
kevinswiber: I do find this a little confusing. Why would we not use Playwright when there's database auth enabled?

rafaelsideguide: Agreed. I've switched to `useFireEngine` instead. No need to check the DB here—using both FireEngine and Playwright is kind of redundant. They both use similar scraping methods, so if one doesn’t work, the other probably won’t either.

yekkhan:
I wanted to kindly inquire about the status of the pull request [#516 ]. Could you let me know when it might be reviewed or merged?

nickscamara (approved):

nickscamara:
@yekkhan merging now!

PR #531, opened 2024-08-12
Self-host fix: Moving comments of .env.example values from end-of-line to above-line.

kevinswiber:
Self-host docs suggest using .env.example as a base. However, Docker doesn't respect end-of-line comments. It sets the comment as the actual value of the variable, leading to invalid behavior. This fix prevents that.


rafaelsideguide (approved):
lgtm

nickscamara:
Thanks @kevinswiber!

Issue #533, opened 2024-08-13
[BUG] go-sdk module name does not match with project structure


KentHsu:
**Describe the Bug**
The module name in [go.mod](https://github.com/mendableai/firecrawl/blob/main/apps/go-sdk/go.mod#L1) does not reflect the project structure. This leads to package import problem.

**To Reproduce**
Go tool does not download this go-sdk when running `go get github.com/mendable/firecrawl` command

**Expected Behavior**
Run `go get github.com/mendable/firecrawl` command should download this go-sdk (with a pseudo-version)
We should be able to see this go-sdk on [Go Packages](pkg.go.dev)

**Screenshots**
![image](https://github.com/user-attachments/assets/eecb5e5f-f60b-4b6b-85ff-98dc6cb47572)

KentHsu:
Hi @ericciarla, The issue still persists. I'm having a hard time making Go tools able to locate go-sdk in subdirectory. Is it possible to move go-sdk to a standalone project under mendableai? We can name it `firecrawl-go` so users can download Firecrawl go-sdk by `go get github.com/mendableai/firecrawl-go`.

ericciarla:
Hi @KentHsu have you tried using Go modules or adjusting the import path? Ideally we want it in the mono repo bc it is easier to maintain.

nickscamara:
Reopening for visibility. @rafaelsideguide feel free to jump in here

nickscamara:
added some comment on the closed pr (not sure if that helps) https://github.com/mendableai/firecrawl/pull/534

rafaelsideguide:
@nickscamara tested here. Didn't work. I'm creating a repo for the go sdk

PR #497, opened 2024-08-04
[Feat] Add Go SDK implementation

KentHsu:
# Description

This PR add go-sdk based on js-sdk, python-sdk and rust-sdk implementations.

## Issue reference

Please reference the issue this PR will close: #481 

## Checklist

- [x] Implement firecrawl.go
- [x] Add tests for firecrawl.go
- [x] Add examples for go-sdk
- [x] Add README.md
- [x] Update GitHub Action workflow

PR #534, opened 2024-08-13
fix: go-sdk module name

KentHsu:
# Description

This PR fixes go-sdk module name, so it matches with the project layout
README and example are also updated with new module name

## Issue reference

Please reference the issue this PR will close: #533  

## Prior work reference

Firecrawl go-sdk is added in this PR: #497

### DIFF
--- a/apps/go-sdk/examples/go.mod
+++ b/apps/go-sdk/examples/go.mod
@@ -1,10 +1,10 @@
-module github.com/mendableai/firecrawl/go-sdk/examples
+module github.com/mendableai/firecrawl/apps/go-sdk/examples
 
 go 1.22.5
 
-replace github.com/mendableai/firecrawl/go-sdk => ../
+replace github.com/mendableai/firecrawl => ../
 
 require (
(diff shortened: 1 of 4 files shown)


apps/go-sdk/examples/go.mod
```diff
@@ -1,10 +1,10 @@
-module github.com/mendableai/firecrawl/go-sdk/examples
+module github.com/mendableai/firecrawl/apps/go-sdk/examples
```
nickscamara, line 1: shouldnt be tree/main/apps?

apps/go-sdk/go.mod
```diff
@@ -1,4 +1,4 @@
-module github.com/mendableai/firecrawl/go-sdk
+module github.com/mendableai/firecrawl/apps/go-sdk
```
nickscamara, line 1: shouldnt be tree/main/apps?

nickscamara: (not sure though) never used go haha

PR #535, opened 2024-08-13
fix docker compose port setting.

matsubo:
# Test

```
❯ cd firecrawl
❯ cp ./apps/api/.env.example ./.env
❯ curl -X POST http://localhost:8181/v0/crawl \
          -d '{
        "url": "https://mendable.ai"
      }'
{"error":"Unauthorized"}⏎
```

### DIFF
--- a/SELF_HOST.md
+++ b/SELF_HOST.md
@@ -42,6 +42,7 @@ To start, we wont set up authentication, or any optional sub services (pdf parsi
 ```
 # ===== Required ENVS ======
 NUM_WORKERS_PER_QUEUE=8
+EXPOSE_PORT=8080
 PORT=3002
 HOST=0.0.0.0
 REDIS_URL=redis://redis:6379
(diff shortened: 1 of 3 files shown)


rafaelsideguide:
@matsubo, I don't understand the need for it. Could we just use 3002 instead of 8181?

gakugaku:
I believe I am facing the same issue as described in this PR, which is how I found my way here. Typically, when launching a container with Docker, the port exposed to the host doesn't necessarily have to match the one used by the app inside the container. It can be arbitrarily assigned, which helps to resolve port conflicts on the host or to run multiple instances of the same app.

However, in the current setup of [docker-compose.yaml](https://github.com/mendableai/firecrawl/blob/a9773a24a36fa1bdbfe76405b5cc465ab44d614c/docker-compose.yaml#L52), the exposed port is hardcoded to 3002, and changing it requires modifying the repository code directly. I don't specifically need to use port 3002 or 8181, but would like the flexibility to assign any arbitrary port.

matsubo:
I tested on the HEAD of main branch, the issue I encountered is resolved.

PR #554, opened 2024-08-19
Check team credits based on the crawl limit

nickscamara:


nickscamara:
Makes sure users has enough credits to perform a crawl when they choose a limit

PR #576, opened 2024-08-26
[Fix] Added a special check for large tables

rafaelsideguide:
corterix.com has a 80933 rows table :D


rafaelsideguide:
This no longer makes sense since we’re now using the go html to md parser.

PR #594, opened 2024-08-30
[v1] Webhooks

rafaelsideguide:


nickscamara:
- [ ] Need docs

apps/api/src/services/webhook.ts
```diff
@@ -50,7 +50,7 @@ export const callWebhook = async (teamId: string, jobId: string, data: any, spec
       },
       body: JSON.stringify({
         success: data.success,
-        jobId: jobId,
+        id: id,
```
nickscamara: cant change this for v0

apps/api/src/services/queue-worker.ts
```diff
@@ -341,11 +347,20 @@ async function processJob(job: Job, token: string) {
...
-
-        await callWebhook(job.data.team_id, job.data.crawl_id, data, job.data.webhook, job.data.v1);
+          // v0 web hooks, call when done with all the data
+        if (!job.data.v1) {
+          callWebhook(job.data.team_id, job.data.crawl_id, data, job.data.webhook, job.data.v1, "crawl.completed");
+        }
+        // v1 web hooks, call when done with no data, but with event completed
+        if (job.data.v1 && job.data.webhook) {
+          callWebhook(job.data.team_id, job.data.crawl_id, [], job.data.webhook, job.data.v1, "crawl.completed");
+        }
```
mogery: do this before gathering all data to spare server RAM

mogery (changes requested):
lgtm other than that one thing. it's a serious issue

apps/api/src/services/queue-worker.ts
nickscamara: fixed

PR #611, opened 2024-09-03
build(js-sdk): simplify build process

MonsterDeveloper:
# Brief history

My initial goal was to fix the error from the [`arethetypeswrong`](https://arethetypeswrong.github.io/):
CleanShot 2024-09-03 at 10 53 07@2x

Annoyed enough with the fix I though dropping CJS entirely might be a viable option, because:
- It would simplify the build process
- Modern packages are to be ESM-only (see a paragraph on the [Rollup website](https://rollupjs.org/faqs/#why-are-es-modules-better-than-commonjs-modules))
- Consumers of the SDK are also expected to be newly created, modern, ESM-only apps, so why we need CJS after all?
- Avoid [dual package hazard](https://github.com/GeoffreyBooth/dual-package-hazard)

## Changes made
- updated `tsconfig.json` to be ESM-only, also enable `verbatimModuleSyntax` to explicitly require `type` prefix when imports are only used as types
- updated exports in `package.json`

`attw` output after the changes:
CleanShot 2024-09-03 at 11 00 18@2x


mogery:
Hi there -- we need to retain CJS support, it would break compatibility with a lot of our major customers.

MonsterDeveloper:
@mogery understood. Made a new commit bringing back CJS exports but with a simplified workflow using `tsup` (per the [guide by Matt Pocock](https://www.totaltypescript.com/how-to-create-an-npm-package?ck_subscriber_id=2306022541#6-using-tsup-to-dual-publish))

Exports look good:
CleanShot 2024-09-11 at 17 53 28@2x

mogery:
Thanks! Will review this today.

mogery (approved):
LGTM! Gonna do some further testing but this can go into mainline at this point.

PR #613, opened 2024-09-03
Prevent Python Firecrawl logger from interfering with loggers in client 

reasonmethis:
This PR updates the logging setup in the Firecrawl Python SDK to replace the use of `basicConfig`. 

**Why this change is important:**
Using `basicConfig` in a library can unintentionally change the logging settings of any application that imports the library. `basicConfig` modifies the global logging configuration, which can cause conflicts, such as:
- Overriding the logging format or level set by the application.
- Adding duplicate log entries if the application has already configured its logging.

To avoid these problems, this PR introduces a more controlled logging setup that:
- Configures logging specifically for the Firecrawl package without affecting the rest of the application.
- Ensures that Firecrawl’s logs are handled properly without interfering with the global logging configuration.

### DIFF
--- a/apps/python-sdk/firecrawl/__init__.py
+++ b/apps/python-sdk/firecrawl/__init__.py
@@ -11,32 +11,54 @@ For more information visit https://github.com/firecrawl/
 import logging
 import os
 
-from .firecrawl import FirecrawlApp
+from .firecrawl import FirecrawlApp # noqa
 
-__version__ = "1.2.1"
+__version__ = "1.2.2"
 
 # Define the logger for the Firecrawl project
 logger: logging.Logger = logging.getLogger("firecrawl")
 
 
-def _basic_config() -> None:
-    """Set up basic configuration for logging with a specific format and date format."""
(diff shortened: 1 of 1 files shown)


apps/python-sdk/firecrawl/__init__.py
```diff
@@ -11,32 +11,54 @@
...
-    _basic_config()
+    # Check if the firecrawl logger already has a handler
+    if logger.hasHandlers():
+        return # To prevent duplicate logging
+
+    # Check if the FIRECRAWL_LOGGING_LEVEL environment variable is set
+    if not (env := os.getenv("FIRECRAWL_LOGGING_LEVEL", "").upper()):
+        # Attach a no-op handler to prevent warnings about no handlers
+        logger.addHandler(logging.NullHandler()) 
+        return
```
mogery, line 56: Good call to have lack of env disable the logger. Gonna need to document this somewhere on release.

mogery (approved):
LGTM! Thank you!

PR #614, opened 2024-09-03
perf(js-sdk): move `dotenv` and `uuid` to `devDependencies`, fix `zod` import

MonsterDeveloper:
## Changes made
- Moved `dotenv` and `uuid` from `dependencies` to `devDependencies` since they're only used in tests
- Fixed `zod` import since before sometimes `import { z } from "zod"` would remain in the build js files, causing the whole zod to be imported even if it's not used

### DIFF
--- a/apps/js-sdk/firecrawl/package-lock.json
+++ b/apps/js-sdk/firecrawl/package-lock.json
@@ -1,19 +1,17 @@
 {
   "name": "@mendable/firecrawl-js",
-  "version": "1.1.0",
+  "version": "1.2.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@mendable/firecrawl-js",
-      "version": "1.1.0",
+      "version": "1.2.1",
       "license": "MIT",
       "dependencies": {
         "axios": "^1.6.8",
-        "dotenv": "^16.4.5",
         "isows": "^1.0.4",
         "typescript-event-target": "^1.1.1",
-        "uuid": "^9.0.1",
         "zod": "^3.23.8",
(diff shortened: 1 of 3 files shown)


mogery (approved):
LGTM, thanks!

MonsterDeveloper:
Another thing to consider: dropping `axios` dependency cause it's huge-[13,7kb](https://bundlephobia.com/package/axios@1.7.7)

Instead use global fetch if it's available, if not `require` the isomorphic fetch or something like that.

Similar to how [Inngest SDK](https://github.com/inngest/inngest-js/blob/9ff872d7dc7be684eef6153234e66e5e57a24536/packages/inngest/src/helpers/env.ts#L491) handles it.

@mogery what do you think?

mogery:
> @mogery what do you think?

Absolutely: we have axios bloat everywhere. Would need to spend some time thinking of how to port it without breaking compatibility WRT error handling.

mogery:
Yep, just worried about whether people are relying on the AxiosError interface for error handling.

Anyways, we'd very much appreciate a PR, as this would be a great improvement but we're bogged down by infrastructure concerns this week.

MonsterDeveloper:
> just worried about whether people are relying on the AxiosError interface for error handling

Valid concern. You could make it a major version bump to be on the safe side.

> we'd very much appreciate a PR, as this would be a great improvement but we're bogged down by infrastructure concerns this week

Perfect, look out for new notifications then:) Take your time.

And btw, thanks for the Firecrawl—an amazing product, but the JS SDK really needs some work. Hopefully going to look a lot better by the end of our edits.

mogery:
Thank you!

PR #623, opened 2024-09-04
fix wrong link to self host documentation

itasli:


rafaelsideguide:
@itasli, the CONTRIBUTING guide includes a "Running Locally" section that explains how to run the services without Docker, which is why we added that link. Maybe we could link both the SELF_HOST and CONTRIBUTING guides? What do you think?

PR #624, opened 2024-09-04
Optimize check credits func w/ caching

nickscamara:
Reduces load in the db and faster requests overall


apps/api/src/services/queue-jobs.ts
nickscamara, line 1: Reduced the polling to 500ms instead

apps/api/src/services/billing/credit_billing.ts
```diff
@@ -325,17 +365,38 @@ export async function supaCheckTeamCredits(team_id: string, credits: number) {
...
+      if (priceError) {
+        throw new Error(
+          `Failed to retrieve price for price_id: ${subscription.price_id}`
+        );
+      }
+
+      price = data;
+      // There are only 21 records, so this is super fine
+      // Cache the price for a long time (e.g., 1 day)
+      await setValue(priceCacheKey, JSON.stringify(price), 86400);
```
nickscamara, line 393: // There are only 21 records, so this is super fine as we don't change the table often
 // Cache the price for a long time (e.g., 1 day)

apps/api/src/services/billing/credit_billing.ts
```diff
@@ -312,7 +352,7 @@ export async function supaCheckTeamCredits(team_id: string, credits: number) {
 
         if (creditUsages && creditUsages.length > 0) {
           totalCreditsUsed = creditUsages[0].total_credits_used;
-          await setValue(cacheKey, totalCreditsUsed.toString(), 1800); // Cache for 30 minutes
+          await setValue(cacheKey, totalCreditsUsed.toString(), 500); // Cache for 8 minutes
```
nickscamara, line 355: decreased a bit the caching on the credits with subs. 30min was too much

PR #634, opened 2024-09-06
[Feat] Added fetch sitemap for yoastSEO generated sitemaps

rafaelsideguide:
Example omnyhealth.com previously had 35 results, now 131.

Cons: pages without sitemap.xml in general can take a little bit longer.

### DIFF
--- a/apps/api/src/scraper/WebScraper/crawler.ts
+++ b/apps/api/src/scraper/WebScraper/crawler.ts
@@ -156,6 +156,13 @@ export class WebCrawler {
     if (sitemapLinks.length > 0) {
       let filteredLinks = this.filterLinks(sitemapLinks, this.limit, this.maxCrawledDepth);
       return filteredLinks.map(link => ({ url: link, html: "" }));
+    } else {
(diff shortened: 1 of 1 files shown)


nickscamara:
This could slow the crawler for everyone that doesn't use yoastSEO, might be better to do it concurrently with the sitemap fetching.

apps/api/src/scraper/WebScraper/crawler.ts
```diff
@@ -156,6 +156,13 @@ export class WebCrawler {
     if (sitemapLinks.length > 0) {
       let filteredLinks = this.filterLinks(sitemapLinks, this.limit, this.maxCrawledDepth);
       return filteredLinks.map(link => ({ url: link, html: "" }));
+    } else {
+      // Yoast SEO sitemap index
+      let sitemapLinks = await this.tryFetchSitemapLinks(this.initialUrl, "/sitemap_index.xml");
```
nickscamara, line 161: Can we do this concurrently when we fetch the sitemaps?

PR #635, opened 2024-09-06
[Docs] upgraded the path of the self-hosted documentation URL to `/v1`.

shige:
I was attempting self-hosting by referring to the following document, and I noticed that the URL path was outdated, so I fixed it.

- https://github.com/mendableai/firecrawl/blob/v1.0.0/SELF_HOST.md

I have compared it with the official documentation at https://docs.firecrawl.dev/api-reference/endpoint/crawl-post, and the issue has been corrected.


rafaelsideguide:
lgtm

PR #654, opened 2024-09-10
fix(py-sdk): removed asyncio package

### LINKED ISSUE #644: [BUG] Python asyncio modules should not be explicitly listed in dependencies
**Describe the Bug**
The `asyncio` module name is specified in `requirement.txt` . This causes old [asyncio](https://pypi.org/project/asyncio/) modules to be imported to support python 3.3 in addition to the core module ones, and extra dependencies to be included.

> This version is only relevant for Python 3.3, which does not include asyncio in its stdlib.
ref: https://pypi.org/project/asyncio/#:~:text=This%20version%20is%20only%20relevant%20for%20Python%203.3%2C%20which%20does%20not%20include%20asyncio%20in%20its%20stdlib.

There are also platforms where this behavior causes bugs. For example, such a post has already been made regarding deploying to aws lambda.

https://stackoverflow.com/questions/59546682/aws-lambda-not-importing-asyncio

**To Reproduce**
After installing `firecrawl-py` , one can easily discover the old asyncio in the dependencies.

```
asyncio==3.4.3
    # via firecrawl-py
```

**Expected Behavior**

**Screenshots**
If applicable, add screenshots or copies of the command line output to help explain the issue.

- OS: Linux
- Firecrawl Version: 1.2.3
- Node.js Version: xxx

**Logs**
```console
Runtime.UserCodeSyntaxError: Syntax error in module 'lambda_function': invalid syntax (base_events.py, line 296)
```

**Additional Context**
Add any other context about the problem here, such as configuration specifics, network conditions, data volumes, etc.

rafaelsideguide:
tested websocket with example.py without asyncio and it works with no problem.

Closes #644

### DIFF
--- a/apps/python-sdk/firecrawl/firecrawl.py
+++ b/apps/python-sdk/firecrawl/firecrawl.py
@@ -13,7 +13,6 @@ import logging
 import os
 import time
 from typing import Any, Dict, Optional, List
-import asyncio
 import json
 
 import requests
--- a/apps/python-sdk/pyproject.toml
+++ b/apps/python-sdk/pyproject.toml
@@ -12,8 +12,7 @@ dependencies = [
     "requests",
     "python-dotenv",
     "websockets",
-    "asyncio",
-"nest-asyncio"
+    "nest-asyncio"
 ]
 authors = [{name = "Mendable.ai",email = "nick@mendable.ai"}]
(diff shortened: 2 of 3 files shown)


moezali1:
@nickscamara @rafaelsideguide when will the new version on pypi be released with this fix?

PR #679, opened 2024-09-17
remove space in the examples/o1_web_crawler folder name

h4r5h4:
there is an unintentional space in the folder name that's preventing syncing forks


rafaelsideguide:
Hey @h4r5h4 , Good catch! We're still getting traffic from a twitter post with the link, so we're going to merge this one next week.

rrg92:
For those experiencing the same issue when cloning on Windows, a workaround:
It seems to be not just a space, but some unsupported Unicode character by NTFS...

```bash
mkdir SomeDir
cd SomeDir
git clone https://github.com/mendableai/firecrawl .
git restore --staged *; git restore *
```

This will prevent the `examples` directory from being created.

ericciarla:
Merging!

PR #685, opened 2024-09-19
bugfix: using onlyIncludeTags and removeTags together

skeptrunedev:
## Issue

`apps/api/src/scraper/WebScraper/utils/removeUnwantedElements.ts` early returned when `onlyIncludeTags` was specified and there were more than `0` matches in the DOM.

## Fix

Removing the early return and modifying `soup`.

## Adjacent fixes

`cheerio` as a top-level import is deprecated and `load` called directly is preferred. I made those changes and verified tests still passed.

## Running Only the Modified Tests

Use the following command from the root:

`npx jest apps/api/src/scraper/WebScraper/utils/__tests__/removeUnwantedElements.test.ts`

## Context

Trieve is currently building a docsearch component like Algolia's. Firecrawl is being used for the scraping. Lots of doc sites have undesirable search index content which we want to remove for better performance. When trying for the [signoz docs](https://signoz.io/docs/introduction/), we noticed that there were issues using both the onlyIncludeTags and removeTags system together like as shown in the test.


mogery (approved):
LGTM! Thank you.

nickscamara (approved):

PR #690, opened 2024-09-22
Fix the error message when trying search in v0

nickscamara:


apps/api/src/scraper/WebScraper/scrapers/fireEngine.ts
```diff
@@ -194,7 +194,7 @@ export async function scrapWithFireEngine({
       logParams.error_message = data.pageError ?? data.error;
       return {
         html: data.content ?? "",
-        screenshots: data.screenshots,
+        screenshots: data.screenshots ?? [data.screenshot] ?? [],
```
nickscamara, line 197: added this in case legacy returns (with fallbacks)

PR #714, opened 2024-09-28
`WebScraper` refactor into `scrapeURL`

mogery:
Directives:
- reduce state
  - stateless, functional programming paradigms to reduce debugging complexity
    - state that is required (e.g. current logger) is passed in an immutable meta object
  - sovereign modules that do not see the whole state (see e.g. `engines/fire-engine/scrape.ts`, it only has `logger`, not the whole `meta` object)
- make the signal flow clear to ease debugging
  - intense verbosity in logging
  - modularity, make it clear where to add things in the future, make it easy to add things in the future without breaking stuff
    - define generic modules that can be implemented and appended to later (e.g. `transformers`, `engines`)
 - better error handling
   - using a rust-like error model, exceptions are freely thrown instead of wrapping it into `{success: false, error: ...}` objects
   - errors that occur are **always** re-thrown, with the original metadata (e.g. `stack`) intact. if doing stuff like retries with a limit, previous errors are passed along using the `cause` property.
   - we may **never** swallow an error.
     - at points where errors are not directly re-thrown, rather put into an object/array (e.g. retry logic/`EngineResultsTracker`), non-expected errors should be explicitly logged and `Sentry.captureException`'d
   - errors are only transformed into a success object at the top level of `scrapeURL`, in order to avoid breaking other parts of the codebase. errors are passed in the `error` metadata
   - never determine what an error is by checking it's `message` -- if you need a specific error that is determinable by parts of the codebase, create a custom error class and use `instanceof` -- see `error.ts` for reference
 - standalone
   - `scrapeURL` should never (even attempt to!) interface with the database. It should be its own standalone thing that could even be lifted out of `firecrawl` as a whole. To keep it fast, reliable, and maintainable, we need to keep its footprint minimal -- DB code can be handled by the surrounding bits that are tangled up in that anyways (e.g. `queue-worker.ts`)


apps/api/src/controllers/v1/crawl.ts
```diff
@@ -34,18 +34,22 @@ export async function crawlController(
...
-  const crawlerOptions = legacyCrawlerOptions(req.body);
-  const pageOptions = legacyScrapeOptions(req.body.scrapeOptions);
+  const crawlerOptions = {
+    ...req.body,
+    url: undefined,
+    scrapeOptions: undefined,
+  };
+  const scrapeOptions = req.body.scrapeOptions;
 
   // TODO: @rafa, is this right? copied from v0
```
tomkosm, line 50: @rafaelmmiller have we checked this?

apps/api/src/controllers/v1/types.ts
```diff
@@ -453,68 +455,88 @@ export function legacyCrawlerOptions(x: CrawlerOptions) {
...
+        systemPrompt: extractorOptions.extractionPrompt,
+        prompt: extractorOptions.userPrompt,
+        schema: extractorOptions.extractionSchema,
+      } : undefined,
+      mobile: pageOptions.mobile,
+    }),
+    internalOptions: {
+      atsv: pageOptions.atsv,
+    },
+    // TODO: fallback, fetchPage  Content, replaceAllPathsWithAbsolutePaths, includeLinks, useFastMode, disableJsDom
```
tomkosm: @mogery dont forget this

apps/api/src/scraper/scrapeURL/engines/fire-engine/scrape.ts
```diff
@@ -0,0 +1,96 @@
+import { Logger } from "winston";
+import * as Sentry from "@sentry/node";
+import { z } from "zod";
+
+import { Action } from "../../../../lib/entities";
+import { robustFetch } from "../../lib/fetch";
+
+export type FireEngineScrapeRequestCommon = {
+    url: string;
```
tomkosm, line 9: @mogery should we have here all of the options fire-engine supports? If so I think we are missing a few. ie atsv,mobileProxy and more

apps/api/src/scraper/scrapeURL/engines/index.ts
```diff
@@ -0,0 +1,273 @@
...
+] as const;
+
+export type FeatureFlag = typeof featureFlags[number];
+
+export const featureFlagOptions: {
+    [F in FeatureFlag]: {
+        priority: number;
+    }
+} = {
+    "actions": { priority: 20 },
```
tomkosm, line 46: @mogery can you explain the idea behind the priorities? Is there a chance that we will send a request to a scraper that doesnt have one of the capabilities?

apps/api/src/scraper/scrapeURL/index.ts
```diff
@@ -0,0 +1,263 @@
...
+    }
+
+    if (urlO.pathname.endsWith(".docx")) {
+        flags.add("docx");
+    }
+
+    return flags;
+}
+
+function buildMetaObject(id: string, url: string, options: ScrapeOptions, internalOptions: InternalOptions): Meta {
```
tomkosm, line 91: @mogery can you add a comment explaining the idea of a meta object + why we remove www?

apps/api/src/scraper/scrapeURL/index.ts
```diff
@@ -0,0 +1,263 @@
...
+            meta.logger.info("Scraping via " + engine + "...");
+            const _engineResult = await scrapeURLWithEngine(meta, engine);
+            if (_engineResult.markdown === undefined) { // Some engines emit Markdown directly.
+                _engineResult.markdown = await parseMarkdown(_engineResult.html);
+            }
+            const engineResult = _engineResult as EngineScrapeResult & { markdown: string };
+
+            // Success factors
+            const isLongEnough = engineResult.markdown.length >= 100;
+            const isGoodStatusCode = engineResult.statusCode < 300;
```
tomkosm: dont think this is a correct way to define success (ie 304), some 300 status codes indicate success really + I am not sure if 100s are valid to think of as success

apps/api/src/scraper/scrapeURL/lib/extractMetadata.ts
```diff
@@ -0,0 +1,132 @@
...
+    dcTermsSubject = soup('meta[name="dcterms.subject"]').attr("content") || undefined;
+    dcTermsAudience = soup('meta[name="dcterms.audience"]').attr("content") || undefined;
+    dcType = soup('meta[name="dc.type"]').attr("content") || undefined;
+    dcTermsType = soup('meta[name="dcterms.type"]').attr("content") || undefined;
+    dcDate = soup('meta[name="dc.date"]').attr("content") || undefined;
+    dcDateCreated = soup('meta[name="dc.date.created"]').attr("content") || undefined;
+    dcTermsCreated = soup('meta[name="dcterms.created"]').attr("content") || undefined;
+
+    try {
+      // Extract all meta tags for custom metadata
```
tomkosm, line 74: why would this not work for the base cases?

apps/api/src/scraper/scrapeURL/lib/fetch.ts
```diff
@@ -0,0 +1,144 @@
...
+        return null as Output;
+    }
+
+    const response = {
+        status: request.status,
+        headers: request.headers,
+        body: await request.text(), // NOTE: can this throw an exception?
+    };
+
+    if (request.status >= 300) {
```
tomkosm, line 89: again dont think this is a valid way to define success

nickscamara:
- [ ] Figure out the skipTLS stuff

apps/api/src/scraper/scrapeURL/index.ts
nickscamara: On the isGoodStatusCode, lets do >= 1 to get rid of that

nickscamara: made it `const isLongEnough = engineResult.markdown.length >= 1;`

nickscamara:
- [ ] Add tests for fire-engine params

apps/api/src/scraper/scrapeURL/lib/urlSpecificParams.ts
```diff
@@ -0,0 +1,78 @@
+import { InternalOptions } from "..";
+import { ScrapeOptions } from "../../../controllers/v1/types";
+
+export type UrlSpecificParams = {
+    scrapeOptions: Partial<ScrapeOptions>,
+    internalOptions: Partial<InternalOptions>,
+};
+
+const docsParam: UrlSpecificParams = {
```
tomkosm: why are we overriding the headers?

nickscamara:
- [ ] Fire engine params like mobile proxy are not there

apps/api/src/scraper/scrapeURL/transformers/uploadScreenshot.ts
```diff
@@ -0,0 +1,26 @@
+// This file is an exception to the "no supabase in scrapeURL" rule,
```
tomkosm, line 1: when do we use this? fire-engine uploads screenshots on its own, is it for playwright?

apps/api/src/scraper/scrapeURL/engines/fire-engine/scrape.ts
mogery: We have atsv in tlsclient options, which is the only thing that does anything with that flag. Looking into mobileProxy

apps/api/src/scraper/scrapeURL/engines/index.ts
mogery: If there is no engine which matches all of the capabilities requested, or if a better engine fails on a low-priority feature, then an engine with the missing feature will be attempted as well, and a `warning` is added to the result. An easy way to test this in action is to scrape a PDF while specifying e.g. `actions` in the options.

apps/api/src/scraper/scrapeURL/index.ts
mogery: We do not remove `www`, only for matching the url specific options object, which is the same format as in WebScraper. Adding comment.

apps/api/src/scraper/scrapeURL/lib/extractMetadata.ts
mogery: What do you mean?

apps/api/src/scraper/scrapeURL/lib/fetch.ts
mogery: It is for what robustFetch is used for, which is making rudimentary API calls. This is a non-issue.

apps/api/src/controllers/v1/types.ts
nickscamara: - [ ] Add these

apps/api/src/scraper/scrapeURL/transformers/uploadScreenshot.ts
mogery: This is for scrapingbee.

apps/api/src/controllers/v1/types.ts
mogery: A lot of these are deprecated args not used by anyone. Talked with Nick about this a bit ago.

apps/api/src/scraper/scrapeURL/engines/fire-engine/scrape.ts
tomkosm: tlsclient also misses disableJsDom, cdp misses screenshot probably more didnt fully check
https://github.com/mendableai/fire-engine/blob/main/api/src/controllers/scrape.ts

apps/api/src/scraper/scrapeURL/lib/urlSpecificParams.ts
mogery: I don't know, took this from WebScraper. @nickscamara any idea?

apps/api/src/scraper/scrapeURL/engines/index.ts
tomkosm: Not sure if thats a good idea, could be hard to debug. Might be better to just fail or retry the same engine a few times.

apps/api/src/scraper/scrapeURL/index.ts
tomkosm: can you explain the meta concept?

apps/api/src/scraper/scrapeURL/lib/extractMetadata.ts
tomkosm: Wont the way you extract the custom ones work for the predefined ones above too?

apps/api/src/scraper/scrapeURL/lib/fetch.ts
tomkosm: what is robustFetch?

mogery: Replacing `axios`, used to make calls to `fire-engine`, `scrapingbee`, the playwright microservice, etc.

apps/api/src/controllers/v1/types.ts
tomkosm: useFastMode, disableJsDom are not, idk about the rest

apps/api/src/scraper/scrapeURL/lib/extractMetadata.ts
mogery: It does not retain compatibility for WebScraper, and breaks tests. This was also in WebScraper.

apps/api/src/scraper/scrapeURL/index.ts
mogery: Adding comment.

apps/api/src/scraper/scrapeURL/engines/index.ts
mogery: That's why there is a warning. There is also an explicit log message for it.

apps/api/src/scraper/scrapeURL/engines/fire-engine/scrape.ts
mogery: CDP most definitely does not miss screenshot. There is a test case for it. It's transformed into an action, since you dropped the `screenshot` parameter support.

mogery: disableJsDom is unused and undocumented, last I checked.

apps/api/src/scraper/scrapeURL/engines/index.ts
tomkosm: But this is a new feature,legacy customers are not checking for the warning. I think if we want to do this it should probably be opt-in

apps/api/src/controllers/v1/types.ts
nickscamara: (me to look over)

apps/api/src/scraper/scrapeURL/lib/fetch.ts
tomkosm: oh ok, if its internal only we are good. Check the other comment about the same issue, I think that might be for non internal stuff

apps/api/src/scraper/scrapeURL/engines/index.ts
mogery: Legacy customers don't use any feature that has a priority low enough to be affected by this.

apps/api/src/scraper/scrapeURL/engines/fire-engine/scrape.ts
tomkosm: disableJsDom is used by atleast one customer + was used by sitemap fetch at some point. We should add it and document it

mogery: Adding...

apps/api/src/scraper/scrapeURL/engines/index.ts
tomkosm: Maybe I am not understanding it correctly, an example of the way I think it works below.
Customer makes a request with ie screenshot, all of the scrapers that can take a screenshot fail. My understanding is that we fallback to a scraper without screenshot, is that correct?

mogery: No. Screenshot is priority 10. For a request that only requires screenshots, the max engine score will be 10 (all required features' priority summed). Engines that support screenshot will have a score of 10, engines that do not will have a score of 0. Only engines with a score above `maxScore / 2` (in this case `5`) will be attempted. Therefore, only engines that support screenshot will be attempted.

This is more useful for complex requests that require multiple features of different importance.

mogery: e.g. if `pdf` is required, it has a priority of `100`, so only the PDF engine can win.

tomkosm: Hmm I see. Wouldnt it best to hardcode this behaviour? Cause this is dynamic, not easy to read/understand and could potentially cause weird cases. Its best to keep things like this easily reproducible/understandable

tomkosm: Its not clear for which features this can happen the way its written, not sure if there is enough benefit in the dynamic way of determining that

mogery: It is always reproducible by re-sending the request. These fallback incidents are logged and reported to the end user. It allows us to have better DX and less pedantic API errors. It's fine and handled well for something that's dynamic IMO.

apps/api/src/scraper/scrapeURL/index.ts
mogery: Added

apps/api/src/scraper/scrapeURL/engines/fire-engine/scrape.ts
mogery: Added

apps/api/src/controllers/v1/types.ts
mogery: > useFastMode, disableJsDom are not, idk about the rest

Added those

apps/api/src/scraper/scrapeURL/engines/fire-engine/scrape.ts
mogery: forgot mobileProxy, looking.

mogery: Wait. We don't use `mobileProxy` anywhere. Not exposed to `v0`, and I got rid of the URL-specific override that used it.

tomkosm: is it not exposed on v1 either?

mogery: Nope. Do a search for `mobileProxy` on the main branch. Unused.

nickscamara: @mogery it was for this site, custom case:
```
"eonhealth.com":{
    defaultScraper: "fire-engine",
    params:{
      fireEngineOptions:{
        mobileProxy: true,
        method: "get",
        engine: "request",
      },
    },
  },
  ```
  
  maybe it doesnt make sense anymore?

mogery: It does not IIRC. Do try it out please

apps/api/src/scraper/scrapeURL/lib/urlSpecificParams.ts
nickscamara: It is old. No need for it.

mogery: YEETED (YOTE?)

apps/api/src/scraper/scrapeURL/index.ts
mogery: Changed success code definition and char thershold

nickscamara:
- [ ] Reminder for Nick and Rafa to test scrape events + dashboard logging

nickscamara:
- [ ] Redlock for emails, same issue in main right now. If someone is someone is doing a big crawl, it will send them multiple emails.

PR #728, opened 2024-10-03
Docs: Remove wait_until_done from python-sdk example

bytrangle:
Adding `wait_until_done` parameter to the `crawl_url` function will throw a TypeError: FirecrawlApp.crawl_url() got an unexpected keyword argument 'wait_until_done'.


nickscamara:
Thank you @bytrangle! That param was only for 0.x versions indeed!

PR #746, opened 2024-10-08
added ollama vision backup

GQAdonis:
I added a backup to use an Ollama vision model if no Anthropic or OpenAI key is provided.


nickscamara:
Thank you @GQAdonis! Will take a look at it soon!

PR #757, opened 2024-10-10
Update README.md

busaud:
I believe wait_until_done was removed as of v1?


nickscamara:
Yes! Thank you! @busaud

PR #772, opened 2024-10-12
chore: added star history graph

barbarian360:
## Star history graph 
image


nickscamara:
That's awesome @barbarian360! Thank you!

PR #773, opened 2024-10-14
Retries for ACUC RPC + Price credits fallback

nickscamara:


apps/api/src/controllers/auth.ts
```diff
@@ -68,26 +73,49 @@ export async function setCachedACUC(api_key: string, acuc: AuthCreditUsageChunk
...
+      retries++;
+      if (retries === maxRetries) {
+        throw new Error(
+          "Failed to retrieve authentication and credit usage data after 3 attempts: " +
+            JSON.stringify(error)
+        );
+      }
+
+      // Wait for a short time before retrying
+      await new Promise((resolve) => setTimeout(resolve, 200));
```
nickscamara, line 114: I was considering add expo backoff but I think this should be enough.

PR #809, opened 2024-10-22
feat: Auto Recharge Credits + Credit Packs

nickscamara:
see firecrawl-web#112 for change explanation.

### DIFF
--- a/apps/api/src/controllers/auth.ts
+++ b/apps/api/src/controllers/auth.ts
@@ -75,15 +75,19 @@ export async function setCachedACUC(
 
 export async function getACUC(
   api_key: string,
-  cacheOnly = false
+  cacheOnly = false,
+  useCache = true
 ): Promise<AuthCreditUsageChunk | null> {
   const cacheKeyACUC = `acuc_${api_key}`;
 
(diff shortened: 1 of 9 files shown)


rafaelsideguide (approved):

apps/api/src/controllers/auth.ts
```diff
@@ -75,23 +75,27 @@ export async function setCachedACUC(
 
 export async function getACUC(
   api_key: string,
-  cacheOnly = false
+  cacheOnly = false,
+  useCache = true
```
mogery, line 79: Not caching ACUC is getting rid of like, 90% of the benefit of even having ACUC.

mogery (approved):

apps/api/src/controllers/auth.ts
nickscamara: This is only used once, to prevent race conditions on charging.

apps/api/src/services/billing/stripe.ts
nickscamara, line 1: This probably needs quite a bit of work to get it right, will iterate after.

nickscamara (commented):
merging

Issue #813, opened 2024-10-23
[Feat] Authenticate webhook


rafaelsideguide:
From discord.

> I don't see any documentation around how to verify whether a webhook event actually is from FireCrawl. If this feature doesn't already exist, can you provide a way to do this? For example, here's how Ashby lets developers verify webhook events are credible:  https://developers.ashbyhq.com/docs/authenticating-webhooks

PR #820, opened 2024-10-24
Added support for webhook authentication

txrp0x9:
Adds HMAC-SHA256 verification on webhooks
If provided with a "secretKey" in the request body, the webhook response will now contain a header,
 `"Webhook-Signature": "sha256=<hash>"`
The \<hash\> being a SHA256 hash with the secretKey on the raw request body string.

Example of verifying the request body with express.js

```
app.post("/webhook", function (req, res) {
    console.log(req.headers["webhook-signature"]);
    let rawBody = "";
    req.on("data", (chunk) => {
        rawBody += chunk;
    });

    req.on("end", () => {
        console.log(rawBody);
        console.log(checkSignature(req.headers["webhook-signature"], rawBody));
        res.status(200).send("Raw body received and logged.");
    });
});

const secret = "1f29rv8owfa0w9wr4tswpg0f9897162dbfkjvdlz";
function checkSignature(signatureFromHeader, rawPayload) {
    const hmac = crypto.createHmac("sha256", secret);
    hmac.update(rawPayload);
    const digest = hmac.digest("hex");
    return "sha256=" + digest === signatureFromHeader;
}
```

Make sure to use the body string before it is parsed to JSON (JSON.parse() or equivalent).

### DIFF
--- a/apps/api/src/controllers/v1/types.ts
+++ b/apps/api/src/controllers/v1/types.ts
@@ -195,7 +195,8 @@ export const crawlRequestSchema = crawlerOptions.extend({
   secretKey: z
     .string()
     .min(32, "The secret key must be at least 32 characters long.")
-    .max(64, "The secret key must not exceed 64 characters."),
+    .max(64, "The secret key must not exceed 64 characters.")
+    .optional(),
   limit: z.number().default(10000),
 }).strict(strictMessage);
 
--- a/apps/api/src/controllers/v1/types.ts
+++ b/apps/api/src/controllers/v1/types.ts
@@ -192,6 +192,10 @@ export const crawlRequestSchema = crawlerOptions.extend({
   origin: z.string().optional().default("api"),
   scrapeOptions: scrapeOptions.omit({ timeout: true }).default({}),
   webhook: z.string().url().optional(),
+  secretKey: z
+    .string()
+    .min(32, "The secret key must be at least 32 characters long.")
+    .max(64, "The secret key must not exceed 64 characters."),
   limit: z.number().default(10000),
 }).strict(strictMessage);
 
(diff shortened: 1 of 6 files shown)


txrp0x9:
While this solution does mimic the version described in the #813, It can be computationally expensive.
A better solution might just be passing the secret in the response body, instead of hashing it (the secret can be a uuid)
Why? If the secret is leaked, the authentication is broken, so as long as the secret remains safe, that is, the client and server are the only ones who know it, hashing the request body may not be necessary.
Unless the request body is susceptible to MITM attacks which can lead to the body being changed, but with HTTPS this should be less of a worry (not non-existent however)

I'd appreciate more input on this matter

nickscamara:
@rafaelsideguide can you review this when you get a chance? tks!

apps/api/src/controllers/v1/types.ts
```diff
@@ -192,6 +192,9 @@ export const crawlRequestSchema = crawlerOptions.extend({
   origin: z.string().optional().default("api"),
   scrapeOptions: scrapeOptions.omit({ timeout: true }).default({}),
   webhook: z.string().url().optional(),
+  verification_token: z
```
rafaelsideguide: @txrp0x9 camelCase for req params pls

rafaelsideguide: and for the responses too. besides that I think it's good to merge

rafaelsideguide (changes requested):

rafaelsideguide (changes requested):
the response still using snake case

txrp0x9:
Renaming seems to have broken something, results have become unpredictable, I'll fix this in a moment.

Nurfen:
Plus one for this update, we can't use webhooks without basic security features

PR #828, opened 2024-10-27
Links are now extracted after applying excludeTags

### LINKED ISSUE #701: [Bug] `format: ["links"]` doesn't respect `excludeTags`
[Discord thread](https://discord.com/channels/1226707384710332458/1287903778112278610)

```js
const INPUT_URL = "https://www.jndla.com/cases/class-action-administration"

const response = await app.scrapeUrl(INPUT_URL, {
  formats: ["links"],
  excludeTags: ["a"],
})

console.log(response.links)
// this still has links, even though we excluded the <a> tags
```

txrp0x9:
fixes #701

### DIFF
--- a/apps/api/src/scraper/WebScraper/single_url.ts
+++ b/apps/api/src/scraper/WebScraper/single_url.ts
@@ -447,7 +447,7 @@ export async function scrapSingleUrl(
     let linksOnPage: string[] | undefined;
 
     if (pageOptions.includeLinks) {
-      linksOnPage = extractLinks(rawHtml, urlToScrap);
+      linksOnPage = extractLinks(html, urlToScrap);
     }
 
     let document: Document = {


apps/api/src/scraper/WebScraper/single_url.ts
```diff
@@ -447,7 +447,7 @@ export async function scrapSingleUrl(
     let linksOnPage: string[] | undefined;
 
     if (pageOptions.includeLinks) {
-      linksOnPage = extractLinks(rawHtml, urlToScrap);
+      linksOnPage = extractLinks(html, urlToScrap);
```
nickscamara, line 450: I think the problem with this is that we rely on this function for our /crawl, which will end up failling to grab all the links if we don't pass the raw version

nickscamara: @mogery lmk if im wrong

txrp0x9: A global code search did not reveal any usage of the `linksOnPage` field anywhere other than api return
I believe the crawler uses a separate extractLinksFromHTML function
https://github.com/mendableai/firecrawl/blob/8a4f4cb9d98884bc70f4cf188a2c4dc87f656462/apps/api/src/services/queue-worker.ts#L374 and 
https://github.com/mendableai/firecrawl/blob/8a4f4cb9d98884bc70f4cf188a2c4dc87f656462/apps/api/src/scraper/WebScraper/crawler.ts#L382
defined as
https://github.com/mendableai/firecrawl/blob/8a4f4cb9d98884bc70f4cf188a2c4dc87f656462/apps/api/src/scraper/WebScraper/crawler.ts#L322-L337

nickscamara: Right makes sense.

PR #831, opened 2024-10-28
chore(README.md): use `satisfies` instead of `as` for ts example

twlite:
When we use `as`, typescript does not validate the types properly in scenarios like the typescript example. The `satisfies` operator is stricter and **does** validate the types, preventing misuse like such:

Using blank object
```ts
interface CrawlParams {
  required: boolean
}

// This is valid
const config = {} as CrawlParams

// this is invalid because 'required' field is missing
const config = {} satisfies CrawlParams
// ^ Type '{}' does not satisfy the expected type 'CrawlParams'. Property 'required' is missing in type '{}' but required in type 'CrawlParams'.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -483,7 +483,7 @@ const crawlResponse = await app.crawlUrl('https://firecrawl.dev', {
   scrapeOptions: {
     formats: ['markdown', 'html'],
   }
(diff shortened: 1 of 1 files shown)


nickscamara:
@twlite Great catch! Thank you!

PR #842, opened 2024-10-29
fix(api): use availableParallelism method

samir-byte:
This method returns an estimate of the default amount of parallelism a program should use, compared to `cpus().length` which returns the cpu core count.


ftonato:
Hello @samir-byte,

Thanks for your contribution ;)

ftonato (approved):

PR #843, opened 2024-10-29
refactor: use native uuid generator.

ghimirebibek:


rafaelsideguide (changes requested):
Is it possible to remove the uuid module entirely?

ghimirebibek:
> Is it possible to remove the uuid module entirely?

I've added a regex validation in place of uuid package's validate method. I've also removed uuid from the dependencies.

PR #844, opened 2024-10-29
refactor: consistent dotenv import

sulav7:


mogery (changes requested):
Breaks everything. See CI logs.

PR #855, opened 2024-10-31
[feat] Iframe support

tomkosm:
Support iframes with inline data


rafaelsideguide:
Tested and benchmarked it :D

PR #867, opened 2024-11-04
[Feat] Added remove base64 images options (true by default)

rafaelsideguide:
Base64 images are huge and this is causing multiple issues


apps/api/src/scraper/WebScraper/utils/removeBase64Images.ts
nickscamara, line 1: @mogery regex good?

mogery: yes!

mogery (approved):

PR #878, opened 2024-11-08
feat(crawl): Similar URL deduplication

mogery:
This PR adds some important features:
 - If a crawl's scrape was redirected, the target URL is also locked, preventing it from being scraped again later.
 - `lockURL` now generates similar URLs that will presumably return the same content, and it will lock all of them, preventing the same content from being scraped twice. This can be disabled with the `deduplicateSimilarURLs` crawl parameter (on by default).
 - Added missing parameters to the JS SDK.

### DIFF
--- a/apps/api/src/lib/crawl-redis.ts
+++ b/apps/api/src/lib/crawl-redis.ts
@@ -97,6 +97,37 @@ export function normalizeURL(url: string): string {
     return urlO.href;
 }
 
+export function generateURLPermutations(url: string | URL): URL[] {
+    const urlO = new URL(url);
+
+    // Construct two versions, one with www., one without
+    const urlWithWWW = new URL(urlO);
+    const urlWithoutWWW = new URL(urlO);
+    if (urlO.hostname.startsWith("www.")) {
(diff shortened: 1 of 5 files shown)


apps/api/src/lib/crawl-redis.ts
```diff
@@ -90,23 +90,56 @@ export async function getThrottledJobs(teamId: string): Promise<string[]> {
...
+
 export async function lockURL(id: string, sc: StoredCrawl, url: string): Promise<boolean> {
     if (typeof sc.crawlerOptions?.limit === "number") {
         if (await redisConnection.scard("crawl:" + id + ":visited") >= sc.crawlerOptions.limit) {
             return false;
         }
     }
 
-    try {
+    url = normalizeURL(url);
```
nickscamara, line 138: what happen if this throws an error?

nickscamara:
Worth adding tests to the dedup stuff

apps/api/src/lib/crawl-redis.ts
mogery: job fails, but it shouldn't unless the URL is invalid anyways which is a pretty good reason to fail

PR #880, opened 2024-11-08
[SDK] Added next handler for python sdk (js is ok)

rafaelsideguide:


nickscamara:
We already have a `next` handler in the crawling itself. No?

```python
while 'next' in status_data:
                          status_response = self._get_request(status_data['next'], headers)
                          status_data = status_response.json()
                          data.extend(status_data['data'])
```

nickscamara:
The `check_crawl_status` is not even used anymore

rafaelsideguide:
@nickscamara User on discord tried to do `async_crawl_url` for a big page and retrieve the data with `check_crawl_status` (as we show on docs https://docs.firecrawl.dev/sdks/python) and the SDK didn't handle its not possible to get more than 10Mb. He had to write a function himself for that.

nickscamara:
Oh I see. Makes sense. Thanks @rafaelsideguide

nickscamara:
- [ ] Publish patch

banaglitch:
@rafaelsideguide I'm trying to figure out if check_crawl_status is usefull for me. But I think there is an error with the call to status_data.pop('next',None) which prevents the while to get any more next url than one of the crawl job no?

PR #893, opened 2024-11-11
fix(crawler): relative URL handling on non-start pages

### LINKED ISSUE #821: [Bug/Investigation] Some relative paths are incomplete
When crawling https://docs.cleanlab.ai, it looks like the URLs are incomplete. For example, this URL in the response:
https://docs.cleanlab.ai/cleanlab/token_classification/index.html
is indeed a 404 page, but it should actually be:
https://docs.cleanlab.ai/stable/cleanlab/token_classification/index.html.

This issue seems to occur because the elements on the page use relative paths like `<a class="reference internal" href="../../tutorials/token_classification.html">Token Classification (text)</a>`.

mogery:
Fixes #821 

Went for an easier fix. Just fixes the logic when adding relative URLs to the crawl from the site content. Was basing the `new URL` off of the wrong base URL.

### DIFF
--- a/apps/api/src/lib/crawl-redis.ts
+++ b/apps/api/src/lib/crawl-redis.ts
@@ -166,10 +166,10 @@ export async function lockURLs(id: string, sc: StoredCrawl, urls: string[]): Pro
     return res;
 }
 
-export function crawlToCrawler(id: string, sc: StoredCrawl): WebCrawler {
+export function crawlToCrawler(id: string, sc: StoredCrawl, initialUrl?: string): WebCrawler {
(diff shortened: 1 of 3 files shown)


mogery:
My bad, forgot about that case. Will be fixing after standup.

mogery:
@rafaelsideguide should work now, retest pls

rafaelsideguide:
Looking good. 1354 pages crawled for this url now (no 404s apparently :D) let's merge it!

rafaelsideguide (approved):

PR #895, opened 2024-11-13
Redlock for sending email notifications

nickscamara:
Prevents multiple emails from being sent.


apps/api/src/services/notification/email_notification.ts
```diff
@@ -88,6 +89,7 @@ export async function sendNotificationInternal(
   if (team_id === "preview") {
     return { success: true };
   }
+  return await redlock.using([`notification-lock:${team_id}:${notificationType}`], 5000, async () => {
```
mogery, line 92: Good call using `using` 👍🏻

mogery (approved):

PR #910, opened 2024-11-19
Pin dependencies to avoid conflict

amrit073:
Pinning the dependencies because Firecrawl was experiencing issues with WebSockets.
Reference: [WebSockets Documentation - Extra Headers](https://websockets.readthedocs.io/en/stable/howto/upgrade.html#extra-headers-additional-headers)


nickscamara:
Thaanks @amrit073!

nickscamara:
@rafaelmmiller can you double check this? tks!

PR #931, opened 2024-11-29
Use correct list typing for py 3.8 support

niazarak:
list[str] is not supported in Python 3.8


nickscamara:
Thank you @niazarak!

nickscamara:
@rafaelsideguide can you double check this one? tks

PR #964, opened 2024-12-11
chore: update contributing guidelines, improve environment setup, and refactor search functionality

ftonato:
- Enhanced the CONTRIBUTING.md file for clarity on environment variable setup and installation instructions.
- Updated package.json to use npx for prettier formatting.
- Refactored search functionality to implement a provider pattern for different search engines, including Google, Serper, and SearchAPI.
- Removed deprecated Google search implementation and related files.
- Improved error handling and logging in search functions.
- Added tests for the new Google search result parser.
- Introduced a new provider factory to streamline search provider creation.


nickscamara:
This is awesome @ftonato! Thank you! You just reminded me that we had to fix our prettier settings.

apps/api/package.json
nickscamara, line 1: No need for this anymore, added prettier as dependency

nickscamara (approved):
awesome work! Thanks!

apps/api/package.json
ftonato: I would say we can keep it here, as it will be removed once we merge #969.

Unless you insist that this should be removed in this PR, I can easily do it anyway.

Boomtokn (approved):

PR #969, opened 2024-12-12
chore: add husky and lint-staged for pre-commit hooks (prettier)

ftonato:
Hello,

As noted in my previous pull request #964, we are not running the `format` script, which has resulted in many files not being properly formatted (non-prettiered). _Can we normalize the idea that "prettier" is a verb? 😅_

To address this, I decided to add Husky to run `prettier --write` on all _important_ files in the `apps/api/src` directory that are staged.

---

**Would you like to see it in action? Your wish will be heard 🧞**

https://github.com/user-attachments/assets/ea3225f2-dcc2-432e-8bbb-a1c8daa59678

> [!TIP]
> Read the comments in the code 🔥

### DIFF
--- a/apps/api/.husky/pre-commit
+++ b/apps/api/.husky/pre-commit
@@ -1,0 +1,2 @@
+cd apps/api
+pnpm lint-staged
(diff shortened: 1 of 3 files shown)


mogery (approved):
need to update for the changes inbetween but we should add this

PR #977, opened 2024-12-14
fix: adjust Playwright service response to match API schema expectations

AsyncArtisan:
### Problem
The Playwright microservice was returning `pageError: false` for successful scrapes, but the main API expects `pageError` to be either a string or undefined (as defined by the Zod schema). This schema mismatch caused Playwright result parsing to fail silently, making the main API fall back to fetch even when Playwright successfully scraped the page.

### Solution
Modified the Playwright service response to:
1. Return `undefined` instead of `false` for successful scrapes (status code 200)
2. Only include `pageError` in the response when there is an actual error
3. Use spread operator to conditionally include pageError: `...(pageError && { pageError })`

### Testing
- Tested with local setup running both services
- Successfully scraped test URLs with Playwright without falling back to fetch
- Confirmed proper error handling for failed scrapes
- Verified logs showing successful Playwright scrapes being properly handled by the main API

### DIFF
--- a/apps/playwright-service-ts/api.ts
+++ b/apps/playwright-service-ts/api.ts
@@ -196,7 +196,7 @@ app.post('/scrape', async (req: Request, res: Response) => {
     }
   }
 
-  const pageError = pageStatusCode !== 200 ? getError(pageStatusCode) : false;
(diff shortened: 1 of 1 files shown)


nickscamara:
Good catch! Thank you @AsyncArtisan!

PR #990, opened 2024-12-17
feat(python-sdk): Make API key optional for self-hosted instances

RutamBhagat:
Fixes: https://github.com/mendableai/firecrawl/issues/961

Description:
Make API key validation conditional based on API URL. API key is now only required when using cloud service (api.firecrawl.dev). This allows easier self-hosting by removing API key requirement for local/self-hosted instances.

Changes:
- Modified API key validation logic in FirecrawlApp
- Updated tests to handle both cloud and self-hosted scenarios
- Added missing test dependencies

### DIFF
--- a/apps/python-sdk/pyproject.toml
+++ b/apps/python-sdk/pyproject.toml
@@ -13,7 +13,6 @@ dependencies = [
     "python-dotenv",
     "websockets",
     "nest-asyncio",
-    "pytest>=8.3.4",
     "pydantic>=2.10.3",
 ]
 authors = [{name = "Mendable.ai",email = "nick@mendable.ai"}]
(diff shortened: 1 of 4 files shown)


apps/python-sdk/pyproject.toml
```diff
@@ -12,7 +12,9 @@ dependencies = [
     "requests",
     "python-dotenv",
     "websockets",
-    "nest-asyncio"
+    "nest-asyncio",
+    "pytest>=8.3.4",
```
nickscamara: we might not need pytest here, @rafaelmmiller lmk.

RutamBhagat: I have removed pytest from pyproject.toml but while testing i get this error, thats why I added it

image

nickscamara: Oh i see

PR #1002, opened 2024-12-21
handle .pdf links that are html

vickeeyuan:


mogery:
Hi there! Unfortunately this is not a valid way to handle this case, as `HEAD` is not implemented by many servers. A different fix for this is in the works.

PR #1018, opened 2024-12-27
[FIR-37] feat: extract and return favicon URL during scraping

ftonato:
Hello,

We cannot simply select the element using `'link[rel="icon"]'` because some websites (e.g., [_https://www.mendable.ai/_](https://www.mendable.ai/)) use tags like `<link href="/favicon.ico" rel="shortcut icon">` where the `rel` property includes additional details. To address this, we filter all `<link>` tags that contain `icon` in their `rel` attribute and select the first matching one.

----

**Preview without `favicon`**

image

----

**Preview with `favicon` where `rel="icon"`**

image

----

**Preview with `favicon` where `rel="whatever-value icon"`**

image


nickscamara:
Looks great! Thanks @ftonato!

PR #1019, opened 2024-12-28
[FIR-125] feat(scrapeURL): add encodeRawHTML transformer for charset handling

ftonato:


apps/api/src/scraper/scrapeURL/transformers/index.ts
```diff
@@ -31,6 +31,35 @@ export function deriveMetadataFromRawHTML(
...
+  try {
+    // Convert the response data if charset is not UTF-8
+    if (charset.toUpperCase() !== "UTF-8") {
+      const decoder = new TextDecoder(charset);
+      document.rawHtml = decoder.decode(data);
+      
+      return document;
+    }
+  } catch (error) {
+    throw new Error("Failed to convert rawHtml to UTF-8");
```
mogery, line 57: Always add the source error to the error object, so we can see the true error in the logs.
```suggestion
    throw new Error("Failed to convert rawHtml to UTF-8", { cause: error });
```

apps/api/src/scraper/scrapeURL/transformers/index.ts
```diff
@@ -31,6 +31,35 @@ export function deriveMetadataFromRawHTML(
...
+  const charset = charsetMatch ? charsetMatch[1] : "UTF-8";
+
+  try {
+    // Convert the response data if charset is not UTF-8
+    if (charset.toUpperCase() !== "UTF-8") {
+      const decoder = new TextDecoder(charset);
+      document.rawHtml = decoder.decode(data);
+      
+      return document;
+    }
```
mogery, line 55: The approach is good, but I think we need to move it to earlier parts of the stack, since the UTF-8 misencode is too destructive to get usable Shift_JIS out of just the parsed string. We need to move it up to a part of the code where we still have access to the raw buffer coming out of the network stack. Let's talk about this on Monday.

mogery (changes requested):

Issue #1011, opened 2024-12-23
[Bug] Firebase error continuation


nickscamara:
https://github.com/mendableai/firecrawl/pull/997/

Giving errors when building for cjs

nickscamara:
Error: Build failed with 1 error:
src/index.ts:12:48: ERROR: Top-level await is currently not supported with the "cjs" output format
    at addChunk (node:internal/streams/readable:368:12)```

RutamBhagat:
@nickscamara I notice this issue references CJS compatibility, but the build is currently working in both CJS and ESM formats. Could you provide the Steps to reproduce the error

```bash
npm run build           

> @mendable/firecrawl-js@1.10.1 build
> tsup

CLI Building entry: src/index.ts
CLI Using tsconfig: tsconfig.json
CLI tsup v8.2.4
CLI Using tsup config: /home/voldemort/Desktop/Code/open_source/firecrawl/apps/js-sdk/firecrawl/tsup.config.ts
CLI Target: es2022
CLI Cleaning output folder
CJS Build start
ESM Build start
ESM dist/index.js 22.59 KB
ESM ⚡️ Build success in 22ms
CJS dist/index.cjs 24.08 KB
CJS ⚡️ Build success in 22ms
DTS Build start
DTS ⚡️ Build success in 1533ms
DTS dist/index.d.cts 15.48 KB
DTS dist/index.d.ts  15.48 KB
```

PR #1021, opened 2024-12-28
fix(js-sdk): Add CJS-compatible WebSocket loading

RutamBhagat:
Fixes: #1011

The build passes for both ESM and CJS
```bash
npm run build

> @mendable/firecrawl-js@1.10.1 build
> tsup

CLI Building entry: src/index.ts
CLI Using tsconfig: tsconfig.json
CLI tsup v8.2.4
CLI Using tsup config: /home/voldemort/Desktop/Code/open_source/firecrawl/apps/js-sdk/firecrawl/tsup.config.ts
CLI Target: es2022
CLI Cleaning output folder
CJS Build start
ESM Build start
ESM dist/index.js 22.21 KB
ESM ⚡️ Build success in 17ms
CJS dist/index.cjs 24.08 KB
CJS ⚡️ Build success in 17ms
DTS Build start
DTS ⚡️ Build success in 1540ms
DTS dist/index.d.cts 15.45 KB
DTS dist/index.d.ts  15.45 KB
```

### DIFF
--- a/apps/js-sdk/firecrawl/src/index.ts
+++ b/apps/js-sdk/firecrawl/src/index.ts
@@ -1069,8 +1069,9 @@ export class CrawlWatcher extends TypedEventTarget<CrawlWatcherEvents> {
     }).bind(this);
   }
 
-  async close() {
-    await this.initialized;
-    this.ws.close();
+  close() {
+    this.initialized.then(() => {
+      this.ws.close();
+    });
   }
 }
--- a/apps/js-sdk/firecrawl/src/index.ts
+++ b/apps/js-sdk/firecrawl/src/index.ts
@@ -1,7 +1,15 @@
 import axios, { type AxiosResponse, type AxiosRequestHeaders, AxiosError } from "axios";
 import type * as zt from "zod";
 import { zodToJsonSchema } from "zod-to-json-schema";
+import { WebSocket as IsowsWebSocket } from "isows";
 import { TypedEventTarget } from "typescript-event-target";
+
+
+let WebSocket: typeof IsowsWebSocket;
+
+if (typeof __WEBSOCKET_LOADER__ === 'string') {
+  WebSocket = new Function('return ' + __WEBSOCKET_LOADER__)();
+}
 
 /**
  * Configuration interface for FirecrawlApp.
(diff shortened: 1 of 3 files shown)


apps/js-sdk/firecrawl/src/index.ts
```diff
@@ -947,10 +947,23 @@ export class CrawlWatcher extends TypedEventTarget<CrawlWatcherEvents> {
   public status: CrawlStatusResponse["status"];
   public id: string;
 
+  private static loadWebSocket() {
+    try {
+      return require('isows').WebSocket;
+    } catch {
+      try {
+        return import('isows').then(m => m.WebSocket);
```
mogery: Missing an await here, right? Either here or on line 964.

mogery (changes requested):

apps/js-sdk/firecrawl/src/index.ts
RutamBhagat: Thanks for catching that! I've updated the code to properly handle the WebSocket loading. Could you please review the changes again?

mogery: This can cause a race condition. I'd prefer importing it lazily in an already-asynchronous function, like the watch crawl endpoint's function.

mogery (changes requested):

apps/js-sdk/firecrawl/src/index.ts
```diff
@@ -942,18 +941,39 @@ interface CrawlWatcherEvents {
...
+  }
+
   constructor(id: string, app: FirecrawlApp) {
     super();
     this.id = id;
-    this.ws = new WebSocket(`${app.apiUrl}/v1/crawl/${id}`, app.apiKey);
     this.status = "scraping";
     this.data = [];
 
+    this.initialized = CrawlWatcher.loadWebSocket()
```
tomkosm: I dont think this is good enough, cause ws is not initialized when the constructor is called but later async, so there can be a race condition.

tomkosm (changes requested):
Thank you for the work you've put into this! I believe the best approach here would be to use top-level await for ESM, while keeping the current implementation for CJS. If you're open to merging this, it might be worth exploring how to configure tsup to use distinct implementations for ESM and CJS.

I really appreciate your efforts on this, and I'm happy to assist further if needed. Let me know how I can help!

RutamBhagat:
> Thank you for the work you've put into this! I believe the best approach here would be to use top-level await for ESM, while keeping the current implementation for CJS. If you're open to merging this, it might be worth exploring how to configure tsup to use distinct implementations for ESM and CJS.
> 
> I really appreciate your efforts on this, and I'm happy to assist further if needed. Let me know how I can help!

tsup builds both formats from the same source file. I tried an if else at the top level with top level await only for ESM, but the CJS compilation kept failing because it sees the top-level await during parsing, but my current approach injects format-specific code at build time using tsup's define feature

tomkosm (changes requested):
Not a fan of that approach, is there no way with tsup to have a different module for cjs and esm? Also we should take into account the the import of isows might fail if the platform doesnt have ws support.

mogery:
Hi Rutam, we're going a different way to fix this, so I'm closing this PR. Thanks for the work.

PR #1038, opened 2025-01-05
feat: use new random user agent instead of the old one

1101-1:
The `random-useragent` package (https://github.com/skratchdot/random-useragent) hasn't been updated for three years and in it repo user agents set manually via json file, therefore they all are easily detected (because old).

so it would be better to add a new modern package that generates different actual user agents daily (https://github.com/intoli/user-agents)


nickscamara:
Thanks @1101-1!

PR #1044, opened 2025-01-06
(feat/extract) Move extract to a queue system

nickscamara:


nickscamara:
This uses a different queue but we can merge it to be on 1 queue only if preferred @mogery

nickscamara:
- [ ] Let beta users know that we are moving it to a queue system

apps/api/src/controllers/v1/extract-status.ts
```diff
@@ -0,0 +1,53 @@
...
+export async function extractStatusController(
+  req: RequestWithAuth<{ jobId: string }, any, any>,
+  res: Response,
+) {
+  try {
+    const rateLimiter = scrapeStatusRateLimiter;
+    const incomingIP = (req.headers["x-forwarded-for"] ||
+      req.socket.remoteAddress) as string;
+    const iptoken = incomingIP;
+    await rateLimiter.consume(iptoken);
```
mogery: Why is this here? This is handled by the authMiddleware.

apps/api/src/controllers/v1/extract-status.ts
```diff
@@ -0,0 +1,53 @@
...
+  supabaseGetJobsById,
+} from "../../lib/supabase-jobs";
+import { scrapeStatusRateLimiter } from "../../services/rate-limiter";
+import { RequestWithAuth } from "./types";
+
+export async function extractStatusController(
+  req: RequestWithAuth<{ jobId: string }, any, any>,
+  res: Response,
+) {
+  try {
```
mogery: You must not wrap v1 endpoint controllers in a try-catch. Error handling is done on the Express middleware level.

apps/api/src/controllers/v1/extract-status.ts
```diff
@@ -0,0 +1,53 @@
...
+      });
+    }
+
+    const jobData = await supabaseGetJobsById([req.params.jobId]);
+    if (!jobData || jobData.length === 0) {
+      return res.status(404).json({
+        success: false,
+        error: "Job not found",
+      });
+    }
```
mogery: Why request the DB twice?

mogery (approved):
Rest is good. Though I think right now while it's in progress it will just return Job not found, which is suboptimal.

mogery (changes requested):
Meant to RC not approve

apps/api/src/lib/extract/extraction-service.ts
```diff
@@ -38,9 +39,8 @@ function getRootDomain(url: string): string {
   }
 }
 
-export async function performExtraction(options: ExtractServiceOptions): Promise<ExtractResult> {
+export async function performExtraction(extractId, options: ExtractServiceOptions): Promise<ExtractResult> {
```
mogery: no type here

mogery (approved):
good!

apps/api/src/lib/extract/extraction-service.ts
mogery:
```suggestion
export async function performExtraction(extractId: string, options: ExtractServiceOptions): Promise<ExtractResult> {
```

mogery (approved):
nice

PR #1045, opened 2025-01-06
(feat/conc) Move fully to a concurrency limit system

nickscamara:


apps/api/src/services/notification/email_notification.ts
```diff
@@ -183,3 +191,89 @@ export async function sendNotificationInternal(
...
+  ) => {
+    const now = new Date();
+    const pastDate = new Date(now.getTime() - daysBetweenEmails * 24 * 60 * 60 * 1000);
+
+    const { data: recentNotifications, error: recentNotificationsError } = await supabase_service
+      .from("user_notifications")
+      .select("*")
+      .eq("team_id", team_id)
+      .eq("notification_type", notificationType)
+      .gte("sent_date", pastDate.toISOString());
```
mogery, line 228: Nonononononono. Absolutely not. We are not doing a DB query on every single CC job insert over 2x. This must be moved to Redis.

mogery (changes requested):
Rest looks good. I'll Redis-ify it later today.

apps/api/src/services/notification/email_notification.ts
nickscamara: you're right.. didn't realize how often that hits

mogery:
@nickscamara moved to redis, pls check again

nickscamara:
Might be wise to keep the db check after redis in case the redis restarts(?)

mogery (approved):

mogery (approved):

PR #1053, opened 2025-01-12
Fix bad WebSocket URL in CrawlWatcher

ProfHercules:
Fixed CrawlWatcher creating WebSocket using standard http url from base app.


mogery (approved):
thank you!

ProfHercules:
Could improve this by using a regex to replace either `http` or `https` with `ws` (only at the start of the string).

mogery (approved):
Awesome!

PR #1069, opened 2025-01-16
Add ENV environment variable to docker compose file

### LINKED ISSUE #930: [Self-Host] Max retries local error
I am trying to self-host Firecrawl, but am running into a weird error. Here is my code: 
`
from firecrawl import FirecrawlApp

app = FirecrawlApp(api_key="hi", api_url="http://localhost:3002")

#Crawl a website:
crawl_status = app.crawl_url(
  'https://www.scu.edu/engineering/', 
  params={
    'limit': 500, 
    'scrapeOptions': {'formats': ['markdown', 'html']}
  },
  poll_interval=100
)
print(crawl_status)
`
I get the following error after it runs for a bit:
`
urllib3.exceptions.MaxRetryError: HTTPSConnectionPool(host='localhost', port=3002): Max retries exceeded with url: /v1/crawl/8db68f10-fbf2-4911-a806-8faa06f478d2?skip=221 (Caused by SSLError(SSLError(1, '[SSL] record layer failure (_ssl.c:1020)')))
`

If I decrease the limit to 100 pages, everything works fine. Any tips on how to fix this? Thank you!

aanokh:
Add ENV environment variable to the docker compose file so setting ENV=local in the .env will work properly. fixes #930

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -46,6 +46,7 @@ services:
       SELF_HOSTED_WEBHOOK_URL: ${SELF_HOSTED_WEBHOOK_URL}
       LOGGING_LEVEL: ${LOGGING_LEVEL}
       FLY_PROCESS_GROUP: app
+      ENV: ${ENV}
     depends_on:
       - redis
       - playwright-service


aanokh:
I think this is the more correct approach than #765

mogery (commented):
Hi! This is also invalid, since we use `ENV=production` both for local development of a prod target and for staging, neither of which have HTTPS support.

PR #1072, opened 2025-01-18
(feat/formats) Extract format renamed to json format

nickscamara:


nickscamara:
- [ ] Update the docs

PR #1073, opened 2025-01-19
(feat/index) Index/Insertion queue

nickscamara:
Add another queue and worker for batch insertion operations in the db such as indexing crawl_maps


mogery (approved):
LGTM

PR #1074, opened 2025-01-20
Enhance extraction service with caching capabilities for testing purposes

tomkosm:


apps/api/src/lib/extract/extraction-service.ts
```diff
@@ -184,7 +195,7 @@ export async function performExtraction(
       error:
         "No valid URLs found to scrape. Try adjusting your search criteria or including more URLs.",
       extractId,
-      urlTrace: urlTraces,
+      urlTraces: urlTraces,
```
rafaelsideguide: why?

apps/api/src/lib/extract/extraction-service.ts
```diff
@@ -411,66 +443,133 @@ export async function performExtraction(
...
     if (docsMap.size == 0) {
       // All urls are invalid
       return {
         success: false,
-        error: "All provided URLs are invalid. Please check your input and try again.",
+        error:
+          "All provided URLs are invalid. Please check your input and try again.",
         extractId,
-        urlTrace: request.urlTrace ? urlTraces : undefined,
+        urlTraces: request.urlTrace ? urlTraces : [],
```
rafaelsideguide: shouldn't return empty array

rafaelsideguide: shouldn't change the response format

apps/api/src/lib/extract/extraction-service.ts
```diff
@@ -411,66 +443,133 @@ export async function performExtraction(
...
+}
+export async function cacheGetDocsAndSchemas(
+  extractId: string,
+  options: ExtractServiceOptions,
+) {
+  const { request } = options;
+  const { cacheKey, cacheMode } = request;
+  const filePath = path.join(__dirname, "./cache", `${cacheKey}.json`);
+
+  if (cacheMode === "direct") {
```
rafaelsideguide: this should be default, not inside if

rafaelsideguide (changes requested):

rafaelsideguide:
I tested with the request.http example and it returned internal server error

tomkosm:
> I tested with the request.http example and it returned internal server error
> 
> ```
> POST {{baseUrl}}/v1/extract HTTP/1.1
> Authorization: Bearer {{$dotenv TEST_API_KEY}}
> Content-Type: application/json
> 
> {
>   "cacheMode":"load",
>   "cacheKey":"testt",
>   "urls": [
>   ...
> ```
> 
> ```
> {
>  "success": false,
>  "error": "Internal server error"
> }
> ```

Did you first run it with save?

apps/api/src/lib/extract/extraction-service.ts
tomkosm: fixed

apps/api/src/lib/extract/extraction-service.ts
tomkosm: it is default cause default value is "direct", made it more clear tho

apps/api/src/lib/extract/extraction-service.ts
tomkosm: fixed

PR #1086, opened 2025-01-24
Rerank with lower threshold + back to map if length = 0

rafaelsideguide:
FIR-666

### DIFF
--- a/apps/api/src/lib/extract/build-prompts.ts
+++ b/apps/api/src/lib/extract/build-prompts.ts
@@ -35,5 +35,5 @@ export function buildRerankerSystemPrompt(): string {
 }
 
 export function buildRerankerUserPrompt(searchQuery: string): string {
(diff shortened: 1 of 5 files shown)


apps/api/src/lib/extract/build-prompts.ts
```diff
@@ -35,5 +35,5 @@ export function buildRerankerSystemPrompt(): string {
 }
 
 export function buildRerankerUserPrompt(searchQuery: string): string {
-  return `Given these URLs and their content, identify which ones are relevant to the user's extraction request: "${searchQuery}". Return an array of relevant links with their relevance scores (0-1). Higher scores should be given to URLs that directly address the user's extraction request. Be very mindful with the links you select, as if they are not that relevant it may affect the quality of the extraction. Only include URLs that have a relvancy score of 0.6+.`;
+  return `Given these URLs and their content, identify which ones are relevant to the user's extraction request: "${searchQuery}". Return an array of relevant links with their relevance scores (0-1). Higher scores should be given to URLs that directly address the user's extraction request. Be very mindful with the links you select, as if they are not that relevant it may affect the quality of the extraction.`;
```
mogery, line 38: Might be worth rewriting the prompt a little to have it just assign a score instead of selecting? Would give us more control of what we include.

nickscamara:
Evals look good. Around ~4% better in most cases.

PR #1101, opened 2025-01-28
(feat/extract) Add sources to the extraction

nickscamara:


nickscamara:
Did it seperate from the helper methods to not introduce bugs for now

apps/api/src/lib/extract/extraction-service.ts
```diff
@@ -439,7 +435,7 @@ export async function performExtraction(
           //   return null;
           // }
 
-          return multiEntityCompletion.extract;
+          return null;
```
mogery, line 438: seems odd, sure this is right?

mogery (approved):
i don't fully understand it, but didn't see any obvious stepping on rakes here

apps/api/src/lib/extract/extraction-service.ts
nickscamara: Returned early now

mogery: oh okay

nickscamara:
Under a __experimental flag for now

PR #1105, opened 2025-01-28
feat(v1/map): Timeout parameter (FIR-393)

mogery:
Accidentally applied to main first, but caught it in time and reverted. Then reverted again to make this PR. Adds a timeout to map + tests.


rafaelsideguide:
@mogery, is there a way to stop the getMapResults promise from continuing processing? If it's too complex, I think it’s okay.

rafaelsideguide (commented):
@mogery is there a way to stop the getMapResults promise to continue processing? if it's too complex, I think its ok

mogery:
good idea!!!!

PR #1106, opened 2025-01-28
feat(scrapeUrl/fire-engine): add blockAds flag (FIR-692)

mogery:
waiting for f-e side fixes and adding tests


mogery:
thomas fixed it, added tests

PR #1108, opened 2025-01-29
(feat/extract) Fixes analyzeSchemaAndPrompt output did not match the schema

nickscamara:
closes FIR-662

### DIFF
--- a/apps/api/src/lib/extract/completions/analyzeSchemaAndPrompt.ts
+++ b/apps/api/src/lib/extract/completions/analyzeSchemaAndPrompt.ts
@@ -6,6 +6,7 @@ import {
   buildAnalyzeSchemaUserPrompt,
 } from "../build-prompts";
 import OpenAI from "openai";
+import { logger } from "../../../lib/logger";
 const openai = new OpenAI();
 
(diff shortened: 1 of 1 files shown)


mogery:
Is this really a fix though? It gets rid of the error, but it band-aids and can lead to bad results.

mogery (changes requested):
Is this really a fix though? It gets rid of the error, but it can lead to bad results which is even harder for us to detect.

Can we add a retry on the LLM at least?

mogery:
Also I can help you untangle this from the other PR if you want.

mogery:
Is it not? It's returning isMultiEntity but the multiEntityKeys array is empty.

nickscamara:
Right, but I feel like if that's the case we should at least log and try the default route instead of erroring out

mogery:
fair

mogery:
we can merge this for now to have some better logs

mogery (approved):

nickscamara:
Lets rename this tasks and have it as a sub task in Linear so we dont forget

mogery:
did it

PR #1111, opened 2025-01-29
feat(v0): store v0 users (team ID) in Redis for collection (FIR-698)

mogery:
Prerequisite of sending out e-mails. I don't think this will cause too much memory pressure on the Redis side -- it's a set, so we're not storing duplicates, and it's all just UUIDs.


nickscamara:
Why redis and not postgres though? haha

mogery:
didn't want to make a table + redis is way easier to query

nickscamara:
You can feed it to the insertion queue so it doesnt add presure.

nickscamara:
But this works too, unless redis crashes

mogery:
I'm very confident it won't cause redis issues, will keep an eye on the mem pressure though

nickscamara:
kk, not worried about that.

PR #1128, opened 2025-02-04
fix: FFI type mismatches between i8 and c_char

lakshh07:
This PR fixes build errors related to type mismatches in the FFI interface. 
Specifically, it changes function signatures to consistently use libc::c_char 
instead of i8 for C string pointers.

Changes:
- Changed return types from *mut i8 to *mut c_char in extract_links, extract_metadata, and transform_html
- Changed parameter type in free_string from *mut i8 to *mut c_char
- Ensures type consistency across the FFI boundary

The changes fix build errors where the compiler expected *mut i8 but found *mut u8.


mogery:
Thanks for the PR! An alternate solution to this has made it into the main branch.

Issue #1126, opened 2025-02-03
[Self-Host] Docker Build Fails with "Cannot find matching keyid" Error During PNPM Install


piotrstarzynski:
When attempting to build a Docker image, the build process fails during the pnpm install --frozen-lockfile (or pnpm install --prod --frozen-lockfile) step with an error indicating a missing key ID. The error message is:

```
Error: Cannot find matching keyid: {"signatures":[{"sig":"MEQCIHGqHbvc2zImUPEPFpT4grh6rMYslel+lAjFArx8+RUdAiBfnJA+bgmUvO5Lctfkq+46KKDQdx/8RhLPge3pA+EdHA==","keyid":"SHA256:DhQ8wR5APBvFHLF/+Tc+AYvPOdTpcIDqOhxsBHRwC7U"}],"keys":[{"expires":null,"keyid":"SHA256:jl3bwswu80PjjokCgh0o2w5c2U4LhQAE57gj9cz1kzA","keytype":"ecdsa-sha2-nistp256","scheme":"ecdsa-sha2-nistp256","key":"MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAE1Olb3zMAFFxXKHiIkQO5cJ3Yhl5i6UPp+IhuteBJbuHcA5UogKo0EWtlWwW6KSaKoTNEYL7JlCQiVnkhBktUgg=="}]}
    at verifySignature (/usr/local/lib/node_modules/corepack/dist/lib/corepack.cjs:21535:47)
    ...
Node.js v20.18.2
```

The error prevents the Docker build from completing successfully.

**To Reproduce**
Clone the repository or obtain the Docker configuration files.
Run docker compose build --no-cache (or a similar build command).
Observe that the build halts with the "Cannot find matching keyid" error while installing packages with PNPM.

**Expected Behavior**
The Docker image should build without throwing a key-matching error, completing the pnpm install step successfully.

ftonato:
Hi @piotrstarzynski,

Thank you for reaching out! I’ll be assisting you with your questions or issues. I will take a look at it and provide you with an update once I have one 😉

ftonato:
Hello @piotrstarzynski,
After

piotrstarzynski:
Hello @ftonato,

Yeah, of course. I am not sure if I did it right.

When trying:
```
pnpm install
Command 'pnpm' not found, did you mean:
  command 'npm' from deb npm (8.5.1~ds-1)
[...]

PR #1131, opened 2025-02-04
Fix corepack and self hosting setup

rothnic:
This is a set of changes that addressed the corepack issue and some additional issues with running the docker compose file in my self-hosted environment.

See issue: #1126 

I fixed the corepack issue by following this approach:
* [pnpm corepack fix suggestion](https://github.com/pnpm/pnpm/issues/9029#issuecomment-2631400936)

After fixing the corepack issue, I was left with this error:
* `ulimit: open files: cannot modify limit: Operation not permitted`
* In my research, you cannot set the ulimit from the docker entrypoint script.
* You need to set it as part of running the image like `docker run --ulimit nofile=65535:65535 your-image` or in the docker compose file as i have implemented

### DIFF
--- a/apps/api/docker-entrypoint.sh
+++ b/apps/api/docker-entrypoint.sh
@@ -1,5 +1,14 @@
 #!/bin/bash -e
 
+if [ $UID -eq 0 ]; then
+  set +e # disable failing on errror
+  ulimit -n 65535
+  echo "NEW ULIMIT: $(ulimit -n)"
+  set -e # enable failing on error
+else
+  echo ENTRYPOINT DID NOT RUN AS ROOT
+fi
+
 if [ $FLY_PROCESS_GROUP = "app" ]; then
   echo "RUNNING app"
   node --max-old-space-size=8192 dist/src/index.js
--- a/apps/api/docker-entrypoint.sh
+++ b/apps/api/docker-entrypoint.sh
@@ -1,12 +1,5 @@
 #!/bin/bash -e
 
-if [ $UID -eq 0 ]; then
-  ulimit -n 65535
-  echo "NEW ULIMIT: $(ulimit -n)"
-else
-  echo ENTRYPOINT DID NOT RUN AS ROOT
-fi
-
 if [ $FLY_PROCESS_GROUP = "app" ]; then
   echo "RUNNING app"
   node --max-old-space-size=8192 dist/src/index.js
--- a/apps/api/Dockerfile
+++ b/apps/api/Dockerfile
@@ -3,9 +3,8 @@ FROM node:20-slim AS base
 ENV PNPM_HOME="/pnpm"
 ENV PATH="$PNPM_HOME:$PATH"
 LABEL fly_launch_runtime="Node.js"
-RUN npm install -g corepack
-RUN corepack enable
-RUN corepack prepare pnpm@10.2.0 --activate
+RUN corepack enable && npm install -g corepack@latest
+
 COPY . /app
 WORKDIR /app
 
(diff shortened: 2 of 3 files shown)


apps/api/docker-entrypoint.sh
```diff
@@ -1,12 +1,5 @@
 #!/bin/bash -e
 
-if [ $UID -eq 0 ]; then
-  ulimit -n 65535
-  echo "NEW ULIMIT: $(ulimit -n)"
-else
-  echo ENTRYPOINT DID NOT RUN AS ROOT
-fi
-
```
mogery: This may not be removed. It has very important implications on production. Is there any reason for removing this?

mogery: Oh, I see you set it in docker-compose.yaml. However, we have no option to do that in production. Let me investigate...

rothnic: @mogery  Correct... in my environment, which is portainer on an unraid server, the ulimit command causes the whole container to error out and not run. Are you not running this with docker? If so, you can see the other approach of just using a docker run command. `docker run --ulimit nofile=65535:65535 your-image`

Should there be a different script for running in your production environment vs using with docker if you aren't using with docker? At the moment, it seems it won't work in all environments, so it should be able to be controlled in some way or another.

mogery: Interesting! We're using Kubernetes on our end. I'll see what we can do to configure ulimit on the YAML side.

mogery: What error are you getting by the way? If the ulimit command doesn't exist I can just make it run only if the command exists.

rothnic: `ulimit: open files: cannot modify limit: Operation not permitted`

rothnic: Maybe it could be controlled through an env variable? I'm not fully certain on the need of using it, just that in my research it was suggested to not do it in the entrypoint script since the runtime env could influence whether it works or not

rothnic: @mogery if it helps, [here is the chatgpt session](https://chatgpt.com/share/67a256dd-37bc-800c-9034-9522526c595e) where i was trying to figure out how to fix the issue. It seems like it is kind of a conflict between how the docker vs kubernetes environments manage resource limits. When running within a docker environment, the limits are already set, so you can't adjust them. 

IMHO, the easiest fix would be to revert my change, then modify the entrypoint script to look for a new env variable like `RUNTIME_MODE=docker`. If that env variable isn't set, then run the command. If the env variable is set to docker, then don't run the command, or something like that. It would not impact your existing systems and we could add the new env variable to the docker-compose.yaml file to default to "docker", so it wouldn't be executed.

mogery: Here's what should be done instead:
```diff
  if [ $UID -eq 0 ]; then
+   set +e # disable failing on errror
    ulimit -n 65535
    echo "NEW ULIMIT: $(ulimit -n)"
+   set -e # enable failing on error
  else
    echo ENTRYPOINT DID NOT RUN AS ROOT
  fi
```

Can you make this change on your branch in the shell file and see if it works for you?

rothnic: I implemented this change and tested rebuilding from scratch via portainer by just pointing to my github branch and the docker-compose file. It looks like setting the ulimit setting in the docker compose file allows the ulimit command to succeed, but probably should leave it as-is in case some other environment causes the command to fail for some reason

rothnic:
@mogery I updated the docker-entrypoint.sh file as requested and it worked fine on my server. However, I then tried to clone the branch fresh on my local macbook pro (intel) to test that and ran into a signature issue with one of the commands in the Dockerfile. So, I made a tweak to it to avoid that issue and it works now on my macbook pro as well without having to modify the docker compose or dockerfile.

This I think is now ready to go

mogery (approved):
LGTM

PR #1146, opened 2025-02-07
[FIR-796] feat(api/types): Add action and wait time validation for scrape requests

ftonato:
- Implement max actions limit (15 actions)
- Add total wait time calculation for actions
- Increase max waitFor time to 60 seconds
- Refactor scrape options schema to include validation

### DIFF
--- a/apps/api/src/controllers/v1/types.ts
+++ b/apps/api/src/controllers/v1/types.ts
@@ -67,6 +67,25 @@ export const extractOptions = z
 
 export type ExtractOptions = z.infer<typeof extractOptions>;
 
+const ACTIONS_MAX_WAIT_TIME = 60;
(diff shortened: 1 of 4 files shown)


apps/api/src/controllers/v1/types.ts
```diff
@@ -113,9 +132,19 @@ export const actionsSchema = z.array(
       script: z.string(),
     }),
   ]),
+).refine(
+  (actions) => actions.length <= MAX_ACTIONS,
+  {
+    message: `Maximum of ${MAX_ACTIONS} actions allowed`,
```
mogery:
```suggestion
    message: `Number of actions cannot exceed ${MAX_ACTIONS}`,
```

mogery (approved):
Great!

PR #1147, opened 2025-02-08
Add Serper and Search API env vars to docker-compose

RealLukeMartin:
Set up self hosting and I popped in the SERPER_API_KEY to the .env but it seemed to not work. Noticed that the docker-compose.yaml didn't have a reference for that so added it in and it worked of course.

This just takes away a small amount of friction when someone wants to use either of those two search services.


nickscamara:
Thank you @RealLukeMartin!

PR #1175, opened 2025-02-12
fix: update gemini library. extract pdf links from scraped content

mayooear:
- Updates logic to latest gemini api library
- Fixes extract of pdf and image urls from the scraped content


ericciarla:
lgtm!

PR #1183, opened 2025-02-14
fix(scrapeURL/sb): enforce timeout (FIR-980)

mogery:
We had a few jobs get stuck running *forever* on SB. Added code to enforce the timeout


nickscamara (changes requested):
LGTM, but please remove the flake files

mogery:
> LGTM, but please remove the flake files

why? i cannot develop from my laptop if i remove it

PR #1184, opened 2025-02-16
fix(crawl-status): consider concurrency limited jobs as prioritized (FIR-851)

mogery:
Fixes a multitude of annoying bugs, including:
 - Crawls that have all their pending jobs in the concurrency queue are considered "completed" (FIR-851)
 - Concurrency limited scrapes do not show up in the `total` immediately


mogery:
Did not write a test because it's really dumb hard to test for this. We should build a harness I think, eventually.

nickscamara (approved):

PR #1191, opened 2025-02-17
fix(map): do not remove query parameters from results (FIR-1015)

mogery:
Fixes `sitemapOnly`'s faithfulness to the real sitemap, and also should improve Extract for sites that use URLs like `https://example.com/product?id=<xyz>` for product pages/anything that needs to be multi-entity extracted.


mogery:
NOTE: test failing is unrelated, reported issue to Thomas

PR #1193, opened 2025-02-17
Add searxng for search endpoint

### LINKED ISSUE #1145: [Feat] SearXNG Support
Problem:
Right now, the search endpoint only uses default engines. This can be a drawback for those who want a more privacy-focused or flexible search option.

Proposed Feature:
Add an option to use SearXNG by letting users set a parameter like backend: "searxng". This should include the ability to provide a custom SearXNG instance URL and a fallback to the default if the SearXNG service is unavailable.

Alternatives Considered:
Sticking with the default engine limits privacy and customizability. While other metasearch options exist, SearXNG is well-supported and IMO aligns nicely with this project’s goals.

Use Case:
Developers who need privacy-friendly, customizable search results will benefit from this feature, making Firecrawl more appealing for self-hosted and privacy-conscious setups.

Thanks for considering this addition!

loorisr:
Fixes #1145 

SearXNG has to be configured with json format

It adds 3 new var env:
* SEARXNG_ENGINES see https://docs.searxng.org/user/configured_engines.html
* SEARXNG_CATEGORIES see https://docs.searxng.org/user/configured_engines.html
* SEARXNG_ENDPOINT=https://your-searxng-instance

### DIFF
--- a/apps/api/src/search/searxng.ts
+++ b/apps/api/src/search/searxng.ts
@@ -1,0 +1,66 @@
+import axios from "axios";
+import dotenv from "dotenv";
+import { SearchResult } from "../../src/lib/entities";
+
+dotenv.config();
+
+interface SearchOptions {
+  tbs?: string;
+  filter?: string;
+  lang?: string;
+  country?: string;
+  location?: string;
+  num_results: number;
+  page?: number;
+}
+
+export async function searxng_search(
+  q: string,
+  options: SearchOptions,
+): Promise<SearchResult[]> {
+  const params = {
+    q: q,
+    language: options.lang,
+    // gl: options.country, //not possible with SearXNG
+    // location: options.location, //not possible with SearXNG
+    // num: options.num_results, //not possible with SearXNG
+    engines: process.env.SEARXNG_ENGINES || "",
+    categories: process.env.SEARXNG_CATEGORIES || "general",
+    pageno: options.page ?? 1,
+    format: "json"
+  };
+
+  const url = process.env.SEARXNG_ENDPOINT as string;
+  if (!url) {
+    console.error(`SEARXNG_ENDPOINT environment variable is not set`);
+  }
+  // Remove trailing slash if it exists
+  const cleanedUrl = url.endsWith('/') ? url.slice(0, -1) : url;
+
+  // Concatenate "/search" to the cleaned URL
+  const finalUrl = cleanedUrl + "/search";
+
+  try {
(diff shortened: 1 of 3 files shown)


loorisr:
I've deleted by mistake this branch on my repo and it automatically closed it :(

apps/api/src/search/searxng.ts
```diff
@@ -0,0 +1,66 @@
...
+    engines: process.env.SEARXNG_ENGINES || "",
+    categories: process.env.SEARXNG_CATEGORIES || "general",
+    pageno: options.page ?? 1,
+    format: "json"
+  };
+
+  const url = process.env.SEARXNG_ENDPOINT as string;
+  if (!url) {
+    console.error(`SEARXNG_ENDPOINT environment variable is not set`);
+  }
```
mogery: Not required since it's only called if the endpoint is defined
```suggestion
  const url = process.env.SEARXNG_ENDPOINT!;
```

apps/api/src/search/searxng.ts
```diff
@@ -0,0 +1,66 @@
...
+      return data.results.map((a: any) => ({
+        url: a.url,
+        title: a.title,
+        description: a.content,
+      }));
+    } else {
+      return [];
+    }
+  } catch (error) {
+    console.error(`There was an error searching for content: ${error.message}`);
```
mogery:
```suggestion
    logger.error(`There was an error searching for content`, { error });
```

apps/api/src/search/searxng.ts
```diff
@@ -0,0 +1,66 @@
+import axios from "axios";
+import dotenv from "dotenv";
+import { SearchResult } from "../../src/lib/entities";
+
```
mogery, line 5:
```suggestion
import { logger } from "../lib/logger"

```

mogery (approved):

mogery:
@loorisr do you know of any public SearXNG instances with JSON enabled that we can point to for testing?

loorisr:
you can use https://search.url4irl.com

mogery:
> you can use https://search.url4irl.com

It's getting a 429 Too Many Requests

loorisr:
it is very rate limited. you may have to setup your own instance.

mogery:
Ended up bootstrapping a whole SearXNG instance in the worker. Dodgy, but works! Merging if this passes

mogery:
Thank you!

PR #1200, opened 2025-02-18
Fixed scrape status not working on self-host

mikelmao:
The scrape status would fail because it was only considering it with supabase authentication. Added a check for bullMQ


mogery:
Hi Mike! Thanks for the PR. An alternative solution made it into #1207 and we're going to merge it soon.

PR #1204, opened 2025-02-19
feat(sdk): enforce timeout on client-side if set

mogery:
needs a thorough review on the Python side as I have no idea what I'm doing on that end. Also make sure to run the SDK tests, we don't have them in the CI


rafaelsideguide (approved):
lgtm

PR #1207, opened 2025-02-19
SELFHOST FIXES

mogery:
Big pile of fixes for running Firecrawl on self-host, including continuous testing in a simulated self-hosted environment.


mogery:
cc @rothnic self-host fix development will continue here!

Issue #1230, opened 2025-02-21
[Self-Host] Proposal: cleanup api Dockerfile


loorisr:
Hello,

Thanks for the better Dockerfile in the last release :)

I have further improvements. It almost divides by 2 the size of the image. I can make a PR if you are ok.

* use Node 22 FROM node:20-slim AS base
* with Node 22-slim I don't have problem with corepack and RUN npm i -g corepack@latest is not necessary
* migrate GO to version 1.24: FROM golang:1.19 AS go-base
* migrate slim image for Rust : FROM rust:1-bullseye AS rust-base
* not necessary: RUN apt-get clean && apt-get update -qq && apt-get install -y ca-certificates && update-ca-certificates
* remove : ENV PUPPETEER_EXECUTABLE_PATH="/usr/bin/chromium"
* the prod-deps image is not useful because you overwrite the /app directory from the build image

mogery:
Hey, this sounds awesome, we would love a PR with this!

PR #1231, opened 2025-02-21
Update API Dockerfile

loorisr:
Update to latest version of tools.
Clean up
Reduce image size to 800mo (uncompressed)

See #1230

### DIFF
--- a/apps/api/Dockerfile
+++ b/apps/api/Dockerfile
@@ -1,28 +1,18 @@
-FROM node:20-slim AS base
+FROM node:22-slim AS base
 
 ENV PNPM_HOME="/pnpm"
 ENV PATH="$PNPM_HOME:$PATH"
 LABEL fly_launch_runtime="Node.js"
 
-#temp solution to https://github.com/nodejs/corepack/issues/612
-RUN npm i -g corepack@latest 
-
 RUN corepack enable
 COPY . /app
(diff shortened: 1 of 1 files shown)


apps/api/Dockerfile
```diff
@@ -41,14 +31,12 @@ RUN cd /app/sharedLibs/html-transformer && \
     chmod +x target/release/libhtml_transformer.so
 
 FROM base
-COPY --from=prod-deps /app/node_modules /app/node_modules
 COPY --from=build /app /app
```
mogery, line 34: Not convinced this is the right move here, as we specifically only want the prod deps to be in the resulting image. A better solution seems to me like re-adding the prod-deps build and moving the copy order around.
```suggestion
COPY --from=build /app /app
COPY --from=prod-deps /app/node_modules /app/node_modules
```

loorisr: Yes I agree, I think you should copy /app/dist from the build docker . not /app
Cf https://pnpm.io/docker

loorisr: I've just tried and it doesn't run. It does compile but then 
```
Cannot find module 'mammoth'
Cannot find module 'express'
```

mogery: Yeah, saw the same. Trying to figure it out.

PR #1269, opened 2025-02-27
feat(self-host/ai): pass in the ollama envs into docker compose

brrock:
# Changes
- Adds embedding model name and ollama base url in the x-common-env
# Why this is necessary 
Ollama in self hosted version doesn't work and extract can't work
In selfhost.md it says you can set  it and it doesn't work.


mogery (approved):
Thank you!

brrock:
Thanks what a great project deserves

PR #1282, opened 2025-03-02
docs: remove undefined "required" field

jmporchet:
The documentation example has a "supports_SSO" field that is specified neither in the prompt nor in the schema.
Removing it will make the docs more understandable.


mogery (approved):

mogery:
Thank you!

PR #1303, opened 2025-03-05
feat(crawl): includes/excludes fixes (FIR-1300)

mogery:
- Update includes/excludes behaviour to apply to the crawl source URL too (after discovery)
- Add `regexOnFullURL` parameter


mogery:
don't merge yet -- just running tests :)

mogery:
@tomkosm adblock tests broken on f-e end

PR #1310, opened 2025-03-08
fix: Normalize URL to lowercase in LLM text generation

ftonato:


mogery (changes requested):
Needs to happen after the body is parsed, not before.

PR #1319, opened 2025-03-10
Fix SearxNG categories

loorisr:
If categories is set, then the engines parameters don't have an effect.

You must use either categories OR engines. I propose to remove the default value for categories so the user can set en engine using the env var


mogery:
Thanks!

PR #1330, opened 2025-03-12
fix: resolve circular JSON structure error in search function

invarrow:
Modified error handlers to avoid logging entire response objects that contain circular references


invarrow:
solves https://github.com/mendableai/firecrawl/issues/1326

PR #1346, opened 2025-03-16
(feat/extract) URLs can now be optional in /extract

nickscamara:


apps/js-sdk/firecrawl/src/index.ts
```diff
@@ -1119,14 +1119,14 @@ export default class FirecrawlApp {
...
+   * @param urls - The URLs to extract information from. Optional if using other methods for data extraction.
    * @param params - Additional parameters for the extract request.
    * @returns The response from the extract operation.
    */
-  async extract<T extends zt.ZodSchema = any>(urls: string[], params?: ExtractParams<T>): Promise<ExtractResponse<zt.infer<T>> | ErrorResponse> {
+  async extract<T extends zt.ZodSchema = any>(urls?: string[], params?: ExtractParams<T>): Promise<ExtractResponse<zt.infer<T>> | ErrorResponse> {
     const headers = this.prepareHeaders();
 
-    let jsonData: { urls: string[] } & ExtractParams<T> = { urls,  ...params };
+    let jsonData: { urls?: string[] } & ExtractParams<T> = { urls: urls || [],  ...params };
```
mogery: i'd prefer if we just sent in undefined here but that's just me nitpicking
```suggestion
    let jsonData: { urls?: string[] } & ExtractParams<T> = { urls,  ...params };
```

mogery (approved):
LGTM! This is great!

apps/js-sdk/firecrawl/src/index.ts
nickscamara: agreed

nickscamara: done

PR #1348, opened 2025-03-16
feat(scrape): add warning to document if it was concurrency limited

mogery:


apps/api/src/services/queue-worker.ts
nickscamara, line 1: maybe change it to something around: "...was throttled at your current concurrent limit. If you'd like to scrape faster, you can upgrade your plan..."

mogery: donezo

PR #1350, opened 2025-03-16
fix(v1/checkCredits): snap crawl limit to remaining credits if over without erroring out (FIR-1450)

mogery:
this causes a lot of customer confusion, better to just minmax it quietly


nickscamara:
Wouldn't they assume it is broken now in cases where we return way lower than the limit? Could we possibly add a warning that goes with it?

mogery:
crawl has no warning field — we'll see in support if it happens

PR #1351, opened 2025-03-17
(feat/deep-research) Add Analysis Prompt

nickscamara:
The user should be able to format the final analysis however they want.


mogery (approved):
LGTM

PR #1359, opened 2025-03-19
Fix Typo

WALL-E:


ftonato:
Hello,

This is not a typo—it was an intentional way of expressing ourselves, as our API key starts with fc, which comes from Firecrawl 🔥.

Thank you for proposing it!

Issue #1327, opened 2025-03-11
[Self-Host] 408 Client Error: Request Timeout / Missing lock for job


JohnDotOwl:
I have 3 workers , and generally around 15 minutes or so of processing data 1 by 1 no concurrency , it would reach a stage where 2 active gets frozen and stuck in active state. Any future job sent to the API causes Firecrawl: 408 Client Error: Request Timeout for url: http://localhost:3002/v1/scrape

i've tested on 2 different machines using docker, both are unstable. Does the production firecrawl use this repository for their cloud services? 

Log , seems like it crashed 

```
2025-03-12 00:55:18 worker-3              | 2025-03-11 16:55:18 debug [ScrapeURL:]: Engine playwright does not meet feature priority threshold 
2025-03-12 00:55:18 worker-3              | 2025-03-11 16:55:18 debug [ScrapeURL:]: Engine fetch meets feature priority threshold 
2025-03-12 00:55:18 worker-3              | 2025-03-11 16:55:18 debug [ScrapeURL:]: Engine pdf meets feature priority threshold 
2025-03-12 00:55:18 worker-3              | 2025-03-11 16:55:18 debug [ScrapeURL:]: Engine docx meets feature priority threshold 
2025-03-12 00:55:18 worker-3              | 2025-03-11 16:55:18 info [ScrapeURL:]: Scraping via fetch... 
2025-03-12 00:55:18 worker-3              | 2025-03-11 16:55:18 info [ScrapeURL:]: Scrape via fetch deemed successful. 
2025-03-12 00:55:18 worker-3              | 2025-03-11 16:55:18 debug [ScrapeURL:]: Executed transformers. 
2025-03-12 00:55:18 worker-3              | 2025-03-11 16:55:18 info [queue-worker:processJob]: 🐂 Job done ded21cc6-5d1b-43a0-9186-f1cac8e4d4d8 
2025-03-12 00:55:18 worker-3              | 2025-03-11 16:55:18 debug [queue-worker:processJobInternal]: Job succeeded -- putting result in Redis 
2025-03-12 00:55:18 worker-2              | 2025-03-11 16:55:18 info [queue-worker:processJob]: 🐂 Worker taking job 600f23da-fe42-45b9-b7de-b5c02d49cbb2 
2025-03-12 00:55:18 worker-2              | 2025-03-11 16:55:18 info [ScrapeURL:]: Scraping URL "https://credence-com.sg/contact-us/"... 
[...]

PR #1370, opened 2025-03-21
fix "Missing lock for job" errors and better concurrency handling for self host mode

ciekawy:
I did that for my personal needs related to #1327, hope others can benefit from this


FelixGYT:
First of all, thank you for your generosity. He has helped me a lot.
I deployed your code and it looks much better. There won't be any errors during 20 concurrent attempts.
But after about 20 concurrent requests on my end for a period of time, it still reports the error Job wait.
I checked the logs and found that the work log should have been successful, but there is a job wait issue in the API log.
I don't know if you have encountered the same problem, I'm curious why there is such a problem
微信图片_20250327165105
微信图片_20250327165110

ciekawy:
unfortunately I have very little time and have no issues, while also it is working for my local instance and I use it daily (though not super intensively)  - this should be debugged. 

IMO `extending lock` itself is not an issue - same ID ls logged few lines below as done. May be worth checking logs from other containers (e.g. `docker compose logs --follow` - without container name shows from all and observe - they will be more in sync ) - I'd guess some jobs might possibly get stuck in playwright or something, but definitly timeout should be better handled. So the error might actually be real.

One thing is that all containers, playwright specifically may need more RAM (may require changing either global docker, or whats assigned to specific container).


BTW PLEASE always copy text, not screen :)

PR #1499, opened 2025-04-24
Fix: Add missing crypto import in crawl controller

### LINKED ISSUE #1431: Missing reference
file: /apps/api/src/controllers/vi/crawl.ts
Missing reference: import crypto from "crypto";

shosseini811:
This PR fixes issue #1431 by adding the missing import statement for the crypto module.

The file was using crypto.randomUUID() but didn't import the crypto module, which would cause errors when running the code. I've added the import statement at the top of the file with the other imports.

Fixes #1431


nickscamara:
Hey @shosseini811, thanks for the PR. We are using Node v22 which works out of the box. If you upgrade your node to that it should work just fine. Will close this for now :)

PR #1505, opened 2025-04-28
[python-sdk] Webhook param for batch scrape 

arvida:
The API endpoint `batch/scrape` supports the webhook param, this PR aligns the python client to support this.

Noticed webhook support missing when we upgraded the `firecrawl-py` package in our project recently, webhooks stopped working and we got the error message `Unsupported parameter(s) for async_batch_scrape_urls: webhook`. This PR makes it work again. 

Noticed that the tests look a bit outdated for the python stuff, so didn't touch those. Let me know if you want a test case for this.


nickscamara:
Thanks @arvida!

PR #1590, opened 2025-05-22
Add new example to examples directory

developersdigest:


developersdigest:
Moving to https://github.com/mendableai/firecrawl-app-examples

PR #1605, opened 2025-05-29
Index (FIR-2177)

mogery:


apps/api/src/controllers/v1/map.ts
```diff
@@ -58,6 +59,7 @@ export async function getMapResults({
   mock,
   filterByPath = true,
   flags,
+  ignoreIndex = false,
```
nickscamara: Let's change this to `useIndex` instead. Having false var names gets confusing. Also default useIndex to `false`

apps/api/src/controllers/v1/types.ts
```diff
@@ -309,6 +309,8 @@ const baseScrapeOptions = z
     useMock: z.string().optional(),
     blockAds: z.boolean().default(true),
     proxy: z.enum(["basic", "stealth", "auto"]).optional(),
+    maxAge: z.number().int().gte(0).safe().default(0),
+    dontStoreInCache: z.boolean().default(false),
```
nickscamara: Same thing here please

apps/api/src/services/queue-worker.ts
```diff
@@ -1027,6 +1029,87 @@ async function processKickoffJob(job: Job & { id: string }, token: string) {
       });
     }
 
+    const trimmedURL = new URL(job.data.url);
+    trimmedURL.search = "";
+
+    const urlSplits = generateURLSplits(trimmedURL.href).map(x => hashURL(x));
+
+      const index = (sc.crawlerOptions.ignoreSitemap || process.env.FIRECRAWL_INDEX_WRITE_ONLY === "true" || !useIndex)
```
nickscamara: Can we do a helper function here. Kinda hard to see what's going on.

apps/test-suite/index-benchmark/run.ipynb
nickscamara, line 1: Please remove this from the main repo

apps/api/src/controllers/v1/map.ts
mogery: Why default useIndex to false? Don't we want to improve map by default?

PR #1654, opened 2025-06-11
fix: PLAYWRIGHT_MICROSERVICE_URL in apps/api/.env.example

pulbyte:
The correct environment variable should be PLAYWRIGHT_MICROSERVICE_URL=http://playwright-service:3000/scrape instead of PLAYWRIGHT_MICROSERVICE_URL=http://playwright-service:3000/html

Using the /html endpoint causes a 404 error when calling http://localhost:3002/v1/scrape in self-hosted deployments. Looking at the source code for the playwright-service the actual endpoint is /scrape, not/html.


mogery:
Thank you! Forgot about this file.

PR #1655, opened 2025-06-11
feat: add credits_billed everywhere (FIR-2286)

mogery:
also a bit of logging improvement for logJob


apps/api/src/services/queue-worker.ts
```diff
@@ -359,6 +357,19 @@ async function finishCrawlIfNeeded(job: Job & { id: string }, sc: StoredCrawl) {
       const num_docs = await getDoneJobsOrderedLength(job.data.crawl_id);
       const jobStatus = sc.cancelled ? "failed" : "completed";
 
+      const creditsRpc = await supabase_service
```
nickscamara: Doesn't this break for self host?

mogery: shit yeah i'll fix

PR #1667, opened 2025-06-13
Remove old cache mechanisms (redis cache, PDF cache, crawl maps, etc.) (FIR-2266)

mogery:
**DO NOT MERGE UNTIL JUNE 15TH 7 AM PT** -- need to populate index with subdomain splits for `/map` to work effectively

### DIFF
--- a/apps/api/src/controllers/v1/map.ts
+++ b/apps/api/src/controllers/v1/map.ts
@@ -195,9 +195,6 @@ export async function getMapResults({
       links.push(...indexResults);
     }
 
-    const twoDaysAgo = new Date();
-    twoDaysAgo.setDate(twoDaysAgo.getDate() - 2);
-
     // If sitemap is not ignored, fetch sitemap
     // This will attempt to find it in the index at first, or fetch a fresh one if it's older than 2 days
     if (


apps/api/src/controllers/v1/types.ts
nickscamara, line 1: This breaks stuff, revert

nickscamara: We use this param in an integration I believe, it will do a bad request if we simple remove it. Mark as `@deprecated` please

mogery: Done

apps/api/src/controllers/v1/map.ts
```diff
@@ -202,10 +199,10 @@ export async function getMapResults({
     twoDaysAgo.setDate(twoDaysAgo.getDate() - 2);
 
     // If sitemap is not ignored and either we have few URLs (<100) or the data is stale (>2 days old), fetch fresh sitemap
+    // TODO: We should always soft-fetch the sitemaps from just the index, which fails very fast if there's no result. - mogery
     if (
       !ignoreSitemap &&
-      (sitemapIndexResult.urls.length < 100 ||
-        new Date(sitemapIndexResult.lastUpdated) < twoDaysAgo)
+      (indexResults.length < 100)
```
nickscamara: Keep the same functionality as it was before

mogery: done

nickscamara (approved):

PR #1682, opened 2025-06-18
fix(scrape): log FIRE-1 credits billed on failures properly (FIR-2331)

mogery:
Also simplifies overcomplicated billing check logic (no more async fetching)


mogery:
test failures unrelated, this is good to go

PR #1693, opened 2025-06-20
feat(scrape): support Google Slides

meetsoni1214:
- Add a similar implementation like Google Docs for Google Slides, so we will also be able to scrape Google Slides links


mogery:
Amazing! Was just thinking about implementing this on request :) I'll take a look!

mogery (approved):
Great!

PR #1715, opened 2025-06-25
feat(firecrawl): add integration parameter support and enhance kwargs handling

ftonato:


nickscamara (commented):
@rafaelmmiller can you just double check the kwards? otherwise looks good

PR #1717, opened 2025-06-26
Allow international URLs to pass validation

micahstairs:
The existing URL validation didn't support valid URLs like http://xn--1lqv92a901a.xn--ses554g/.


mogery (approved):

mogery:
good catch!

PR #1731, opened 2025-06-28
fix(logger): correct method names in logger.child calls

ahnafatef:
## Changes
- Fixed incorrect function names passed to `logger.child()`.


ahnafatef:
Couldn't go through the entire codebase yet, so there might be similar typos :)

PR #1803, opened 2025-07-15
feat(types): add JSON schema validation to schema options in extractO…

rafaelsideguide:
…ptions and baseScrapeOptions

@micahstairs we're now using the same avj to validate every json schemas

@mogery can you pls take a look?
    

---

## Summary by cubic
Added JSON schema validation to the schema fields in extractOptions and baseScrapeOptions to ensure only valid schemas are accepted.

- **Validation**
 - Now checks that provided schemas compile with ajv, returning an error if invalid.

### DIFF
--- a/apps/api/src/controllers/v1/types.ts
+++ b/apps/api/src/controllers/v1/types.ts
@@ -95,7 +95,23 @@ export type AgentOptions = z.infer<typeof agentOptionsExtract>;
 export const extractOptions = z
   .object({
     mode: z.enum(["llm"]).default("llm"),
-    schema: z.any().optional(),
+    schema: z
+      .any()
+      .optional()
+      .refine(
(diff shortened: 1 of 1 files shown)


mogery:
LGTM

mogery (approved):

micahstairs:
@rafaelsideguide if this is a breaking change, should it wait until v2?

mogery:
IMO this is not a breaking change, all schemas that this rejects would return an error later in the process

rafaelsideguide:
ok merging

micahstairs:
> IMO this is not a breaking change, all schemas that this rejects would return an error later in the process

FYI @rafaelsideguide @mogery this was actually backwards incompatible. We need to be more careful about things like this in the future!

PR #1849, opened 2025-07-25
Enhancement: Add SSL Verification Control to `search()` Method in Python SDK

Sushil2308:
## Summary

This pull request enhances the `search()` method of the Firecrawl Python SDK by adding support for SSL certificate verification control via a `verify` keyword argument. This gives users more flexibility when interacting with endpoints that use self-signed or custom CA certificates.

## What's Changed

- Added support for `verify` in the `**kwargs` of the `search()` method.
- The `verify` option is passed directly to `requests.post(...)`:
    - `verify=False`: disables SSL cert validation (use only in trusted/dev environments).
    - `verify=path/to/cacert.pem`: uses a custom CA bundle.
    - `verify=True` (default behavior if omitted): validates using system cert store.
- Updated the function docstring to document the `verify` parameter.
- Added internal handling for defaulting to `verify=False` to maintain current behavior unless overridden by the user.

## Why This Is Needed

While working with the search() method in a production server environment, I encountered SSL certificate validation failures. This was due to the presence of self-signed or privately issued certificates within the infrastructure. To address this, the SDK has been enhanced to allow configurable SSL verification, enabling secure and flexible integration in production and enterprise settings.

## Usage Examples

```python
# 1. Disable SSL certificate verification (not recommended for production)
firecrawl.search("openai", verify=False)

# 2. Use a custom CA bundle
firecrawl.search("openai", verify="/path/to/ca_bundle.pem")

    
<!-- This is an auto-generated description by cubic. -->
---

## Summary by cubic
Added a verify option to the search() method in the Python SDK, allowing users to control SSL certificate verification for API requests.

- **New Features**
  - Accepts verify as a keyword argument to enable, disable, or customize SSL verification when making requests.

<!-- End of auto-generated description by cubic. -->

### DIFF
--- a/apps/python-sdk/firecrawl/firecrawl.py
+++ b/apps/python-sdk/firecrawl/firecrawl.py
@@ -636,6 +636,9 @@ class FirecrawlApp:
             location (Optional[str]): Geo-targeting
             timeout (Optional[int]): Request timeout in milliseconds
             scrape_options (Optional[ScrapeOptions]): Result scraping configuration
+            verify (Union[bool, str]): Whether to verify SSL certificates. Can be:
+                - False: Disable SSL verification (not recommended for production)
+                - str: Path to a CA_BUNDLE file or directory with certificates of trusted CAs.
             **kwargs: Additional keyword arguments for future compatibility
 
         Returns:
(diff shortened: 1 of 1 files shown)


mogery:
There is no self-signed cert on prod, you're missing the `certifi` package.

Sushil2308:
> There is no self-signed cert on prod, you're missing the `certifi` package.

This is installed but there is issue in raise post request for search and while post request need verify either be false or require ssl path.

I have only fix the passing verify flag in post api.

PR #1879, opened 2025-07-29
queue-worker: move scrape worker to separate threads (ENG-3008)

mogery:
This makes use of BullMQ's sandboxed threads feature, allowing us to:
1. Simplify worker logic
2. Have less event loops tangled together, increasing performance and resilience

https://docs.bullmq.io/guide/workers/sandboxed-processors
    

---

## Summary by cubic
Moved the scrape worker logic to a separate thread using BullMQ's sandboxed worker feature to improve performance and simplify the main queue worker code.

- **Refactors**
  - Extracted scrape job processing into its own file and thread.
  - Reduced complexity in the main queue-worker by removing scrape-specific logic.


mogery:
fail unrelated

PR #1891, opened 2025-07-30
Fix User-Agent handling in Playwright service

allen12921:
- Remove User-Agent from browser context initialization
- Handle User-Agent at request level with proper priority
- Prioritize user-provided User-Agent from headers over random generation
- Merge all headers into single setExtraHTTPHeaders call to prevent overwrites
- Add case-insensitive User-Agent header detection
- Improve logging for User-Agent usage


allen12921:
Hi @mogery ,could u help to review the change ?

PR #1904, opened 2025-08-01
fix(docs): correct link to Map

ChetanGoti:
## Summary by cubic
Fixed the Map link in the README so it now points to the correct section.


mogery:
ty!

Issue #1836, opened 2025-07-23
[Bug] JS SDK #extract Returns Empty Data with Zod Schema


JonoAugustine:
**Describe the Bug**
Attempting to extract using a zod schema returns an empty object. Including a prompt does not affect the result.

**To Reproduce**
Easily reproducible:

```typescript
import { z } from "zod";
import FirecrawlApp from "@mendable/firecrawl-js";

const firecrawl = new FirecrawlApp({ apiKey: "<key>"  });

const base = z.object({
  commonName: z.string(),
  scientificName: z.string(),
})

firecrawl.extract(["https://en.m.wikipedia.org/wiki/Zebra_acara"], {
  schema: base
}).then((response) => {
  if (response.success) {
    console.log("as documented", response.data)
  }
})
```
output:
```bash
as documented {}
```

**Expected Behavior**
Should return data fitting the schema extracted from the page. 

I have found a workaround by "manually" converting zod to a jsoncshema:

```typescript
const modifiedSchema = z.toJSONSchema(base)
delete modifiedSchema.$schema # !! NB - if $schema is not deleted, the endpoint rejects the schema with 400

firecrawl.extract(["https://en.m.wikipedia.org/wiki/Zebra_acara"], {
  schema: modifiedSchema
}).then((response) => {
  if (response.success) {
    console.log("with workaround", response.data)
  }
})
```

output:
```bash
with workaround { commonName: 'Zebra acara', scientificName: 'Ivanacara adoketa' }
```

**Environment**
- OS: Arch
- Deployment Type: Cloud
- Firecrawl Version: 1.29.1, 1.29.2
- Node.js Version: 24.3.0
- zod Version: 4.0.5

Jackpkn:
Hey @JonoAugustine  👋

I’ve pushed a fix for the Zod schema bug!

---

### What was wrong
The issue was caused by the `zod-to-json-schema` converter adding a `$schema` property, which the Firecrawl backend didn’t like. This aligns with what the original issue suggested as a workaround (manually removing `$schema`).

---

###  Fix
In both `extract` and `asyncExtract` (inside `apps/js-sdk/firecrawl/src/index.ts`), I added a line to remove `$schema` right after the schema conversion:

```ts
// Inside the schema processing try...catch block in both extract and asyncExtract
try {
  jsonSchema = zodToJsonSchema(params.schema as zt.ZodType);
  // My fix: Remove the $schema property if it was added
[...]

PR #1906, opened 2025-08-01
fix(js-sdk): remove $schema property from Zod-converted schema (#1836)

Jackpkn:
This PR fixes issue #1836.

**The Problem:**
When you tried to use a Zod schema with the `extract` function in the JS SDK, it didn't work and returned empty data `{}`. The reason is that the tool used to convert the Zod schema (`zod-to-json-schema`) automatically adds a `$schema` property. The Firecrawl backend doesn't like this extra property and rejects the request.

**The Solution:**
I changed the code in `src/index.ts` for both the `extract` and `asyncExtract` functions. Right after the Zod schema is converted to a regular JSON schema, I added a small piece of code to check if the `$schema` property is there, and if it is, it deletes it. This matches the workaround that the person who reported the bug found.

Here's the code I added:
```typescript
// Inside the schema processing try...catch block in both extract and asyncExtract
try {
  jsonSchema = zodToJsonSchema(params.schema as zt.ZodType);
  // My fix: Remove the $schema property if it was added
  if (jsonSchema && typeof jsonSchema === 'object' && jsonSchema.$schema) {
    delete jsonSchema.$schema;
  }
} catch (_) {
  jsonSchema = params.schema;
}
    
<!-- This is an auto-generated description by cubic. -->
---

## Summary by cubic
Fixed extract and asyncExtract functions in the JS SDK to remove the $schema property from Zod-converted schemas, preventing backend errors when using Zod schemas.

- **Bug Fixes**
  - Strips the $schema property after converting Zod schemas to JSON schema before sending to the backend, matching the workaround in issue #1836.

<!-- End of auto-generated description by cubic. -->

### DIFF
--- a/apps/js-sdk/firecrawl/src/index.ts
+++ b/apps/js-sdk/firecrawl/src/index.ts
@@ -1500,9 +1500,6 @@ export default class FirecrawlApp {
           ) {
             delete jsonSchema.$schema;
           }
-          console.log(
-            "✅ Applied manual Zod-to-JSON conversion and removed $schema for workaround test."
-          );
         } catch (_) {
           // If conversion fails, assume it's already a JSON schema object
           jsonSchema = params.schema;


Jackpkn:
@rafaelsideguide can you review it ?

Issue #1925, opened 2025-08-06
Add Asynchronous Search Function to Improve Performance and Scalability


CodeDieSleep:
---

### **🛠️ Problem Description**

I find it difficult to scale the current search system efficiently under high concurrency. The existing synchronous search function blocks the event loop, especially during I/O-heavy operations (e.g., database lookups or external API calls). This results in increased response time, degraded user experience, and inefficient resource utilization.

---

### **✅ Proposed Feature**

Introduce an **asynchronous search function** using Python's `async/await` syntax. This function should support non-blocking execution of search queries, allowing the application to continue processing other requests while awaiting results.

---

### **🔄 Alternatives Considered**

* **Thread-based Concurrency:** Considered using threads with synchronous functions. However, thread pools increase complexity, consume more memory, and don’t scale as efficiently as async I/O.
* **Task Queues (e.g., Celery):** While useful for background jobs, they introduce latency and complexity for real-time search operations, which need immediate responsiveness.

These alternatives were not suitable because they either increase infrastructure costs or degrade real-time performance.

---

### **🧩 Implementation Suggestions**

* Use Python's `asyncio` or `trio` to handle asynchronous search operations.
* If using a web framework like **FastAPI**, ensure the search function is declared with `async def` and compatible with async DB clients (e.g., `asyncpg`, `motor`, `elasticsearch-py-async`).
* Abstract the search logic into an async service layer to allow reuse across APIs and background tasks.
* Optionally support **timeout control**, **cancellation tokens**, and **streamed results** for large datasets.

Example:

```python
async def async_search(query: str, filters: dict) -> List[SearchResult]:
    results = await search_engine.query_async(query, filters)
    return results
```

---

### **📌 Use Case**

* A user initiates a document search while other users are uploading files or querying metadata.
* Without async, the search blocks the server and slows down the system for everyone.
* With async, the search runs in parallel, improving responsiveness and throughput even under high load.

Example:

> “In a document management system, multiple users may search and upload files concurrently. Async search ensures the UI stays responsive and scales with user load.”

---

### **📎 Additional Context**

* Similar async search implementations are available in **Elasticsearch async clients**, **Supabase**, and **FastAPI search plugins**.
* Systems like **Weaviate**, **Qdrant**, and **Typesense** also benefit from non-blocking I/O, showing improved performance with async operations.
* \[Link to prototype or code snippet if available]

---

mogery:
This should be fixed in the latest SDKs

[...]

PR #1926, opened 2025-08-06
Add async_search method for async-compatible querying

Sushil2308:
This PR introduces an async_search method to the Firecrawl client, enabling asynchronous usage of the /v1/search endpoint via httpx.AsyncClient. It mirrors the existing search method in functionality while providing full support for non-blocking I/O operations in asynchronous environments.
✨ Key Highlights
Adds async_search method with the same signature and parameters as search, ensuring consistency.
Uses httpx.AsyncClient for making asynchronous HTTP POST requests.
Supports all existing parameters, including:
limit, tbs, filter, lang, country, location, timeout, scrape_options, and **kwargs
Integrates seamlessly with the existing internal logic:
_validate_kwargs
_handle_error
Preserves the return type as SearchResponse, ensuring a backward-compatible data structure.

🚀 Benefits of async_search (Async I/O)
1. Non-blocking I/O for High-Concurrency Apps
Traditional requests are blocking — the thread waits until a response is received.
With asyncio and httpx.AsyncClient, async_search lets your app fire multiple search requests concurrently, without blocking the event loop.
2. Improved Performance in Asynchronous Frameworks
Ideal for FastAPI, aiohttp, or any asyncio-based architecture.
Reduces latency when calling multiple endpoints or integrating with other async tasks like:
scraping
parsing
database access
file I/O
3. Better Resource Utilization
Frees up CPU while waiting for the network.
Unlike multi-threading, it avoids thread overhead and is more memory-efficient for I/O-bound workloads.
4. Scalable API Integrations
You can run 100s or 1000s of concurrent searches with minimal overhead.
Great for data pipelines, crawlers, AI agents, or multi-query dashboards.

This feature is open #1925 

I have added one extra package `httpx` and created `asyn_search` function.

### DIFF
--- a/apps/python-sdk/firecrawl/firecrawl.py
+++ b/apps/python-sdk/firecrawl/firecrawl.py
@@ -23,6 +23,7 @@ import websockets
 import aiohttp
 import asyncio
 from pydantic import Field
+import httpx
 
 # Suppress Pydantic warnings about attribute shadowing
 warnings.filterwarnings("ignore", message="Field name \"json\" in \"FirecrawlDocument\" shadows an attribute in parent \"BaseModel\"")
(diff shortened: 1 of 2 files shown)


Sushil2308:
Please let me know when this can be merged. We need this functionality for a large-scale project; otherwise, we’ll have to write a custom library to integrate Firecrawl. It would be great to avoid duplicating efforts if this can be merged soon.
Thanks!

PR #1992, opened 2025-08-21
feat(api): support extraction of data-* attributes in scrape endpoints (ENG-3197)

### LINKED ISSUE #1981: [Feat] Support Extraction of data-* Attributes in API
**Problem Description**
Many sites store meaningful content in data-* attributes rather than in text. For example, in below screenshot, vehicle info are placed in data-vehicle-name. I tried to retrieve these values by passing the attribute name via the API’s tag field, but it didn’t work. This makes it hard to extract accurate data when the DOM text is empty or normalized differently than the attribute value.

Image

**Proposed Feature**
Add support for selecting and returning values from HTML data-* attributes. Ideally, the API should allow users to specify a CSS selector plus the attribute to extract (e.g., data-vehicle-name) and receive those values in the response.

vishkrish200:
### Summary
Adds a new `extractDataAttributes` option to scrape endpoints that allows extracting specific data attributes from HTML elements using CSS selectors.
Closes #1981

### Key Changes
- **New API option**: `extractDataAttributes` array with `selector` and `attribute` fields
- **Response enhancement**: Adds `dataAttributes` field to scrape responses
- **Flexible attribute handling**: Supports both `data-attribute` and `attribute` formats
- **Multi-value support**: Returns arrays when multiple elements match

### Usage Example
```json
{
  "url": "https://example.com",
  "extractDataAttributes": [
    {
      "selector": ".product-card",
      "attribute": "data-product-id"
    }
  ]
}
```

### Response Format
```json
{
  "dataAttributes": [
    {
      "selector": ".product-card",
      "attribute": "data-product-id",
      "values": ["123", "456"]
    }
  ]
}
```

### Implementation
- New transformer `deriveDataAttributesFromHTML` in scraping pipeline
- Core extraction logic in `extractDataAttributes` function
- Added to both v1 and v2 API types
- Comprehensive test coverage and examples

### Breaking Changes
None - fully backward compatible addition.
    

---

## Summary by cubic
Adds extractDataAttributes to v1/v2 scrape endpoints to pull data-* attributes via CSS selectors. Responses now include dataAttributes with the matched values, enabling structured scrapes without custom parsers. Addresses ENG-3197 and closes #1981.

- **New Features**
  - New request option: extractDataAttributes [{ selector, attribute }] in v1 and v2.
  - Response includes dataAttributes [{ selector, attribute, values: [] }].
  - Attribute can include or omit the data- prefix; both are supported.
  - Supports multiple matches per selector; returns arrays.
  - Added transformer to the scraping pipeline, with tests and usage example.
  - Backward compatible; no breaking changes.

### DIFF
--- a/apps/api/src/controllers/v1/types.ts
+++ b/apps/api/src/controllers/v1/types.ts
@@ -834,7 +834,6 @@ export type Document = {
   json?: any;
   summary?: string;
   warning?: string;
-
   actions?: {
     screenshots?: string[];
     scrapes?: ScrapeActionContent[];


apps/api/src/examples/extract-data-attributes.ts
```diff
@@ -0,0 +1,132 @@
+/**
```
mogery: Not necessary, this should be turned into brief documentation in the `firecrawl-docs` repo

apps/api/src/controllers/v1/types.ts
```diff
@@ -357,6 +357,10 @@ const baseScrapeOptions = z
     mobile: z.boolean().default(false),
     parsePDF: z.boolean().default(true),
     actions: actionsSchema.optional(),
+    extractDataAttributes: z.array(z.object({
```
mogery: No need to add to v1

apps/api/src/controllers/v2/types.ts
```diff
@@ -293,6 +293,10 @@ const baseScrapeOptions = z
     mobile: z.boolean().default(false),
     parsers: parsersSchema.optional(),
     actions: actionsSchema.optional(),
+    extractDataAttributes: z.array(z.object({
```
mogery: Should not be a separate property on its own -- should be in `formats` instead. See JSON for example, let's call it `attributes` since it doesn't have to be a `data-` to work, and could be useful for other things too.

apps/api/src/controllers/v2/types.ts
```diff
@@ -634,6 +638,11 @@ export type Document = {
   json?: any;
   summary?: string;
   warning?: string;
+  dataAttributes?: Array<{
```
mogery: Rename to `attributes` here too

apps/api/src/scraper/scrapeURL/lib/extractDataAttributes.ts
```diff
@@ -0,0 +1,72 @@
+import { load } from "cheerio";
```
mogery: Logic should be moved to `html-transformer` Rust shared library if you're up for it, since cheerio can freeze the event loop.

apps/api/src/__tests__/e2e/data-attributes.test.ts
```diff
@@ -0,0 +1,156 @@
+import { describe, it, expect } from '@jest/globals';
```
mogery: Wrong place to put the test, integrate it into `snips/v2/scrape.test.ts` using the existing helpers there

mogery (changes requested):
Good start! API schema goes against the design we're trying to achieve with v2 -- you had no way of knowing that beforehand, so it's fine :) Rest are nitpicks and perf concerns.

mogery:
Also, needs to be added to JS and Python SDKs!

vishkrish200:
i also want to squash the commits. seems a bit messy right now

mogery:
@vishkrish200 It's all good, we merge with squash always

apps/api/src/controllers/v1/types.ts
```diff
@@ -834,6 +834,7 @@ export type Document = {
   json?: any;
   summary?: string;
   warning?: string;
+
```
mogery:
```suggestion
```

apps/api/src/controllers/v2/types.ts
```diff
@@ -219,7 +229,8 @@ export type FormatObject =
   | { type: "summary" }
   | JsonFormatWithOptions
   | ChangeTrackingFormatWithOptions
-  | ScreenshotFormatWithOptions;
+  | ScreenshotFormatWithOptions
+  | AttributesFormatWithOptions;
```
mogery:
```suggestion
  | AttributesFormatWithOptions
```
preventing future multidiffs

mogery:
Closing to re-fork

PR #2009, opened 2025-08-22
Minor fix, 503 status code

jsmaxi:
Minor fix, 503 status code


mogery:
What are the effects of this?

jsmaxi:
Consistency with the the existing code comments. 500 indicates an unexpected internal server error, whereas 503 means the service is temporarily unavailable (e.g., due to overload or maintenance) and may succeed if retried later.
If you have an established method, please test this end-to-end, before approval and merge.

Issue #1912, opened 2025-08-03
[Bug] JS SDK network issue


khanal-samir:
# Intermittent "No response received while trying to scrape URL.This may be a network error or the server is unreachable." Error with Firecrawl JS SDK

**Describe the Bug**
The Firecrawl JS SDK intermittently fails with a "No response received while trying to scrape URL.This may be a network error or the server is unreachable." error. The same configuration and URL work successfully sometimes but fail at other times, suggesting this may be a network timeout or server availability issue rather than a configuration problem.

**To Reproduce**
Steps to reproduce the issue:
1. Configure Firecrawl with the following setup:
```javascript
const app = new FirecrawlApp({ apiKey: SERVER_ENV.FIRE_CRAWL_API_KEY });
```
2. Call `scrapeUrl` with various options including JSON schema, proxy settings, and timeout configurations
3. Execute the scraping operation multiple times with the same URL and parameters
4. Observe that some requests succeed while others fail with the network error

**Expected Behavior**
The scraping operation should consistently succeed or fail based on the URL validity and configuration, not intermittently fail due to network issues. If there are server-side timeouts or rate limits, these should be handled gracefully with appropriate error messages.

**Screenshots**
```
Error: No response received while trying to scrape URL. This may be a network error or the server is unreachable.
    at FirecrawlApp.handleError (/home/samir/Documents/work-2/reasonyx/node_modules/.pnpm/@mendable+firecrawl-js@1.29.1/node_modules/@mendable/firecrawl-js/dist/index.cjs:946:13)
    at FirecrawlApp.scrapeUrl (/home/samir/Documents/work-2/reasonyx/node_modules/.pnpm/@mendable+firecrawl-js@1.29.1/node_modules/@mendable/firecrawl-js/dist/index.cjs:226:12)
    at getFirecrawlScrapper (/home/samir/Documents/work-2/reasonyx/dist/apps/backend/webpack:/apps/backend/src/template/import/api/firecrawl.api.ts:17:25)
    at ImportService.importFromFirecrawlScrape (/home/samir/Documents/work-2/reasonyx/dist/apps/backend/webpack:/apps/backend/src/template/import/import.service.ts:444:18)
```

**Environment:**
- OS: Linux Mint
- Firecrawl Version: @mendable/firecrawl-js@1.29.1
- Backend: Nestjs
- Package manager: pnpm 9.12.3

**Logs**
The error occurs at the `scrapeUrl` method call with the following configuration:
```javascript
[...]

PR #2013, opened 2025-08-23
[Bug] JS SDK network issue #1912 - Intermittent "No response received while trying to scrape URL" Error

Jeelislive:
Fixed intermittent network failures in Firecrawl JS SDK with automatic retry mechanism and exponential backoff

The Firecrawl JS SDK was experiencing intermittent failures with the error message "No response received while trying to scrape URL. This may be a network error or the server is unreachable." This occurred even when the same configuration and URL worked successfully at other times, indicating transient network issues rather than configuration problems.

solution --> 

- Added automatic retry logic with up to 3 attempts for network errors
- Implemented exponential backoff with jitter (1s, 2s, 4s delays)
- Enhanced error classification to distinguish retryable vs non-retryable errors
- Improved timeout handling with sensible defaults (30 seconds)
- Added support for retrying on specific HTTP status codes (408, 429, 502, 503, 504)
- Maintained 100% backward compatibility

    

---

## Summary by cubic
Fixes intermittent “No response received while trying to scrape URL” errors (issue #1912) by adding automatic retries with exponential backoff and improved timeout/error handling. This makes scraping and SDK network calls more reliable under transient failures.

- **Bug Fixes**
  - Added automatic retry (up to 3 attempts) with exponential backoff + jitter for scrape, POST, and GET requests.
  - Retries on network errors and HTTP 408/429/502/503/504; clearer separation of retryable vs non-retryable errors.
  - Default timeout set to 30s (respects per-call overrides with a small buffer); better final error messages after all retries.
  - Added apps/js-sdk/test-retry-logic.js to simulate flaky networks; no breaking changes.

### DIFF
--- a/apps/js-sdk/firecrawl/src/v1/index.ts
+++ b/apps/js-sdk/firecrawl/src/v1/index.ts
@@ -623,7 +623,7 @@ export default class FirecrawlApp {
   }
 
   /**
-   * Scrapes a URL using the Firecrawl API.
+   * Scrapes a URL using the Firecrawl API with automatic retry on network errors.
    * @param url - The URL to scrape.
    * @param params - Additional parameters for the scrape request.
    * @returns The response from the scrape operation.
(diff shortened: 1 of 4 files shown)


Jeelislive:
@mogery  
hey i've removed all 5xx code inside retryable method can you re-review please

Issue #2024, opened 2025-08-25
[Bug] v2 /search won't include summary response with 'scrapeOptions: { formats: [ 'summary' ] }'


subuta:
**Describe the Bug**

Call v2 '/search' endpoint with 'scrapeOptions: { formats: [ 'summary' ] }' won't include 'summary' key in response.

**To Reproduce**

by Node.js

```node
const url = 'https://api.firecrawl.dev/v2/search';
const options = {
    method: 'POST',
    headers: {Authorization: `Bearer ${FIRE_CRAWL_API_KEY}`, 'Content-Type': 'application/json'},
    body: JSON.stringify({
        query: 'site:firecrawl.dev',
        limit: 1,
        scrapeOptions: {
            formats: ['summary']
        }
    })
};

try {
    const response = await fetch(url, options);
    const data = await response.json();
    console.log(JSON.stringify(data, null, 2));
} catch (error) {
    console.error(error);
}
```

and this is what I get in my console.

```json
{
  "success": true,
  "data": {
    "web": [
      {
        "url": "https://www.firecrawl.dev/",
        "title": "Firecrawl - The Web Data API for AI",
        "description": "The web crawling, scraping, and search API for AI. Built for scale. Firecrawl delivers the entire internet to AI agents and builders.",
        "position": 1,
        "metadata": "{...omitted...}"
      } 
    ]
  },
  "creditsUsed": 1
}
```

**Expected Behavior**

Include 'summary' field like this in each search result (of 'web').
This is the 'summary' I got from v2 '/scrape' endpoint.

```
{
  "summary": "Firecrawl has successfully raised Series A funding and launched Firecrawl v2, an open-source web scraping tool designed to convert websites into LLM-ready data for AI applications. Key features include scraping, searching, and crawling capabilities, with support for various data formats like JSON and Markdown. The platform is trusted by over 5000 companies and emphasizes ease of use with zero configuration, fast performance, and the ability to handle dynamic content. Firecrawl is integrated with popular tools and offers a community-driven development approach. Users can benefit from use cases such as AI chat enhancements, lead enrichment, and deep research."
}
```

**Environment (please complete the following information):**

- macOS Sonoma 14.6.1

I can also reproduce this error at official playground.

<img width="678" height="509" alt="Image" src="https://github.com/user-attachments/assets/723e99f1-2eeb-4613-8b8b-40ef9f4adbba" />

mogery:
Fixed by #2027

[...]

PR #2027, opened 2025-08-25
fix(search): include 'summary' in v2 /search response when using scrapeOptions.formats (#2024)

Jeelislive:
This PR fixes issue #2024 where the v2 /search endpoint did not include the 'summary' field 
in results when using scrapeOptions: { formats: ['summary'] }.

- Added support to properly attach summary field to each search result
- Ensured consistent behavior with v2 /scrape endpoint
- Verified fix in API playground and local environment

    

---

## Summary by cubic
Fixes missing summary in v2 /search results when scrapeOptions.formats includes 'summary'. Responses now include summary for each result, matching v2 /scrape and addressing #2024.

### DIFF
--- a/apps/api/src/controllers/v2/search.ts
+++ b/apps/api/src/controllers/v2/search.ts
@@ -318,6 +318,7 @@ export async function searchController(
           rawHtml: doc.rawHtml,
           links: doc.links,
           screenshot: doc.screenshot,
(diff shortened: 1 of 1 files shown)


mogery (approved):

mogery:
Thanks!

Issue #1977, opened 2025-08-19
[Bug] Playwright scrape fails: browserContext.newPage crashes with "Target page, context or browser has been closed


smarkars:
**Describe the Bug**

When scraping with Playwright, the job fails with the error and the service crashes

```
browserContext.newPage: Target page, context or browser has been closed
```

This happens during `scrapeURLWithPlaywright`, and the worker crashes without retrying.

---

**To Reproduce**

Steps to reproduce the issue:

1. Configure a self-hosted Firecrawl instance using docker. (Playwright image both build and the prebuilt image having issue)
2. Run a scrape job, I have configured this as a search engine and WebLoader in OpenWebUI. The first page successfully get scraped.
3. Observe the worker logs:

   * `Scraping via playwright...`
   * `browserContext.newPage: Target page, context or browser has been closed`
4. The scrape fails, and the worker does not recover.

**Screenshots / Logs**
Log snippet:

```
playwright-service-1  | browserContext.newPage: Target page, context or browser has been closed
playwright-service-1  |     at /usr/src/app/dist/api.js:185:39
playwright-service-1  |     at fulfilled (/usr/src/app/dist/api.js:5:58)
```
---

**Environment**

* OS: Ubuntu 22.04
* Deployment Type: Self-hosted
* Firecrawl Version: latest (build from docker file)
* Node.js Version: v18.20.8

grahamholley:
^ Having the same issue running the containers locally. Have not tried deploying them yet.

zliebersbach:
Yup identical problem here. Which makes it unusable at scale.

grahamholley:
I was able to fix this issue by removing the --single-process argument from the chromium browser instantiation in apps/playwright-service-ts/api.ts. You may also need to increase the shared memory in the playwright-service docker-compose.yml: 
shm_size: '2g'

Not sure to what extent this will affect performance but it did solve the browser context issue for me. I assume it was configured that way for a reason so not going to raise a PR for that.

zliebersbach:
Amazing @grahamholley worked for me! Nice job figuring out the problem, I tweaked the flags too but couldn't get it.

mogery:
Weird! @grahamholley would appreciate if you opened a PR! Thanks!

PR #2039, opened 2025-08-25
fix(scrape): Remove --single-process argument from chrome in playwright-service

grahamholley:
Fixes: #1977 
Bug:
* When scraping with Playwright, the job fails and the service crashes with error: "browserContext.newPage: Target page, context or browser has been closed".
* Occurs from the second attempt to scrape with playwright onward.
Changes:
* Removed the --single-process argument from chrome in playwright-service-ts.


mogery:
Thank you!

grahamholley:
@mogery np, happy to help!

PR #2040, opened 2025-08-25
Fix Playwright Content-Type Handling

amplitudesxd:
The old implementation checked the second character of the `Content-Type` header instead of the actual value. Updated to properly detect JSON/plain text.


amplitudesxd:
@cubic-dev-ai rerun

mogery (approved):
LGTM, thank you!

PR #2085, opened 2025-09-01
Fix sitemap parsing (ENG-3361)

amplitudesxd:
## Summary by cubic
Fix sitemap discovery and robots.txt handling so /map returns the full set of site URLs instead of stalling around ~200 (ENG-3361). We now parse sitemaps from robots.txt and the root, respect robots rules, and feed those URLs into map results with caching.

- **Bug Fixes**
  - Discover and parse sitemaps from robots.txt and /sitemap.xml, add URLs to map results, and cache sitemap responses for 48h when called from /map.
  - Apply robots.txt allow/deny consistently (including trailing-slash cases) and honor crawl-delay; allow opt-out via ignoreRobotsTxt.
  - Filter strictly by domain and the requested path (when not mapping the root); drop external, social, and anchor-only links; improve backward-crawling checks and URL normalization.
  - Add a robots.txt parser test to verify directives, allow/deny, and sitemap URLs.

- **Refactors**
  - Replace axios with undici and a shared secure dispatcher across robots, scraping, file downloads, and webhooks; proper TLS-skip support.
  - Improve charset decoding and error handling in fetch paths; minor logging cleanups.
  - Batch webhook log inserts and send webhooks via the shared dispatcher.

### DIFF
--- a/apps/api/src/__tests__/snips/v2/robots-txt.test.ts
+++ b/apps/api/src/__tests__/snips/v2/robots-txt.test.ts
@@ -1,15 +1,5 @@
 import { createRobotsChecker, fetchRobotsTxt } from "../../../lib/robots-txt";
-import { scrapeTimeout, idmux, Identity } from "./lib";
-
-let identity: Identity;
-
-beforeAll(async () => {
-  identity = await idmux({
-    name: "robots-txt",
-    concurrency: 100,
-    credits: 1000000,
-  });
-}, 10000);
+import { scrapeTimeout } from "./lib";
 
 describe("Robots.txt tests", () => {
   const robotsUrl = "https://www.google.com/robots.txt";


apps/api/src/__tests__/snips/v2/robots-txt.test.ts
```diff
@@ -0,0 +1,72 @@
...
+
+let identity: Identity;
+
+beforeAll(async () => {
+  identity = await idmux({
+    name: "robots-txt",
+    concurrency: 100,
+    credits: 1000000,
+  });
+}, 10000);
```
mogery: Please remove the `idmux` call -- since you don't call the API here, you don't need an Identity

mogery (approved):

PR #2120, opened 2025-09-09
Fix graceful credit handling in account object (ENG-3495)

amplitudesxd:
## Summary by cubic
Fixes remainingCredits computation in auth middleware to include grace credits when price_should_be_graceful is true. Prevents crawls from stopping early under grace mode (ENG-3495).

- **Bug Fixes**
  - Set req.account.remainingCredits to remaining_credits + price_credits when in graceful pricing; otherwise use remaining_credits.


mogery (approved):
o nice

PR #2129, opened 2025-09-09
feat(api): return job IDs consistently

gustavovalverde:
## The problem

Right now, v2 endpoints are inconsistent about returning job IDs:
- Async endpoints like crawl and batch-scrape always return an `id`
- Sync endpoints like scrape and search sometimes don't return IDs

This makes it hard when you're building an application, because you can't reliably get the job ID to track what happened.

## Rational

When an application triggers a Firecrawl job, we need the job ID so we can:
- Link it back to our own records ("this user request corresponds to Firecrawl job X")
- Check if the job succeeded or failed
- Look up the job in Firecrawl logs when something goes wrong
- Potentially: Show users a link to the Firecrawl dashboard for their job

**NOTE:** Ideally, even failed jobs should return the jobid, but that's a different thing and it required more analysis. We can discuss that here.

## What changed

All V2 sync endpoints now return job IDs:
- `/v2/scrape` - returns `id` and `scrape_id`
- `/v2/search` - returns `id` for both sync and async modes
- `/v2/map` - returns `id` alongside the existing `job_id`

## Backward compatibility

No breaking changes. All the existing fields are still there (`scrape_id`, `scrapeIds`, `job_id`). We just added the `id` field to make things consistent.
    

---

## Summary by cubic
Return job IDs consistently across v2 endpoints so apps can track and debug jobs reliably. Adds id to scrape, search, and map responses without breaking existing fields.

- **New Features**
  - v2/scrape: adds id alongside scrape_id (for website-origin requests).
  - v2/search: returns id in both sync and async responses.
  - v2/map: returns id alongside existing job_id.
  - Updated TypeScript response types to include the new id fields.

- **Migration**
  - No changes required. Existing fields (scrape_id, scrapeIds, job_id) remain.

### DIFF
--- a/apps/api/src/controllers/v2/batch-scrape.ts
+++ b/apps/api/src/controllers/v2/batch-scrape.ts
@@ -29,6 +29,7 @@ export async function batchScrapeController(
   req: RequestWithAuth<{}, BatchScrapeResponse, BatchScrapeRequest>,
   res: Response<BatchScrapeResponse>,
 ) {
+  const id = req.body?.appendToId ?? uuidv4();
   const preNormalizedBody = { ...req.body };
   if (req.body?.ignoreInvalidURLs === true) {
     req.body = batchScrapeRequestSchemaNoURLValidation.parse(req.body);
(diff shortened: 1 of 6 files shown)


mogery:
I like it! We were planning to do this. A few notes:
- Can you get rid of it in `/v2/scrape`? We already have `metadata.scrapeId` there.
- This needs to happen when a running job (e.g. scrape) errors too. `ErrorResponse` should have an `id` field and it should be routed correctly.

gustavovalverde:
> * Can you get rid of it in `/v2/scrape`? We already have `metadata.scrapeId` there

@mogery I added so the responses are consistent (all returning `id`). This one is the only one that's not consistent with the others. I can't remove the previous value as that would be a breaking change in the API. I could add a TODO though, to deprecate that key, and it can be removed in v3.

PR #2131, opened 2025-09-10
Perform watching inside of harness (ENG-3514)

amplitudesxd:
prevents rebuilds of natives etc every time files are changed (use `dev` or `start:production` script)

### DIFF
--- a/apps/api/src/harness.ts
+++ b/apps/api/src/harness.ts
@@ -3,6 +3,7 @@ import { type ChildProcess, spawn } from "child_process";
 import * as net from "net";
 import { basename } from "path";
 import { HTML_TO_MARKDOWN_PATH } from "./natives";
+import { createWriteStream } from "fs";
 
 const childProcesses = new Set<ChildProcess>();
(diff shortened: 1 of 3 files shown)


apps/api/src/harness.ts
```diff
@@ -357,58 +338,137 @@ async function startServices() {
...
   console.error(
     `${colors.bold}Usage:${colors.reset} pnpm harness <command...>\n`,
   );
   console.error(`${colors.bold}Special commands:${colors.reset}`);
-  console.error(`  --start        Start services and wait for termination`);
-  console.error(`  --start-built  Start services without rebuilding`);
-  console.error(`  --start-docker Start services (skip build completely)\n`);
   console.error(
-    `The harness ensures dependencies are installed, everything is built,`,
+    `  --start        Start in development mode (auto-restart on changes)`,
```
mogery, line 474: There should still be a `--start` option that does not auto-restart on changes (my preference, sorry!)

mogery (changes requested):
It's not running the command, check CI logs (e.g. production)

PR #2132, opened 2025-09-10
Add gzipped sitemap support (ENG-3520)

amplitudesxd:
## Summary by cubic
Adds support for gzipped sitemaps (.xml.gz) so we can crawl compressed sitemap indexes and URL lists. This fulfills ENG-3520 by expanding sitemap discovery and parsing coverage.

- **New Features**
  - Detects .xml.gz in crawler sitemap URL resolution and logs.
  - Downloads and decompresses .xml.gz with fetchFileToBuffer + zlib.gunzip, then parses XML.
  - Treats .xml.gz as valid sitemap entries in indexes; filters non-sitemap URLs accordingly.

- **Bug Fixes**
  - Fixes reduce usage when aggregating child sitemap counts by adding an initial value.

### DIFF
--- a/apps/api/src/scraper/WebScraper/crawler.ts
+++ b/apps/api/src/scraper/WebScraper/crawler.ts
@@ -14,7 +14,6 @@ import {
   isUrlAllowedByRobots,
 } from "../../lib/robots-txt";
 import { ScrapeJobTimeoutError } from "../../lib/error";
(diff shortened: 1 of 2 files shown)


amplitudesxd:
seems best to be left for the test site, couldn't find any big sites using .gz sitemaps

Issue #2139, opened 2025-09-11
[Self-Host] No data returned for search endpoint


ay-cw-gbio:
**Describe the Issue**
Been using the self-hosted version of firecrawl for the crawl, scrape and search endpoints for over a month. Yesterday, I pulled the latest version of the repo and the search endpoint is no longer working. 
Crawl and scrape still work fine, but I'm also no longer seeing any jobs on _http://localhost:3002/admin/CHANGEME/queues/_

I followed all the instructions provided here - https://github.com/firecrawl/firecrawl/blob/main/SELF_HOST.md

**Expected Behavior**
I've tried different configurations (changed the query, limit, also tried the v2/search endpoint instead of v1/search). Every single time, it says that there's no search results found. It should be returning scraped data after the search and it was a few weeks ago. 

**Screenshots**
Sample terminal output - 
```
# V1 / SEARCH 
curl -X POST http://localhost:3002/v1/search \
    -H 'Content-Type: application/json' \
    -d '{
      "query": "what is firecrawl?",
        "limit": 3
    }'
{"success":true,"data":[],"warning":"No search results found"}

curl -X POST http://localhost:3002/v1/search \
  -H "Content-Type: application/json" \
  -d '{
    "query": "How is icecream made?",
    "limit": 2,
    "scrapeOptions": {
      "formats": ["markdown", "links"]
    }
  }'

{"success":true,"data":[],"warning":"No search results found"}

# V2 / SEARCH 
curl -X POST http://localhost:3002/v2/search \
  -H "Content-Type: application/json" \
  -d '{
    "query": "Most spoken languages",
    "limit": 5,
    "scrapeOptions": {
      "formats": ["markdown", "links"]
    }
  }'

{"success":true,"data":{},"creditsUsed":0}
```

**Environment:**
- OS: macOS
- Firecrawl Version: [v2.1.0 (updated yesterday)]
- Docker Version: [v4.45.0 (re-installed yesterday)]
- Node.js Version: v24.5.0

**Logs**
Docker Logs:

api          2025-09-11 17:33:24 info [search:searchController]: Searching completed 

api          2025-09-11 17:33:24 info [search:searchController]: No search results found 

api          2025-09-11 17:33:24 info [search:searchController]: Logging job

grahamholley:
[...]

PR #2152, opened 2025-09-14
[UPDATE]: Prioritize SearXNG for self-host

Chadha93:
Fixes: #2139 
use existing SearXNG service in docker-compose with multi-engine support

tested OK:
v2 - 
image
v1 - 
image


Chadha93:
Already resolved

PR #2159, opened 2025-09-15
Custom Rust document parser

amplitudesxd:
Makes document parsing ~75x faster + adds support for RTF & ODT


amplitudesxd:
tests done - tried doing them in Rust but they failed to compile on Linux, was fine on Windows however which is odd... doing them in JS for now

mogery (approved):
coolio

PR #2171, opened 2025-09-17
Add unwrapTags feature (ENG-3563)

amplitudesxd:
sometimes a page may include something similar to:

```html
<div>
  <code>
    <a href="https://example.com">test</a>
  </code>
</div>
```

the a tag will not get parsed since it's inside of a code block, so it wouldn't be semantically valid.

unwrapping with the `code` tag would make the above html snippet resolve to:

```html
<div>
  <a href="https://example.com">test</a>
</div>
```

and result in the `a` tag being included in the markdown
    

---

## Summary by cubic
Adds an unwrapTags option to the HTML transformer and scraper to remove specified wrapper elements and keep their children. Fixes ENG-3563 by ensuring links wrapped in tags like code are preserved in scraped output.

- **New Features**
  - Added unwrapTags to TransformHtmlOptions (native/html.rs) to unwrap matched elements and keep their children.
  - Exposed unwrapTags in v1 and v2 API schemas; accepts CSS selectors (trimmed and filtered).
  - Applied unwrapping in scraper htmlTransform before exclusion to rescue links and other content inside wrapper tags.
  - Updated tests to cover unwrapping behavior and interaction with includeTags/excludeTags.

### DIFF
--- a/apps/api/native/src/html.rs
+++ b/apps/api/native/src/html.rs
@@ -339,6 +339,8 @@ pub struct TransformHtmlOptions {
   pub include_tags: Vec<String>,
   #[serde(default)]
   pub exclude_tags: Vec<String>,
+  #[serde(default)]
+  pub unwrap_tags: Vec<String>,
   pub only_main_content: bool,
   pub omce_signatures: Option<Vec<String>>,
 }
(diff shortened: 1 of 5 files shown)


mogery:
This is a weird one. I feel like we should be trying to get the most out of the HTML in general, so I would classify this as an HTML parser-level issue that we should deal with automatically, rather than expose to the user to mess with.

Issue #2173, opened 2025-09-17
[Bug] Docker container no longer builds on linux/arm64/v8


zliebersbach:
**Describe the Bug**
Docker container no longer builds on linux/arm64/v8.

Last successful build was approx 2 weeks ago.

**To Reproduce**
Steps to reproduce the issue:
1. `git clone https://github.com/firecrawl/firecrawl`
2. `cd firecrawl`
3. `docker compose up -d --build`

**Expected Behavior**
The container builds.

**Environment (please complete the following information):**
- OS: Linux
- Deployment Type: Self-hosted
- Firecrawl Version: main branch
- Node.js Version: N/A

**Logs**
<details>
<summary>Click to expand logs</summary>

```
 => ERROR [api build 6/8] RUN --mount=type=cache,id=pnpm,target=/pnpm/store     --mount=type=cache,target=/usr/local/carg  72.2s
------
 > [api build 6/8] RUN --mount=type=cache,id=pnpm,target=/pnpm/store     --mount=type=cache,target=/usr/local/cargo/registry     --mount=type=cache,target=/app/native/target     pnpm install --frozen-lockfile:
0.284 ! Corepack is about to download https://registry.npmjs.org/pnpm/-/pnpm-10.16.1.tgz
1.726 Scope: all 2 workspace projects
1.861 Lockfile is up to date, resolution step is skipped
2.023 Progress: resolved 1, reused 0, downloaded 0, added 0
2.282 Packages: +1409
2.282 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
3.025 Progress: resolved 1409, reused 303, downloaded 0, added 0
4.026 Progress: resolved 1409, reused 649, downloaded 0, added 0
5.050 Progress: resolved 1409, reused 830, downloaded 0, added 0
6.051 Progress: resolved 1409, reused 1138, downloaded 0, added 0
7.051 Progress: resolved 1409, reused 1404, downloaded 0, added 21
8.052 Progress: resolved 1409, reused 1404, downloaded 0, added 54
9.052 Progress: resolved 1409, reused 1404, downloaded 0, added 102
10.05 Progress: resolved 1409, reused 1404, downloaded 0, added 189
11.05 Progress: resolved 1409, reused 1404, downloaded 0, added 219
12.05 Progress: resolved 1409, reused 1404, downloaded 0, added 275
13.05 Progress: resolved 1409, reused 1404, downloaded 0, added 310
[...]

PR #2181, opened 2025-09-18
fix: [Dockerfile] Install Python and configure for node-gyp

KaranocaVe:
This pull request updates the Dockerfile to improve compatibility with Node.js native modules that require Python during build steps. The main change is the installation and configuration of Python 3 for use with `node-gyp`.

see #2173 

Dependency and build environment updates:

* Added installation of `python3` and `python3-dev` packages to ensure Python 3 is available in the build environment.
* Set the `PYTHON` environment variable to `/usr/bin/python3` to configure Python for `node-gyp` builds.
    

---

## Summary by cubic
Install Python 3 in the API Docker image and configure node-gyp so Node.js native modules build successfully.

- **Dependencies**
  - Add python3 and python3-dev to the image.
  - Set PYTHON=/usr/bin/python3 for node-gyp.

### DIFF
--- a/apps/api/Dockerfile
+++ b/apps/api/Dockerfile
@@ -24,7 +24,12 @@ RUN apt-get update && apt-get install -y \
     curl \
     build-essential \
     pkg-config \
+    python3 \
+    python3-dev \
     && rm -rf /var/lib/apt/lists/*
+
+# Configure Python for node-gyp
+ENV PYTHON=/usr/bin/python3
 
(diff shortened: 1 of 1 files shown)


amplitudesxd:
Hey! Thanks so much for taking the time to put this together!

It looks like only the `python3` package is needed for the builds to succeed, so I’ll close this in favor of #2198, which makes the same change with just a few fewer adjustments.

PR #2186, opened 2025-09-19
feat(api): move blocklist to DB

mogery:


mogery:
@cubic-dev-ai rereview pls

PR #2209, opened 2025-09-25
fix: Use port from env.PORT for API

### LINKED ISSUE #2204: [Bug] PORT env variable is ignored by harness; startup fails if changed
**Describe the Bug**
There is a mismatch between the code that starts the API service and the process manager (harness). The Express server in `apps/api/src/index.ts` respects the `PORT` environment variable, and will listen on whatever value it is set to. However, the harness script (`apps/api/src/harness.ts`) is hardcoded to wait for port 3002, regardless of the `PORT` env variable. As a result, if you change the port using the env variable, the API will not be detected as ready, causing a fatal error on startup:

```
Error: Port 3002 did not become available within 20000ms
```

**To Reproduce**
Steps to reproduce the issue:
1. Set the `PORT` environment variable to something other than 3002 (e.g. `PORT=4000`)
2. Start the service via Docker or the harness script
3. Observe that the API listens on the configured port, but the harness waits for 3002
4. See the fatal error about port 3002 not being available

**Expected Behavior**
If the `PORT` environment variable is set, both the Express server and the harness should use that value, so that startup works on any valid port.

**Screenshots**
If applicable, add screenshots or copies of the command line output to help explain the issue.

- OS: [e.g. macOS, Linux, Windows]
- Deployment Type: [Cloud (firecrawl.dev) / Self-hosted]
- Firecrawl Version: [e.g. 1.2.3]
- Node.js Version: [e.g. 14.x]

**Logs**
Example log output:
```
Error: Port 3002 did not become available within 20000ms
    at Timeout._onTimeout (/app/dist/src/harness.js:129:20)
    ...
```

**Additional Context**
- The API server in `apps/api/src/index.ts` does respect the `PORT` env variable.
- The harness hardcodes port 3002.
- This means the `PORT` env variable is effectively useless unless it is set to 3002, or the harness is updated to read the env var.

abimaelmartell:
This pull request updates the API harness to allow the port number to be configurable via an environment variable, instead of being hardcoded. This makes it easier to run the API on different ports and improves flexibility in various environments.

Configuration improvements:

* Introduced a new `PORT` constant in `apps/api/src/harness.ts` that reads the port number from the environment variable, defaulting to `"3002"` if not set.

Port usage updates:

* Updated all instances where the port was previously hardcoded (`3002`) to use the new `PORT` constant in log messages and when waiting for the API to be ready, both in development and production modes. [[1]](diffhunk://#diff-2e4ed36a25f740c3c1e68ec159a3082da40a8efa78f2f0e4145ca72cf4c010e3L403-R407) [[2]](diffhunk://#diff-2e4ed36a25f740c3c1e68ec159a3082da40a8efa78f2f0e4145ca72cf4c010e3L419-R422) [[3]](diffhunk://#diff-2e4ed36a25f740c3c1e68ec159a3082da40a8efa78f2f0e4145ca72cf4c010e3L445-R449)

Fixes #2204

### DIFF
--- a/apps/api/src/harness.ts
+++ b/apps/api/src/harness.ts
@@ -68,6 +68,9 @@ function formatDuration(nanoseconds: bigint): string {
 
 const stream = createWriteStream("firecrawl.log");
 
+// Get the port from environment variable, defaulting to 3002
+const PORT = process.env.PORT ?? "3002";
+
 const logger = {
   section(message: string) {
     console.log(
@@ -400,8 +403,8 @@ async function runDevMode(): Promise<void> {
 
     currentServices = startServices();
 
-    logger.info("Waiting for API on localhost:3002");
-    await waitForPort(3002, "localhost");
+    logger.info(`Waiting for API on localhost:${PORT}`);
+    await waitForPort(Number(PORT), "localhost");
     logger.success("API is ready");
 
     isFirstStart = false;
@@ -416,7 +419,7 @@ async function runDevMode(): Promise<void> {
 
       currentServices = startServices();
 
-      await waitForPort(3002, "localhost");
+      await waitForPort(Number(PORT), "localhost");
       logger.success("Services restarted");
     }
   });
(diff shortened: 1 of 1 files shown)


mogery:
Thanks!

PR #2230, opened 2025-09-30
feat(nuq): RabbitMQ support for job finish events and waiting

mogery:


mogery:
@cubic-dev-ai Is there any pattern in this that could break existing environments without a NUQ_RABBITMQ_URL configured?

PR #2233, opened 2025-09-30
feat(nuq/rabbitmq): add prefetching jobs from psql to rabbitmq

mogery:
## Summary by cubic
Adds a RabbitMQ-backed prefetch layer for NuQ. A new prefetch worker moves jobs from Postgres to a RabbitMQ prefetch queue, and workers consume from it for lower DB contention and faster job pickup. Supports ENG-3662 job distribution.

- New Features
  - Added nuq-prefetch-worker with /metrics and /health; loops prefetchJobs every 250ms.
  - Prefetches up to 500 queued jobs per batch, marks them active with a UUID lock, and publishes to queueName.prefetch (durable, TTL 30s, max 20k).
  - Workers now consume from the prefetch queue when NUQ_RABBITMQ_URL is set, falling back to Postgres if not. Uses job.lock for renew/finish/fail and avoids exponential backoff under RabbitMQ.
  - Sender/channel robustness: close handlers, renamed sendJobEnd, and prefetch(1) for listener channel.
  - Harness/scripts: added nuq-prefetch-worker scripts and lifecycle; auto-starts when NUQ_RABBITMQ_URL is present.
  - Tests updated to ignore new prefetch metrics logs.

- Migration
  - Set NUQ_RABBITMQ_URL to enable prefetching.
  - Run nuq-prefetch-worker (pnpm nuq-prefetch-worker:production). Default port 3011 (override with NUQ_PREFETCH_WORKER_PORT).
  - No schema changes; deploy alongside existing NuQ workers.

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -24,6 +24,8 @@
     "worker:production": "node --import ./dist/src/otel.js dist/src/services/queue-worker.js",
     "nuq-worker": "tsc-watch --onSuccess \"node --import ./dist/src/otel.js dist/src/services/worker/nuq-worker.js\"",
     "nuq-worker:production": "node --import ./dist/src/otel.js dist/src/services/worker/nuq-worker.js",
+    "nuq-prefetch-worker": "tsc-watch --onSuccess \"node --import ./dist/src/otel.js dist/src/services/worker/nuq-prefetch-worker.js\"",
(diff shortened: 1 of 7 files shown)


mogery:
@cubic-dev-ai Is there anything in this that could break pre-existing deployments without a NUQ_RABBITMQ_URL var?

PR #2237, opened 2025-09-30
js-sdk: require API key only for cloud API (not self-hosted)

abimaelmartell:
Update the v2 `FirecrawlClient` to only require an API key when using the **Cloud version of the API**.  

    

---

## Summary by cubic
Updated the v2 FirecrawlClient to require an API key only when using the cloud API (https://api.firecrawl.dev). Self-hosted deployments using a custom apiUrl can now run without an API key.

### DIFF
--- a/.github/workflows/publish-python-sdk.yml
+++ b/.github/workflows/publish-python-sdk.yml
@@ -17,7 +17,7 @@ jobs:
 
     steps:
       - name: Checkout repository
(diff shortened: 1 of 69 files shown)


mogery (approved):
Thanks! Can you bump the package version so it publishes when I merge?

abimaelmartell:
@mogery updated

PR #2238, opened 2025-10-01
fix: handle `$ref` for recursive schema validation

Chadha93:
ref: https://discord.com/channels/1226707384710332458/1421817976566452234
    

---

## Summary by cubic
Fix handling of $ref in schema normalization and validation to support recursive JSON Schemas. Prevents false errors and infinite recursion in OpenAI schema processing.

- **Bug Fixes**
  - Preserve $ref nodes as-is during normalization.
  - Skip recursion into $ref objects.
  - Treat $ref as valid in validation and update the error message to note recursive schema support.

### DIFF
--- a/apps/api/src/scraper/scrapeURL/transformers/llmExtract.ts
+++ b/apps/api/src/scraper/scrapeURL/transformers/llmExtract.ts
@@ -25,6 +25,46 @@ import { extractData } from "../lib/extractSmartScrape";
 import { CostTracking } from "../../../lib/cost-tracking";
 import { isAgentExtractModelValid } from "../../../controllers/v1/types";
 import { hasFormatOfType } from "../../../lib/format-utils";
+
+// Smart model selection based on schema
+function detectRecursiveSchema(schema: any): boolean {
+  if (!schema || typeof schema !== "object") return false;
+
+  const schemaString = JSON.stringify(schema);
+  const hasRefs =
+    schemaString.includes('"$ref"') ||
+    schemaString.includes("#/$defs/") ||
+    schemaString.includes("#/definitions/");
+  const hasDefs = !!(schema.$defs || schema.definitions);
+
+  return hasRefs || hasDefs;
+}
+
+function selectModelForSchema(schema?: any): {
+  modelName: string;
+  reason: string;
+} {
+  if (!schema) {
+    return { modelName: "gpt-4o-mini", reason: "no_schema" };
+  }
+
+  const isRecursive = detectRecursiveSchema(schema);
+
+  if (isRecursive) {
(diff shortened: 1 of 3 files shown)


Chadha93:
@mogery, could you please also test this on local production with the OpenAI key to verify the fix, as I don't have one with more credits.

to test, ref: https://discord.com/channels/1226707384710332458/1421817976566452234/1422645895580618834

mogery:
Does this also need to be put into the v1 types?

Chadha93:
> Does this also need to be put into the v1 types?

Nope, we can skip it, since in v1, the schemas are managed directly without validation.

Chadha93:
@mogery there was an issue of infinite recursion occurring, added a fix for that and tested OK:

sample schema you can try with:

```
curl -X POST http://localhost:3002/v2/scrape \
  -H "Content-Type: application/json" \
  -d '{
    "url": "https://www.mordorintelligence.com/industry-reports/canada-cloud-computing-market",
    "formats": [
      {
        "type": "json",
        "schema": {
          "type": "object",
          "properties": {
            "table_of_contents": {
              "type": "array",
              "items": {
                "$ref": "#/definitions/section"
              },
              "description": "Extract the document structure as a hierarchical table of contents"
            }
          },
          "required": ["table_of_contents"],
          "definitions": {
            "section": {
              "type": "object",
              "properties": {
                "title": {
                  "type": "string"
                },
                "children": {
                  "type": "array",
                  "items": {
                    "$ref": "#/definitions/section"
                  }
                }
              },
              "required": ["title"],
              "additionalProperties": false
            }
          }
        }
      }
    ]
  }'
```

apps/api/src/scraper/scrapeURL/transformers/llmExtract.ts
```diff
@@ -259,7 +259,7 @@ export async function generateCompletions({
   markdown,
   previousWarning,
   isExtractEndpoint,
-  model = getModel("gpt-4o-mini", "openai"),
+  model = getModel("gpt-4o", "openai"),
```
Chadha93: @mogery  Also, updated the model to gpt-4o, as mini doesn't work for recursive schema extraction, lmk if this is ok?

mogery: This is not OK. This will 10x our costs.

mogery: Is there a way to detect recursive schemas and only use `gpt-4o` if they are detected? Also print a log if this happens so we can monitor

Chadha93: yes, I've added model selection and logging function

Issue #1397, opened 2025-03-31
[Self-Host]


529507929:
Firecrawl version is v1.6.0 
I get the following error when using the '"ignoreSitemap": true' parameter in the /map endpoint.
```
2025-03-31 16:18:19 api-1                 | 2025-03-31 08:18:19 warn [:]: You're bypassing authentication {}
2025-03-31 16:18:19 api-1                 | 2025-03-31 08:18:19 warn [:]: You're bypassing authentication {}
2025-03-31 16:18:19 api-1                 | 2025-03-31 08:18:19 warn [:]: You're bypassing authentication {}
2025-03-31 16:18:19 api-1                 | 2025-03-31 08:18:19 warn [:]: You're bypassing authentication {}
2025-03-31 16:18:28 api-1                 | 2025-03-31 08:18:28 warn [:]: You're bypassing authentication {}
```
Here is my POST request
```
{
        "url": "https://www.foshannews.net/h/191/",
		"includeSubdomains": true,
        "ignoreSitemap": true
	}
```
The results are shown below
```
{
    "success": true,
    "links": [
        "https://www.foshannews.net/h/191"
    ]
}
```
Do I need additional settings to enable the "ignoreSitemap" feature?

zliebersbach:
Map seems to not exist in the Open Source repo:

```
api-1                 | 2025-08-06 01:21:58 info [:]: Map request 
api-1                 | 2025-08-06 01:22:00 warn [:]: (v1/map Beta) Results might differ from cloud offering currently. {}
api-1                 | 2025-08-06 01:22:00 warn [:]: (v1/map Beta) Results might differ from cloud offering currently. {}
api-1                 | 2025-08-06 01:22:00 warn [:]: (v1/map Beta) Results might differ from cloud offering currently. {}
api-1                 | 2025-08-06 01:22:00 warn [:]: (v1/map Beta) Results might differ from cloud offering currently. {}
api-1                 | 2025-08-06 01:22:00 warn [:]: (v1/map Beta) Results might differ from cloud offering currently. {}
```

zliebersbach:
Seems like this functionality is literally a stub: https://github.com/mendableai/firecrawl/blob/899c6b454b9a77c35cb7ad37a59c0e38a4e11696/apps/api/src/search/fireEngine.ts#L84-L87

I feel like it should raise a hard exception, because the endpoint is broken without Fire Engine.

Chadha93:
[...]

PR #2273, opened 2025-10-13
[self-host] fix: Add fallback for `sitemap` discovery due to Fire-engine limited access 

Chadha93:
ref: https://github.com/firecrawl/firecrawl/issues/1397#issuecomment-3396040607

tested OK:
image

    

---

## Summary by cubic
Add sitemap fallback in /map for self-hosted setups when Fire Engine isn’t available, preventing empty results. If search finds nothing and sitemap=skip, we now try sitemap discovery and log the outcome.

- **Bug Fixes**
  - Detect Fire Engine availability via FIRE_ENGINE_BETA_URL.
  - On empty search results with sitemap=skip, call crawler.tryGetSitemap with timeout/abort support.
  - Add debug logs for fallback start, success count, and failure.

### DIFF
--- a/apps/api/src/lib/map-utils.ts
+++ b/apps/api/src/lib/map-utils.ts
@@ -22,6 +22,7 @@ import { performCosineSimilarityV2 } from "./map-cosine";
 import { Logger } from "winston";
 import { logger } from "./logger";
 
+// Max Links that "Smart /map" can return
 const MAX_FIRE_ENGINE_RESULTS = 500;
 const MAX_MAP_LIMIT = 1000;
 


Chadha93:
This is just a workaround, not to display empty results, although a long-term permanent solution is to extend Fire-engine's capabilities.

mogery:
Well, this is an issue on self-hosting only, no? Then a fix is to simply not use `sitemap=skip`, no?

Chadha93:
> Well, this is an issue on self-hosting only, no? Then a fix is to simply not use `sitemap=skip`, no?

Right only for self-host, I think, then I'll just update this thing in the docs, but won't we be adding support to `sitemap=skip` later?

micahstairs:
What about a simpler fix where if it's being self-hosted and sitemap=skip is passed, a clearly worded error message is thrown before it is executed?

PR #2274, opened 2025-10-13
fix: Remove Redis TTL for expired crawls (status)

Chadha93:
ref: https://discord.com/channels/1226707384710332458/1426937118688546816
Ensures failed crawls expire from Redis and display correct status, while active crawls continue working normally.
    

---

## Summary by cubic
Removed the TTL reset in getCrawl so failed/expired crawls in Redis actually expire and show the correct status. Active crawls are unaffected.

- **Bug Fixes**
  - Stopped setting a 24h expire on "crawl:{id}" during reads.


mogery:
That is there so that jobs continually running refresh the Redis entry

PR #2284, opened 2025-10-14
feat: Implement Scraping Excel xlsx files

abimaelmartell:
## Summary by cubic
Add Excel scraping support. Detects .xlsx and .xls and converts each sheet to HTML with an H2 heading and a table. Supports ENG-3676.

- **New Features**
  - Added XlsxProvider (Rust, calamine) to parse workbooks and emit sheet headings and table rows/cells.
  - Extended DocumentType and ProviderFactory; exported xlsx module.
  - Updated scraper to detect via URL suffix and content-type (xlsx/xls).
  - Added sample.xlsx and a test asserting the exact HTML output.

- **Dependencies**
  - Added calamine = "0.26".

### DIFF
--- a/apps/api/src/scraper/scrapeURL/engines/document/index.ts
+++ b/apps/api/src/scraper/scrapeURL/engines/document/index.ts
@@ -86,10 +86,12 @@ export async function scrapeDocument(meta: Meta): Promise<EngineScrapeResult> {
   let response: Response;
   let buffer: Buffer;
   let proxyUsed: "basic" | "stealth" = "basic";
+  let tempFilePath: string | null = null;
 
   if (meta.documentPrefetch !== undefined && meta.documentPrefetch !== null) {
     // Use prefetched document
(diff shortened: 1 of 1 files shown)


apps/api/src/__tests__/snips/v2/document-converter.test.ts
```diff
@@ -43,4 +43,18 @@ describe("Document Converter tests", () => {
...
+    it("should convert XLSX document and return expected HTML", async () => {
+      const filePath = path.join(samplesDir, "sample.xlsx");
+      const fileBuffer = fs.readFileSync(filePath);
+      const html = await converter.convertBufferToHtml(
+        new Uint8Array(fileBuffer),
+        DocumentType.Xlsx,
+      );
+      expect(html).toBe(expectedHtml);
+    });
+  });
```
abimaelmartell, line 59: can't reuse test from above, since that expects a document in a single page format, this one is being converted to tables as excel does

PR #2287, opened 2025-10-14
fix(crawl): surface robots.txt warning reliably

ftonato:
## Summary by cubic
Show a clear robots.txt warning in crawl results by reliably recording blocked URLs and only surfacing the message after scraping completes. Prevents missing or premature warnings and covers both link-level and top-level robots rejections.

- **Bug Fixes**
  - Added recordRobotsBlocked helper to store blocked URLs in Redis with 24h expiration.
  - Scrape worker records robots blocks from denial reasons and from top-level "URL blocked by robots.txt" errors.
  - Crawl status warns only when status is not "scraping" and respects result count checks for base-domain warnings.


mogery (approved):
lgtm!

Issue #2075, opened 2025-09-01
[Feat] Allow SDK to Work Without API Key for Self-Hosted Instances


SeeYangZhi:
**Problem Description**
Currently, when self-hosting Firecrawl and using the Python SDK, the Firecrawl constructor requires an API key even if I provide a custom api_url for my self-hosted instance. This prevents using the SDK with a self-hosted API that does not require authentication.

```python
from firecrawl import Firecrawl

firecrawl = Firecrawl(api_url="https://example.com")

firecrawl.scrape(url, formats=["markdown"])
```

This throws an error saying an API key must be provided, even though my self-hosted instance does not require one.

**Proposed Feature**
Allow Firecrawl(api_url="https://example.com") to work without an API key when pointing to a self-hosted instance that does not require authentication.
For example:
```python
firecrawl = Firecrawl(api_url="https://example.com", api_key=None)
```
or make api_key optional when a custom api_url is provided.

**Use Case**
Users who self-host Firecrawl should be able to leverage the official SDK without needing to hack around API key checks. This improves developer experience and supports the self-hosting use case as intended.

jschmdt:
Same for me with local Docker setup on macOS. I wanted to test some things with the Firecrawl MCP Server via Inspector, but I get API error messages with every tool call:

> [error] {"message":"Request failed: API key is required. Set FIRECRAWL_API_KEY env or pass apiKey.","tool":"firecrawl_scrape","arguments":{"url":"URL_PLACEHOLDER","formats":["markdown"],"onlyMainContent":true,"includeTags":[],"excludeTags":[],"waitFor":0,"actions":[],"mobile":false,"skipTlsVerification":false,"removeBase64Images":false,"location":{},"storeInCache":true,"maxAge":172800000},"timestamp":"2025-09-08T03:54:10.207Z","duration":1}

cccadet:
From what I've seen, the version 2 client still has the following validation:

```
        if not api_key:
            raise ValueError(
                "API key is required. Set FIRECRAWL_API_KEY environment variable "
                "or pass api_key parameter."
            )
```

abimaelmartell:
this is fixed in the latest version, please reach back again if you find any issues in the future

Best,
Abi

PR #2290, opened 2025-10-15
python-sdk: Don't require API Key when running Self Hosted

abimaelmartell:
## Summary by cubic
Make API keys optional for self-hosted Python SDK usage. Keys are now only required when targeting the cloud API (api.firecrawl.dev).

- **Bug Fixes**
  - Enforce API key only for api.firecrawl.dev in sync and async clients.
  - Make api_key optional in ClientConfig; add Authorization header only when a key is present.
  - Update HttpClient and AsyncHttpClient to handle missing keys safely.
  - Add tests covering cloud (requires key) vs self-hosted (no key) behavior.

### DIFF
--- a/apps/python-sdk/firecrawl/__init__.py
+++ b/apps/python-sdk/firecrawl/__init__.py
@@ -17,7 +17,7 @@ from .v1 import (
     V1ChangeTrackingOptions,
 )
 
-__version__ = "4.4.0"
+__version__ = "4.5.0"
 
 # Define the logger for the Firecrawl project
 logger: logging.Logger = logging.getLogger("firecrawl")


abimaelmartell:
fixes #2075

apps/python-sdk/firecrawl/v2/client_async.py
```diff
@@ -47,11 +47,15 @@
 from .watcher_async import AsyncWatcher
 
 class AsyncFirecrawlClient:
+    @staticmethod
+    def _is_cloud_service(url: str) -> bool:
+        return "api.firecrawl.dev" in url.lower()
```
micahstairs, line 52: Can we use `process.env.CLOUD_SERVICE === 'true'` instead? Seems we already do it elsewhere?

micahstairs (changes requested):

apps/python-sdk/firecrawl/v2/client_async.py
abimaelmartell: hmm, i don't see that env in this repo, also this is for the SDK, so we don't really want to introduce a new env var for that if we can detect it from the url

nickscamara: This is fine

nickscamara (approved):

PR #2310, opened 2025-10-22
fix: Fix commit hook

abimaelmartell:
ignore those unused files to avoid having to `--no-verify`
    

---

## Summary by cubic
Fix pre-commit hook failures by updating Knip config to ignore specific unused files and a dependency. This removes the need to use --no-verify.

- **Bug Fixes**
  - Added src/services/search-index-db.ts and src/lib/search-index-client.ts to Knip ignore.
  - Added @pinecone-database/pinecone to ignoreDependencies.


mogery:
we should just remove those files instead of ignoring. i'll do it

mogery:
fixed in 3d2ddbf09a2d17fe75ba23433cb13e29947ce317

PR #2387, opened 2025-11-10
Fix failed extract jobs stuck in processing

ktsaou:
## Summary
- set Redis status to "failed" when extraction returns success: false without throwing
- add a focused Jest test ensuring the v2 extract status controller reports failed jobs correctly

## Testing
- pnpm test -- extract-status-controller

---
## Summary by cubic
Ensure extract jobs that return success: false are marked as failed instead of staying stuck in processing. This fixes stale statuses and makes the v2 extract status API report failures correctly.

- **Bug Fixes**
  - Mark Redis extract status as failed when the worker gets success: false (preserves the error message).
  - Add a Jest test for the v2 extract-status controller to confirm it returns success: false with status: failed, error, empty data, and expiresAt from Redis.

<sup>Written for commit 2513676236d464bdb7c6821a8853bf37fd1b825a. Summary will update automatically on new commits.</sup>

### DIFF
--- a/apps/api/src/controllers/__tests__/extract-status-controller.test.ts
+++ b/apps/api/src/controllers/__tests__/extract-status-controller.test.ts
@@ -1,0 +1,83 @@
+import { extractStatusController } from "../v2/extract-status";
+import { getExtract, getExtractExpiry } from "../../lib/extract/extract-redis";
+import { getExtractQueue } from "../../services/queue-service";
+import { supabaseGetJobByIdDirect } from "../../lib/supabase-jobs";
+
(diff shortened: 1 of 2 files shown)


ktsaou:
Closing this — current `main` already handles it. The extract worker's non-success branch in `apps/api/src/services/extract-worker.ts` now sets `status: "failed"` (alongside the error), so a failed extract is no longer left in `processing`. Every terminal path — success, non-success result, thrown error, and the DLQ handler — now writes an explicit terminal status. Fixed independently upstream, so closing.

PR #2427, opened 2025-11-20
fix(api): logJob for noq requests

delong3:
## Summary by cubic
Fixes job logging for NoQ requests by making logging non-blocking when skipNuq is true and ensuring GCS blob writes only happen when required.

- **Bug Fixes**
  - Added requiresBlob to logJob; save to GCS only if a bucket is set and either data retention allows it or a blob is required.
  - In scrape-worker, pass requiresBlob as !skipNuq; await logging for NuQ jobs, fire-and-forget for NoQ, and log errors if logging fails.

<sup>Written for commit 0573252cca4f906d5d5348c2e0277d05bb0c3a81. Summary will update automatically on new commits.</sup>

### DIFF
--- a/apps/api/src/services/logging/log_job.ts
+++ b/apps/api/src/services/logging/log_job.ts
@@ -18,6 +18,7 @@ export async function logJob(
   job: FirecrawlJob,
   force: boolean = false,
   bypassLogging: boolean = false,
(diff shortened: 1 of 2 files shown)


mogery:
closing this as it conflicts with #2445

Issue #2403, opened 2025-11-15
[Self-Host] ExtractResponse pydantic parsing error


HadiSDev:
**Describe the Issue**
```
I get this error when using the extract api with python sdk:
return ExtractResponse(**resp.json())
pydantic_core._pydantic_core.ValidationError: 1 validation error for ExtractResponse error
Input should be a valid string [type=string_type, input_value={}, input_type=dict]
For further information visit https://errors.pydantic.dev/2.12/v/string_type

```
**To Reproduce**
```python
class WebExtractResult(BaseModel):
    supplier_description: str


async def web_search(client: AsyncFirecrawl, query: str) -> str:
    """Perform a web search using Firecrawl and return the combined text of the top results."""
    search_results: SearchData = await client.search(query=query,
                                                     sources=["web", "news"],
                                                     limit=3, 
                                                     scrape_options=ScrapeOptions(
        formats=["summary", "markdown"],
        parsers=["pdf"],
    ))

    urls = [result.url for result in search_results.web]

    extracted = await client.extract(urls=urls, prompt=query, schema=WebExtractResult.model_json_schema())
    combined_text = extracted.data
    return combined_text
```
**Expected Behavior**
A parsed response or better error handling

**Environment (please complete the following information):**
- OS: WSL2 Windows 11
- Firecrawl Version: main branch but sdk is 4.8.0
- Node.js Version: [e.g. 14.x]
- Docker Version (if applicable): Version 4.50.0 (209931)

**Docker logs:**
[...]

PR #2450, opened 2025-11-27
fix: improve pydantic parsing error handling | ENG-4070

Chadha93:
Fixes: #2403

---
## Summary by cubic
Normalize extract-status error to a string to prevent Pydantic parsing failures. Updated v1 and v2 controllers to use the string value, or for objects use message/error or JSON.stringify; otherwise omit the error.

<sup>Written for commit bd9b46979d27e4ae5fb401e060d9aa82c9ec7f3d. Summary will update automatically on new commits.</sup>

### DIFF
--- a/apps/api/src/controllers/v1/extract-status.ts
+++ b/apps/api/src/controllers/v1/extract-status.ts
@@ -119,7 +119,7 @@ export async function extractStatusController(
     error: (() => {
       if (typeof extract?.error === "string") return extract.error;
       if (extract?.error && typeof extract.error === "object") {
-        return extract.error.message || extract.error.error || JSON.stringify(extract.error);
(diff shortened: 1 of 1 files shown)


mogery:
can we attempt to stringify this error somehow? like, if type error, get message, if type string, get string, otherwise it's fine to give up

Chadha93:
> can we attempt to stringify this error somehow? like, if type error, get message, if type string, get string, otherwise it's fine to give up

Something like this?  Yeah
```error: (() => {
      if (typeof extract?.error === "string") return extract.error;
      if (extract?.error && typeof extract.error === "object") {
        return extract.error.message || extract.error.error || JSON.stringify(extract.error);
      }
      return undefined;
    })(),
```

mogery:
sure!

PR #2470, opened 2025-12-02
Filter undefined values on crawl status

abimaelmartell:
## Summary by cubic
Filters undefined job return values in the crawl status and search endpoints to prevent invalid documents from reaching clients. This stabilizes legacy document conversion and avoids intermittent API errors.

- **Bug Fixes**
  - Crawl Status: filter out null and undefined before and after mapping job return values.
  - Search: exclude null/undefined job results before converting to legacy documents.

<sup>Written for commit cd8d2d3ee51303220a741c00fae6299676472c0f. Summary will update automatically on new commits.</sup>


mogery:
this is suspicious to me. why is there an undefined document in the first place? that should not happen with the current logic if working correctly

PR #2484, opened 2025-12-03
fix(api): don't bill where stealth proxy was unsupported

amplitudesxd:
## Summary by cubic
Stop charging the stealth proxy fee when a site doesn’t support stealth mode. This fixes incorrect billing and aligns with Linear ENG-4090.

- **Bug Fixes**
  - Track unsupportedFeatures in the scrape pipeline and return it with the document.
  - Skip the stealthProxy surcharge in calculateCreditsToBeBilled when stealthProxy is unsupported, even if proxyUsed is "stealth".
  - Pass unsupportedFeatures from scrapeURL to the worker so billing has the full context.

<sup>Written for commit b61a4680c86dbd94493179392fb52855abfd9ced. Summary will update automatically on new commits.</sup>

### DIFF
--- a/apps/api/src/lib/scrape-billing.ts
+++ b/apps/api/src/lib/scrape-billing.ts
@@ -8,6 +8,7 @@ import {
 import { CostTracking } from "./cost-tracking";
 import { hasFormatOfType } from "./format-utils";
 import { TransportableError } from "./error";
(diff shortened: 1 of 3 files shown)


mogery (approved):
lgtm ty!

PR #2486, opened 2025-12-03
chore(js-sdk): pass `ignoreRobotsTxt` option through

minyoung:
Add and pass the `ignoreRobotsTxt` option from the js-sdk to the API

Related ticket: 17299

---
## Summary by cubic
Pass the ignoreRobotsTxt option from the JS SDK to the API so callers can bypass robots.txt when needed. Addresses Linear 17299 by adding the option to CrawlOptions and including it in the crawl payload.

<sup>Written for commit d3767c6d41006b6ee02947007c2dc5f5b657bc52. Summary will update automatically on new commits.</sup>


micahstairs:
Sorry, this is for internal use only.

PR #2504, opened 2025-12-05
fix: prevent empty cookie header in webhook callbacks

devhims:
### Problem

When Firecrawl sends webhook callbacks, the `http-cookie-agent` library sends an empty `cookie: ""` header even when the cookie jar is empty. This causes strict HTTP parsers like Laminas to crash, viewing the empty cookie header as malformed.

**Affected users:** Anyone using webhook callbacks with strict HTTP libraries (e.g., Expose, Laminas).

### Root Cause

The webhook delivery code (`delivery.ts`) uses `getSecureDispatcher()` which includes cookie handling via `http-cookie-agent`. Even with an empty cookie jar, this sends: cookie: ""

### Solution

Introduced a separate dispatcher (`getSecureDispatcherNoCookies`) for webhook delivery that doesn't include cookie handling, since webhooks don't need to maintain cookie state.

### Changes

- **`safeFetch.ts`**: Added `getSecureDispatcherNoCookies()` - a dispatcher without cookie agent
- **`delivery.ts`**: Updated to use the new cookie-free dispatcher

### Testing

Verified locally that:
1. `getSecureDispatcher()` sends empty cookie header (expected for scraping)
2. `getSecureDispatcherNoCookies()` does NOT send cookie header (fix confirmed)

Also tested via webhook.site - the `cookie` header no longer appears in webhook callbacks.

---
## Summary by cubic
Prevents empty Cookie headers in webhook callbacks by switching webhook delivery to a cookie-free dispatcher, avoiding crashes in strict HTTP parsers.

- **Bug Fixes**
  - Use getSecureDispatcherNoCookies() for webhook delivery to stop sending `cookie: ""`.

<sup>Written for commit 4190ce68ec4305d826271a9858068738f988d1b3. Summary will update automatically on new commits.</sup>

### DIFF
--- a/apps/api/src/scraper/scrapeURL/engines/utils/safeFetch.ts
+++ b/apps/api/src/scraper/scrapeURL/engines/utils/safeFetch.ts
@@ -18,12 +18,12 @@ export function isIPPrivate(address: string): boolean {
   return addr.range() !== "unicast";
 }
 
-function makeSecureDispatcher(skipTlsVerification: boolean) {
+function createBaseAgent(skipTlsVerification: boolean) {
   const agentOpts: undici.Agent.Options = {
     maxRedirections: 5000,
   };
 
-  const baseAgent = process.env.PROXY_SERVER
+  return process.env.PROXY_SERVER
     ? new undici.ProxyAgent({
         uri: process.env.PROXY_SERVER.includes("://")
           ? process.env.PROXY_SERVER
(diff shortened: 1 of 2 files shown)


apps/api/src/scraper/scrapeURL/index.ts
mogery, line 76: leftover changes from rewriteUrl PR?

devhims: I accidentally created the new branch from the last fix branch. Just corrected that.

PR #2546, opened 2025-12-12
fix(api): resolve extract status stuck at "processing" indefinitely

devhims:
This PR fixes a race condition where the v2 extract status endpoint (GET /v2/extract/:id) could return status: "processing" with empty data indefinitely, even after an extraction had successfully completed.

## Root Cause

Three issues were identified:

1. **Fire-and-forget status update**: In both `performExtraction` and `performExtraction_F0`, the Redis status update to "completed" was done in a `.then()` callback after `logExtract()`. The function returned before this callback executed, causing a race condition.

2. **Missing status on failure**: When `performExtraction` returned `{ success: false }` (e.g., "No search results found"), the worker set the error but never updated status to "failed", leaving it stuck at "processing" forever.

3. **No fallback for processing state**: The status endpoint only checked BullMQ/DB when Redis was null or "completed". When Redis said "processing", it returned that status without verifying against BullMQ.

## Changes

- **extraction-service.ts / extraction-service-f0.ts**: Replace fire-and-forget `.then()` pattern with proper `await` to ensure status is updated before return
- **extract-worker.ts**: Add `status: "failed"` when extraction returns `success: false`
- **v1/extract-status.ts / v2/extract-status.ts**: Check BullMQ/DB even when Redis says "processing" to handle race conditions
- Added comprehensive test suite to verify all fixes

---
## Summary by cubic
Fixes a race condition that left GET /v1|v2/extract/:id stuck on "processing" with empty data. Status updates are now reliable on both success and failure.

- **Bug Fixes**
  - Await logExtract and updateExtract to avoid fire-and-forget races.
  - Set status: "failed" in extract-worker when success: false (and on errors).
  - In v1/v2 status controllers, check BullMQ/DB even when Redis says "processing".
  - Align status fallback with DB flag (use is_successful) when BullMQ job is missing.
  - Added tests covering all scenarios.

<sup>Written for commit 2d7382249822635cd2562707d90726ec3fbb8fab. Summary will update automatically on new commits.</sup>

### DIFF
--- a/apps/api/src/controllers/v1/extract-status.ts
+++ b/apps/api/src/controllers/v1/extract-status.ts
@@ -10,7 +10,7 @@ import { logger as _logger } from "../../lib/logger";
 
 type DBExtract = {
   id: string;
-  success: boolean;
+  is_successful: boolean;
   options: any;
   created_at: any;
   error: string | null;
@@ -47,7 +47,7 @@ async function getExtractJob(
   const job: ExtractPseudoJob<any> = {
     id,
     getState: dbExtract
-      ? () => (dbExtract.success ? "completed" : "failed")
+      ? () => (dbExtract.is_successful ? "completed" : "failed")
       : () => bullJob!.getState(),
     returnvalue: data,
     data: {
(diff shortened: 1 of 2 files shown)


devhims:
This PR should also address the extract playground issue in dashboard where it's stuck at "Parallelizing requests..."

amplitudesxd:
superseded by #2548 -- tried to fix actual root cause!

PR #2600, opened 2025-12-23
fix(api): handle browser-rendered XML in sitemap parsing

devhims:
### Problem

When the map endpoint fetches a `sitemap.xml` file using browser-based engines (fire-engine with Chrome), the XML gets rendered as an HTML page with the content wrapped in a viewer div:

```html
<html>
  <body>
    <div id="webkit-xml-viewer-source-xml">
      <urlset>...actual sitemap content...</urlset>
    </div>
  </body>
</html>
````

The XML parsers expect raw XML but instead receive this HTML-wrapped content, causing parsing to fail silently.
This results in `sitemap: "only"` returning **0 results intermittently**, depending on which engine fetches the content.

### Solution

Added an `extractXmlFromHtmlWrapper()` helper function that detects and extracts raw XML from browser-rendered HTML before parsing:

* Detects WebKit XML viewer wrapper (`webkit-xml-viewer-source-xml`)
* Detects browser-mangled XML declaration (`<!--?xml`)
* Falls back to regex extraction of `<urlset>` or `<sitemapindex>`
* Passes through raw XML unchanged

### Testing

* Added unit tests covering all extraction scenarios
* Verified the fix with real browser-rendered content from
  `prairie-dogs-e2e.zapier-staging.com/sitemap.xml`

### Files Changed

* `apps/api/src/scraper/WebScraper/sitemap.ts`

  * Added helper function and applied fix
* `apps/api/src/scraper/crawler/sitemap.ts`

  * Applied fix
* `apps/api/src/scraper/WebScraper/__tests__/sitemap.test.ts`

  * Added unit tests

---
## Summary by cubic
Fix sitemap parsing when browser engines render XML as HTML. Extracts raw XML before parsing so sitemap-only runs return the correct links instead of zero results.

- **Bug Fixes**
  - Added extractXmlFromHtmlWrapper to strip WebKit XML viewer HTML and handle mangled XML declarations.
  - Fallback extraction of urlset or sitemapindex when wrapper detection fails; raw XML passes through unchanged.
  - Applied in WebScraper and crawler code paths; added unit tests covering all scenarios.

<sup>Written for commit c119a738070da9b92504011608f5f20ad6e0e08e. Summary will update automatically on new commits.</sup>

### DIFF
--- a/apps/api/src/scraper/WebScraper/__tests__/sitemap.test.ts
+++ b/apps/api/src/scraper/WebScraper/__tests__/sitemap.test.ts
@@ -1,43 +1,4 @@
-// Test the extractXmlFromHtmlWrapper function in isolation
-// We inline the function here to avoid importing the full sitemap module
-// which has complex dependencies that Jest can't handle
-
-/**
- * Extracts raw XML from HTML-wrapped content that browsers generate when rendering XML files.
- */
-function extractXmlFromHtmlWrapper(content: string): string {
-  // Only process if content looks like HTML-wrapped XML
-  if (
-    content.includes("webkit-xml-viewer-source-xml") ||
-    (content.startsWith("<!--?xml") && content.includes("<html"))
-  ) {
-    // Try to extract from WebKit XML viewer div
-    const webkitMatch = content.match(
-      /<div[^>]*id=["']webkit-xml-viewer-source-xml["'][^>]*>([\s\S]*?)<\/div>/i,
-    );
-    if (webkitMatch && webkitMatch[1]) {
-      const extracted = webkitMatch[1].trim();
-      if (
-        extracted.includes("<urlset") ||
-        extracted.includes("<sitemapindex")
-      ) {
-        return extracted;
-      }
-    }
-
-    // Fallback: extract urlset or sitemapindex directly via regex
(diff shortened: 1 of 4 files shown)


micahstairs:
@devhims why is there a duplicate copy of the function in the testing file? Doesn't make sense to branch the implementation and test the branch that's not actually being run in production.

micahstairs:
I'm also not convinced this is the most elegant solution here. We probably should just use a different engine for sitemap files.

devhims:
> @devhims why is there a duplicate copy of the function in the testing file? Doesn't make sense to branch the implementation and test the branch that's not actually being run in production.

I duplicated it because importing from the actual sitemap.ts pulls in the entire module with complex dependencies (Redis, scrapeURL, native modules, etc.) that Jest couldn't handle - it failed with ESM/CommonJS issues.

I can either:
1. Move `extractXmlFromHtmlWrapper()` to a separate utility file with no dependencies, then import it in both the sitemap files and the test
2. Remove the test file if the fix approach changes

micahstairs:
Definitely just move the function into a separate utility file. Duplicating the function is way too error-prone (and also completely unecessary)

devhims:
@micahstairs Created `xml-utils.ts `with `extractXmlFromHtmlWrapper()` function

Issue #2802, opened 2026-02-05
[Bug] scrape headers.user-agent value not used by playwright


Zambonilli:
**Describe the Bug**
When calling the scrape endpoint with the headers object that includes a `user-agent` field, the value of the `user-agent` is not used by playwright as the request's user-agent.

**To Reproduce**
Steps to reproduce the issue:
1. Deploy self-hosted version of v2.8.0
2. create a webhook.site test endpoint or similar tool for logging HTTP requests
3. call POST v2/scrape endpoint with headers user-agent set
4. Observe the user-agent used by playwright for url request is not the value specified

```bash
curl --request POST \
  --url https://<self-hosted-api-endpoint>/v2/scrape \
  --header 'content-type: application/json' \
  --data '{
  "url": "https://webhook.site/<your-webhook-site-url>",
  "storeInCache": true,
  "maxAge": 172800000,
  "timeout": 60000,
  "formats": [
    "markdown",
    "links"
  ],
  "headers": {
    "user-agent": "<your-desired-user-agent>"
  }
}'
```

**Expected Behavior**
Requests sent to the url use the user-agent specified in the scrape API request headers.user-agent field.

**Environment (please complete the following information):**
- OS: Linux
- Deployment Type: self-hosted
- Firecrawl Version: v2.0.8
- Node.js Version: 22

**Additional Context**
I believe the problem is that the playwright service is setting the context's user agent to a derived UserAgent package user agent so later when the headers are being set playwright ignores the `user-agent` key when it's passed through.

Line 255 creates the playwright context user agent from UserAgent package: https://github.com/firecrawl/firecrawl/blob/main/apps/playwright-service-ts/api.ts#L255

Line 259 sets the playwright extra headers which if user-agent is included is ignored because it's already set on the context to the UserAgent package value: https://github.com/firecrawl/firecrawl/blob/main/apps/playwright-service-ts/api.ts#L259

PR #2806, opened 2026-02-06
fix(playwright-service): use custom user-agent from headers when provided

kaigritun:
## Summary
Fixes #2802

When calling the scrape endpoint with a custom `user-agent` header, the value was being ignored because Playwright sets the user-agent on the browser context, not via `setExtraHTTPHeaders`.

## Changes
- Modified `createContext` to accept an optional `customUserAgent` parameter
- Extract user-agent from headers before creating context
- Pass custom user-agent to `createContext` if provided
- Remove user-agent from headers before calling `setExtraHTTPHeaders` to avoid redundancy

## Testing
To test this fix:
1. Deploy the updated playwright-service
2. Create a test endpoint (e.g., webhook.site) to log HTTP requests
3. Call the scrape endpoint with a custom user-agent:
```bash
curl --request POST \
  --url http://localhost:3000/v2/scrape \
  --header 'content-type: application/json' \
  --data '{
  "url": "https://webhook.site/<your-webhook-id>",
  "headers": {
    "user-agent": "MyCustomUserAgent/1.0"
  }
}'
```
4. Verify that the user-agent logged at webhook.site is `MyCustomUserAgent/1.0`

---
## Summary by cubic
Honors a custom User-Agent sent to the /scrape endpoint by setting it on the Playwright browser context. Sites now see the provided UA instead of the default.

- **Bug Fixes**
  - Set user-agent on the browser context, not via setExtraHTTPHeaders.
  - Use case-insensitive header lookup to extract the custom UA before creating the context.
  - Pass the custom UA to createContext; fall back to a generated UA if not provided.
  - Strip the UA from headers before setExtraHTTPHeaders to avoid duplication.

<sup>Written for commit 89f4ef95a4a68a7da6d4c8c79b938d629985a310. Summary will update on new commits.</sup>

### DIFF
--- a/apps/playwright-service-ts/api.ts
+++ b/apps/playwright-service-ts/api.ts
@@ -252,17 +252,27 @@ app.post('/scrape', async (req: Request, res: Response) => {
   let page: Page | null = null;
 
   try {
-    // Extract user-agent from headers if provided, as it needs to be set on the context
-    const customUserAgent = headers?.['user-agent'] || headers?.['User-Agent'];
+    // Extract user-agent from headers if provided (case-insensitive), as it needs to be set on the context
+    let customUserAgent: string | undefined;
+    let headersWithoutUA: Record<string, string> | undefined;
+    
+    if (headers) {
+      const headerEntries = Object.entries(headers);
+      const userAgentEntry = headerEntries.find(([key]) => key.toLowerCase() === 'user-agent');
+      customUserAgent = userAgentEntry?.[1];
+      
+      // Filter out user-agent header (case-insensitive) since it's set on the context
+      const filteredEntries = headerEntries.filter(([key]) => key.toLowerCase() !== 'user-agent');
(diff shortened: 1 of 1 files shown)


kaigritun:
Fixed the case-insensitivity issue. Now iterates through headers and checks with `key.toLowerCase() === 'user-agent'` instead of checking specific casings. This properly handles all valid casing variants per RFC 7230.

kaigritun:
Good catch! Fixed the case-insensitive header lookup - now properly handles any casing of 'User-Agent' (e.g., 'USER-AGENT', 'user-Agent', etc.) using `.toLowerCase()` comparison.

MichaelDeBoey:
I'm not making any claims about the quality of their work, but I wanted to let you know that @kaigritun is a fully-autonomous non-human actor
https://socket.dev/blog/ai-agent-lands-prs-in-major-oss-projects-targets-maintainers-via-cold-outreach

PR #2948, opened 2026-02-28
fix(rust-sdk): handle metadata fields returned as arrays

### LINKED ISSUE #1304: [Bug] Rust SDK check_crawl_status errors if there are arrays in the metadata
**Describe the Bug**
I suspect that when calling check_crawl_status, the Rust SDK expects all of the metadata fields to be strings, even though sometimes they can be Vec<Strings>.

For example
```
"viewport": ["width=device-width, initial-scale=1", "width=device-width,initial-scale=1.0,maximum-scale=1.0,user-scalable=no,viewport-fit=cover"]
```

or

```
"twitter:image": ["http://localhost:3000/#{FALLBACK_IMAGE}", "https://images.contentstack.io/v3/assets/bltdc2476c7b6b194dd/blt46acaa0d3fe3fd06/66db08a7b89507bdf18d1c3d/Sports_Impact_Report.jpg"]
```

If you have a crawl with vector metadata like this, then the Rust SDK will throw the error
```
Error: Failed to parse response: invalid type: sequence, expected a string
```

However, if your crawl has only string fields in the metadata, it works as expected.

**To Reproduce**
1. Get a crawl with vectors in the metadata
2. Call check_crawl_status with that crawl id using the rust sdk

**Expected Behavior**
The sdk should throw an error

- OS: macOS Sequoia 15.3
- Firecrawl Rust SDK Version: 1.0.0
- Node.js Version: N/A

madmikeross:
The API returns arrays for DocumentMetadata fields (e.g. `"robots": ["index", "follow"]`) when a page has duplicate meta tags, but the Rust SDK types these as `Option<String>`, causing serde to fail with `"invalid type: sequence, expected a string"`.

This adds a custom `deserialize_string_or_array` serde helper that accepts both `String` and `[String]`, joining arrays with `", "`. All `Option<String>` metadata fields in both v1 and v2 DocumentMetadata structs are annotated with it. The public type signature stays `Option<String>` so this is a non-breaking change.

New unit tests cover string values, array values, null, missing fields, empty arrays, single-element arrays, serialization roundtrip, and full document parsing.

Fixes #1304

### DIFF
--- a/apps/rust-sdk/src/lib.rs
+++ b/apps/rust-sdk/src/lib.rs
@@ -42,11 +42,11 @@ pub mod crawl;
 pub mod document;
 pub mod error;
 pub mod extract;
-pub(crate) mod serde_helpers;
 pub mod llmstxt;
 pub mod map;
 pub mod scrape;
 pub mod search;
+pub(crate) mod serde_helpers;
 
 // v2 module (new API)
 pub mod v2;
--- a/apps/rust-sdk/src/serde_helpers.rs
+++ b/apps/rust-sdk/src/serde_helpers.rs
@@ -49,8 +49,7 @@ mod tests {
 
     #[test]
     fn test_array_value() {
-        let h: Helper =
-            serde_json::from_value(json!({"field": ["index", "follow"]})).unwrap();
+        let h: Helper = serde_json::from_value(json!({"field": ["index", "follow"]})).unwrap();
         assert_eq!(h.field, Some("index, follow".to_string()));
     }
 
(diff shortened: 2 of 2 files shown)


madmikeross:
@mogery or @abimaelmartell could you please approve workflows when you have a chance? Or offer guidance on preferred approach?

madmikeross:
@mogery @abimaelmartell thanks for kicking off the workflows. Build and test passed. It looks like the other failures are pre-existing or unrelated (usual report publish failure due to fork perms).

abimaelmartell:
hey @madmikeross could you reuse `deserialize_string_or_array`? Seems to be duplicated.

Thank you for your contribution!

apps/rust-sdk/src/document.rs
```diff
@@ -1,6 +1,33 @@
...
+            if strings.is_empty() {
+                Ok(None)
+            } else {
+                Ok(Some(strings.join(", ")))
+            }
+        }
+        Some(other) => Ok(Some(other.to_string())),
+    }
+}
+
```
abimaelmartell: Could you reuse this? Maybe extract into a shared module (eg serde_helpers.rs), and importing it in both places to keep things DRY.

Same for the test suites which are nearly identical. Code looks good, but this could help reduce maintenance burden.

Thanks!

madmikeross: Great suggestion. Fixed here [9f90212](https://github.com/firecrawl/firecrawl/pull/2948/commits/9f90212210e7ae99d8ae38c10e88f10cf7401b9e)

abimaelmartell:
Looks good, thank you so much for this contribution, will merge this in a moment.

Best,
Abimael Martell

PR #2980, opened 2026-03-05
fix: use user-agent from headers if provided in scrape request

fuleinist:
When user-agent is passed in headers to the scrape endpoint, it was being ignored because Playwright sets the context's user-agent first (from UserAgent package), then extra HTTP headers are applied. Playwright ignores user-agent in extra headers if it's already set on the context.

This fix checks if headers['user-agent'] is provided and uses that for the context's userAgent instead of the derived UserAgent package value.

Fixes #2802

---
## Summary by cubic
Use the user-agent from scrape request headers when provided, so Playwright uses the requested UA instead of the generated default. Fixes cases where the header user-agent was being ignored.

- **Bug Fixes**
  - Pass headers to createContext and set the context userAgent from a case-insensitive "user-agent" header when available.
  - Prevent Playwright from overriding the header UA by setting it at context creation.

<sup>Written for commit a0df5d176fa9b548722cb35691322b02d7e882b9. Summary will update on new commits.</sup>

### DIFF
--- a/apps/playwright-service-ts/api.ts
+++ b/apps/playwright-service-ts/api.ts
@@ -99,8 +99,8 @@ const initializeBrowser = async () => {
   });
 };
 
-const createContext = async (skipTlsVerification: boolean = false) => {
-  const userAgent = new UserAgent().toString();
+const createContext = async (skipTlsVerification: boolean = false, headers?: Record<string, string>) => {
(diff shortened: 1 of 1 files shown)


GaneshGone (approved):

fuleinist:
Stale PR (no activity for 4+ weeks). Feel free to reopen if still relevant.

PR #3020, opened 2026-03-08
fix: use urlWithoutWww in site: search query for www URLs (Issue #2580)

### LINKED ISSUE #2580: [Bug] unable to map website with www
**Describe the Bug**
import FirecrawlApp from '@mendable/firecrawl-js';

```js
const fw = new FirecrawlApp({ apiKey: 'fc-API_KEY' });

fw.map('https://www.deztrox.ai/', {
  maxDepth: 3,
  limit: 50,
  includeSubdomains: false,
})
  .then((data) => {
    console.log('Crawling completed:', data);
  })
  .catch((error) => {
    console.error('Error during crawling:', error);
  });
```

Crawling completed: { links: [] }

**Expected Behavior**
Should give the root path

**Screenshots**

Image

gambletan:
When mapping a website with www prefix (e.g., https://www.example.com), the search query was incorrectly built as 'site:https://www.example.com' instead of 'site:example.com'. This caused the search engine to return no results because site: searches don't work with full URLs with protocols.

This fix ensures the URL without www is used in the site: search query, which properly indexes the domain regardless of whether the user includes www in their original URL.

**Fixes #2580**

---
## Summary by cubic
Fixes site: search queries for www-prefixed URLs and makes the scraper respect a custom User-Agent header. Improves mapping accuracy and parity with client headers.

- **Bug Fixes**
  - Build `site:` queries with `urlWithoutWww` (e.g., `site:example.com`) instead of full URLs to return results. Fixes #2580.
  - Respect custom User-Agent by setting it on the Playwright context (case-insensitive extraction) and removing it from extra headers; `createContext` now accepts an optional user agent. Fixes #2802.

<sup>Written for commit 3911413d2ffd504bd9afef6e591de8984a8324e3. Summary will update on new commits.</sup>

### DIFF
--- a/apps/api/src/controllers/v1/map.ts
+++ b/apps/api/src/controllers/v1/map.ts
@@ -187,7 +187,7 @@ export async function getMapResults({
         ? `${search} ${urlWithoutWww}`
         : search
           ? `${search} site:${urlWithoutWww}`
-          : `site:${url}`;
+          : `site:${urlWithoutWww}`;
 
     const resultsPerPage = 100;
     const maxPages = Math.ceil(
--- a/apps/api/src/lib/map-utils.ts
+++ b/apps/api/src/lib/map-utils.ts
@@ -198,7 +198,7 @@ export async function getMapResults({
         ? `${search} ${urlWithoutWww}`
         : search
           ? `${search} site:${urlWithoutWww}`
-          : `site:${url}`;
+          : `site:${urlWithoutWww}`;
 
     const resultsPerPage = 100;
     const maxPages = Math.ceil(


gambletan:
### Contributor License Agreement



By submitting this pull request, I confirm that I have read and fully agree to the [Contributor License Agreement (CLA)](https://github.com/firecrawl/firecrawl/blob/main/CONTRIBUTOR_LICENSE_AGREEMENT), and I am providing my contributions under its terms.

gambletan:
This has been approved — can it be merged?

PR #3024, opened 2026-03-08
fix: resolve race condition causing Job not found + add skipData param (fixes #2662, #2599)

### LINKED ISSUE #2599: [Feat] Provide a way to omit data in 'GET batch scrape status' API response
**Problem Description**

When I'm using async batch scrape API(`POST /batch/scrape`) and check its status with GET api (`GET /batch/scrape/:id`), GET api returns whole "data" even when status is not "completed" and its a burden because I only use data result when status is "completed".

**Proposed Feature**

Provide a way to omit "data" field in response

**Alternatives Considered**

- Ignore response no matter what size of http response
  ➡️ receiving huge (like 5MB) response every time before "completed" batch scrape causes too much overhead, so want to avoid this.
- Use some custom parser and stop parse when met `"status": ..(other than 'completed')`
  ➡️ this is not reliable("status" can be placed at the end of JSON), and don't think it's so reasonable way.

**Implementation Suggestions**

- Introduce optional boolean `skipData` query parameter. (default: false)

**Use Case**
User can selectively skip huge "data" and can GET batch scrape's status only.

**Additional Context**
N/A

### LINKED ISSUE #2662: [Bug] High probability "Job not found" when immediately querying crawl status with valid job_id
### Bug Description
When calling the crawl status endpoint **immediately** with a **valid job_id** (returned by the `/v1/crawl` create API), there is a **high probability** of receiving a `{"success":false,"error":"Job not found"}` response.

However, if I wait for a short period of time (**1-3 seconds**) and call the same `/v1/crawl/{job_id}` status endpoint with the **exact same job_id**, the API will successfully return the correct crawl job status data (e.g. `pending`, `running`, `completed`) with no errors.

The job_id is always valid (directly extracted from the successful 200 response of the `/v1/crawl` API), no typo, no invalid parameter, no environment mismatch. This issue occurs stably and frequently.

### Steps to Reproduce (100% reproducible with high probability)
1. Call the `POST /v1/crawl` API with valid crawling parameters (target URL, configs etc.)
2. Receive a **successful response** from the API, extract the valid `job_id` from the response body
3. **Immediately invoke** the `GET /v1/crawl/{job_id}` status API right after getting the job_id (no delay)
4. High chance to get error response: `{"success":false,"error":"Job not found"}`
5. Wait **1-3 seconds**, call the same `GET /v1/crawl/{job_id}` API with the same job_id
6. The API returns the correct crawl status data successfully

### Expected Behavior
Once the `/v1/crawl` API returns a valid `job_id` (successfully), the job should be persisted in the backend immediately. 
Calling the status endpoint with this job_id **should return a valid job status** (e.g. `queued`/`pending`/`running`) at any time, **NOT** a `Job not found` error.

A returned job_id implies a successful job creation, which should guarantee the job is queryable immediately.

### Actual Behavior
- High probability of `{"success":false,"error":"Job not found"}` when querying status **immediately**
- Successful status response when querying the same job_id after a short delay (1-3s)

### Additional Context
- No network issues / invalid API keys / wrong base URL
- Tested with multiple different job_ids, the issue is consistent
- This forces all clients to implement a **retry mechanism** for the status endpoint, which is an unnecessary integration cost and violates the API contract
- Curl test example for the failed case:
  ```bash
  curl http://localhost:80/v1/crawl/[valid_job_id]

gambletan:
When calling the crawl status endpoint immediately after creating a crawl, there was a high probability of receiving a 'Job not found' error due to a race condition between the crawl metadata being saved to Redis (fast) and the group being created in the database (slower).

This fix checks Redis first for the crawl metadata. If the crawl exists and the team_id matches, it returns an initial 'scraping' status even if the group hasn't been created in the database yet.

Additionally, this PR adds a new `skipData` query parameter to the crawl/batch-scrape status endpoints. When `skipData=true` and the job status is not 'completed', the `data` field is omitted from the response. This reduces overhead for users who only need status info before the job completes (fixes #2599).

Fixes #2662
Fixes #2599

### DIFF
--- a/apps/api/src/__tests__/snips/v2/crawl.test.ts
+++ b/apps/api/src/__tests__/snips/v2/crawl.test.ts
@@ -396,8 +396,8 @@ describe("Crawl tests", () => {
 
       expect(res.success).toBe(true);
       if (res.success) {
-        expect(res.data.length).toBeGreaterThan(0);
-        for (const page of res.data) {
+        expect(res.data?.length).toBeGreaterThan(0);
+        for (const page of res.data || []) {
           const url = new URL(page.metadata.url ?? page.metadata.sourceURL!);
           const hostname = url.hostname;
 
@@ -453,7 +453,7 @@ describe("Crawl tests", () => {
           expect(res.success).toBe(true);
           if (res.success) {
             // Verify that explicit options were respected
-            for (const page of res.data) {
+            for (const page of res.data || []) {
               const url = new URL(
                 page.metadata.url ?? page.metadata.sourceURL!,
               );
--- a/apps/api/src/controllers/v1/types.ts
+++ b/apps/api/src/controllers/v1/types.ts
@@ -1179,7 +1179,7 @@ export type CrawlStatusResponse =
       creditsUsed: number;
       expiresAt: string;
       next?: string;
-      data: Document[];
+      data?: Document[];
     };
 
 export type OngoingCrawlsResponse =
--- a/apps/api/src/controllers/v2/crawl-status.ts
+++ b/apps/api/src/controllers/v2/crawl-status.ts
@@ -373,7 +373,7 @@ export async function crawlStatusController(
     creditsUsed: outputBulkA.creditsUsed ?? 0,
     expiresAt: (await getCrawlExpiry(req.params.jobId)).toISOString(),
     next: outputBulkB.next,
-    data: outputBulkB.data,
+    ...(!skipData || outputBulkA.status === "completed" ? { data: outputBulkB.data } : {}),
     ...(warning && { warning }),
   });
 }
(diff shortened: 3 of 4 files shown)


gambletan:
Thanks for the review! 

identified by cubic - The concern is valid. The Redis-first approach could mask real DB status. A better fix would be to:

1. Check both Redis and DB status
2. Return DB status if available, otherwise fallback to Redis
3. Or add a timestamp to the Redis cache to invalidate it after the job completes

Would you prefer I update the PR with a more robust solution that checks DB status as the primary source?

gambletan:
Thanks for the feedback! 

identified by cubic - You're right that the Redis-first approach has this limitation. A more robust solution would be:

1. Check both Redis AND DB status
2. Return DB status if available (primary), fallback to Redis (secondary)
3. Only return Redis status if DB record doesn't exist yet

Would you prefer I update the PR with this improved approach? Let me know and I'll revise the code.

PR #3030, opened 2026-03-08
fix: use custom user-agent from headers in Playwright context

gambletan:
When calling the scrape endpoint with headers.user-agent, the value was not being used because Playwright first sets the user-agent at context level, then ignores the user-agent key when setExtraHTTPHeaders is called.

This fix:
1. Passes headers to createContext so custom user-agent can be used at context creation time
2. Removes user-agent from headers before calling setExtraHTTPHeaders since it's already set at context level

Fixes #2802

---
## Summary by cubic
Ensure custom User-Agent from `headers.user-agent` is applied in Playwright by setting it during context creation, not via extra headers. This makes the `/scrape` endpoint honor the caller’s UA. Fixes #2802.

- **Bug Fixes**
  - Pass `headers` into `createContext` and use `headers['user-agent']` if provided; fallback to a generated UA.
  - Remove `user-agent` before `page.setExtraHTTPHeaders`; only set extra headers when others are present.

<sup>Written for commit 327290dec7f6ac5d80fa455e7199b670f3e82b42. Summary will update on new commits.</sup>

### DIFF
--- a/apps/playwright-service-ts/api.ts
+++ b/apps/playwright-service-ts/api.ts
@@ -99,8 +99,9 @@ const initializeBrowser = async () => {
   });
 };
 
-const createContext = async (skipTlsVerification: boolean = false) => {
-  const userAgent = new UserAgent().toString();
+const createContext = async (skipTlsVerification: boolean = false, customHeaders?: Record<string, string>) => {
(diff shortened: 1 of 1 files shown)


gambletan:
Superseded by #3054.

PR #3031, opened 2026-03-08
fix: use custom user-agent from headers in Playwright context (fix #2802)

gambletan:
When calling the scrape endpoint with headers.user-agent, the value was not being used because Playwright first sets the user-agent at context level, then ignores the user-agent key when setExtraHTTPHeaders is called.

This fix:
1. Passes headers to createContext so custom user-agent can be used at context creation time
2. Removes user-agent from headers before calling setExtraHTTPHeaders since it's already set at context level

Fixes #2802

---
## Summary by cubic
Make the scrape endpoint honor a custom user-agent from `headers.user-agent` by setting it on the Playwright context at creation, not via extra headers. Fixes #2802.

- **Bug Fixes**
  - `createContext` accepts optional `customHeaders` and sets `user-agent` from `headers.user-agent`, falling back to a generated value.
  - `/scrape` passes headers to `createContext` and excludes `user-agent` from `setExtraHTTPHeaders`.

<sup>Written for commit 1c29e0a161c3d1f9c7f9a2f606133a24bf0d5516. Summary will update on new commits.</sup>

### DIFF
--- a/apps/playwright-service-ts/api.ts
+++ b/apps/playwright-service-ts/api.ts
@@ -99,8 +99,9 @@ const initializeBrowser = async () => {
   });
 };
 
-const createContext = async (skipTlsVerification: boolean = false) => {
-  const userAgent = new UserAgent().toString();
+const createContext = async (skipTlsVerification: boolean = false, customHeaders?: Record<string, string>) => {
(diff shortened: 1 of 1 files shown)


gambletan:
Superseded by #3054.

Issue #2891, opened 2026-02-22
[Bug] Get agent status endpoint never returns creditsUsed attribute until status=complete


jschellerat:
**Describe the Bug**
When polling the Get Agent Status endpoint (`GET /v1/agent/{agentId}`) while an agent job is in progress, the `creditsUsed` attribute is never included in the response. It only appears once the agent job reaches `status=complete`. This makes it impossible to track credit consumption in real time while a job is running.

**To Reproduce**
1. Start an agent crawl job via `POST /v1/agent`
2. Poll the agent status endpoint `GET /v1/agent/{agentId}` while the job is still running (`status` is not `complete`)
3. Observe that the `creditsUsed` field is missing from the response body
4. Wait for the job to complete (`status=complete`)
5. Poll the status endpoint again — `creditsUsed` now appears in the response

**Expected Behavior**
The `creditsUsed` attribute should be returned in the Get Agent Status response at all stages of the job lifecycle (e.g. `processing`, `running`, `complete`), reflecting the credits consumed so far. This would allow consumers to monitor credit usage in real time rather than only after completion.

**Screenshots**
N/A

**Environment (please complete the following information):**
- OS: macOS
- Deployment Type: Cloud (firecrawl.dev)
- Firecrawl Version: Latest (API)
- Node.js Version: N/A (using REST API directly)

**Logs**
N/A

**Additional Context**
The API documentation for Get Agent Status lists `creditsUsed` as a response field, but it is only populated when the job is complete. It would be helpful to either update the docs to clarify this limitation, or (preferably) return the running credit total at every status check.

Godkunn:
Hi team, I'd love to take a look at this. It seems like the creditsUsed variable just needs to be piped into the response payload during the running state in the agent controller. Can I get assigned to this?

PR #3035, opened 2026-03-08
fix: return real-time creditsUsed in agent status endpoint (Issue #2891)

gambletan:
When polling the Get Agent Status endpoint while an agent job is in progress, the creditsUsed attribute was never included in the response - it only appeared once the agent job reached status=complete.

This fix:
1. Uses the same RPC call (credits_billed_by_crawl_id_2) that crawl status uses to get real-time credits during processing
2. Falls back to agent.credits_cost when the job is complete
3. Adds proper error handling with logging

Now users can track credit consumption in real time while a job is running, not just after completion.

---
## Summary by cubic
Return real-time `creditsUsed` in the agent status endpoint so users can track credit consumption while a job is running. Addresses #2891.

- **Bug Fixes**
  - Agent status: fetch live credits via Supabase RPC `credits_billed_by_crawl_id_2`, fall back to `agent.credits_cost`, and log errors.
  - Playwright service and map search: honor request `User-Agent` by setting it on the browser context and removing it from extra headers; build `site:` queries with `urlWithoutWww` for `www` URLs.
  - `firecrawl` JS SDK: add `webhook` option to `startExtract` to enable async callbacks.

<sup>Written for commit 2ddf10986936ea96b9db702c9b331cb151d9b8d3. Summary will update on new commits.</sup>

### DIFF
--- a/apps/js-sdk/firecrawl/src/v2/methods/extract.ts
+++ b/apps/js-sdk/firecrawl/src/v2/methods/extract.ts
@@ -1,4 +1,4 @@
-import { type ExtractResponse, type ScrapeOptions, type AgentOptions } from "../types";
+import { type ExtractResponse, type ScrapeOptions, type AgentOptions, type WebhookConfig } from "../types";
 import { HttpClient } from "../utils/httpClient";
 import { ensureValidScrapeOptions } from "../utils/validation";
 import { normalizeAxiosError, throwForBadResponse } from "../utils/errorHandler";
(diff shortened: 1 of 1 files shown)


gambletan:
Thanks for the review! 

identified by cubic - 

For issue #1 (map-utils.ts): The URL transformation is intentional for the site: query - we want to convert www.example.com to example.com for the search. This is the expected behavior.

For issue #2 (agent-status.ts): Good point! I'll add error handling for the RPC response. Let me update the code to check for creditsRpc.error as well.

gambletan:
Superseded by #3051.

Issue #2582, opened 2025-12-19
[Bug] The `startExtract` method in the Node.js SDK does not accept webhook options


andreisocaciu:
**Describe the Bug**
The `startExtract` method in the Node.js SDK does not accept webhook options in its TypeScript type definitions, even though the Firecrawl API accepts webhook configuration for extract jobs. This prevents TypeScript users from properly configuring webhooks for async extraction jobs.

**To Reproduce**
Steps to reproduce the issue:
1. Install `@mendable/firecrawl-js` version 4.9.3
2. Create a webhook configuration object:
```typescript
const webhookOptions = {
  url: 'https://example.com/webhook',
  metadata: { jobId: '123', workspaceId: '456', userId: '789' },
  events: ['started', 'completed', 'failed'] as const
};
```
3. Attempt to pass webhook options to `startExtract`:
```typescript
const firecrawl = new Firecrawl();
const result = await firecrawl.startExtract({
  urls: ['https://example.com'],
  schema: mySchema,
  prompt: 'Extract data',
  webhook: webhookOptions, // ❌ TypeScript error: Property 'webhook' does not exist
});
```
4. Observe the TypeScript error: `Property 'webhook' does not exist on type '{ urls?: string[]; prompt?: string; schema?: ... }'`

**Expected Behavior**
The `startExtract` method should accept a `webhook` parameter of type `string | WebhookConfig | null`, consistent with:
- The `crawl` method which already supports `webhook?: string | WebhookConfig | null` (as seen in `CrawlParams` interface)
- The `batchScrapeUrls` method which also accepts webhook options
- The Firecrawl API itself, which accepts and processes webhook configurations for extract jobs

**Screenshots**
TypeScript error output:
```
error TS2345: Argument of type '{ urls: string[]; schema: ...; prompt: string; webhook: { url: string; metadata: { ... }; events: readonly ["started", "completed", "failed"]; }; }' is not assignable to parameter of type '{ urls?: string[]; prompt?: string; schema?: Record<string, unknown> | ZodTypeAny; systemPrompt?: string; allowExternalLinks?: boolean; enableWebSearch?: boolean; showSources?: boolean; scrapeOptions?: ScrapeOptions; ignoreInvalidURLs?: boolean; integration?: string; agent?: AgentOptions; }'.
  Property 'webhook' does not exist on type '{ urls?: string[]; prompt?: string; schema?: Record<string, unknown> | ZodTypeAny; ... }'.
```

[...]

PR #3039, opened 2026-03-09
fix(sdk): add webhook parameter to startExtract method

gambletan:
The startExtract method in the Node.js SDK did not accept webhook options in its TypeScript type definitions, even though the Firecrawl API accepts webhook configuration for extract jobs. This prevented TypeScript users from properly configuring webhooks for async extraction jobs.

This fix:
- Adds webhook parameter to prepareExtractPayload function
- Imports WebhookConfig type from types
- Adds webhook to the request body when provided

The fix is consistent with how other methods (crawl, batchScrape) handle webhook configuration.

Fixes #2582

---
## Summary by cubic
Add support for a `webhook` option in `startExtract` in the Node.js SDK to match the API. TypeScript users can now configure webhooks for async extract jobs.

- **Bug Fixes**
  - Added `webhook?: string | WebhookConfig | null` to `prepareExtractPayload` and included it in the request body when provided.
  - Imported `WebhookConfig` and aligned behavior with `crawl` and `batchScrape`.

<sup>Written for commit 12f79f5f8a867b966cf63f9b28cc31d428c912eb. Summary will update on new commits.</sup>

### DIFF
--- a/apps/js-sdk/firecrawl/src/v2/methods/extract.ts
+++ b/apps/js-sdk/firecrawl/src/v2/methods/extract.ts
@@ -1,4 +1,4 @@
-import { type ExtractResponse, type ScrapeOptions, type AgentOptions } from "../types";
+import { type ExtractResponse, type ScrapeOptions, type AgentOptions, type WebhookConfig } from "../types";
 import { HttpClient } from "../utils/httpClient";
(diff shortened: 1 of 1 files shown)


gambletan:
Superseded by #3053.

PR #3040, opened 2026-03-09
fix: treat empty page content as success instead of engine failure

### LINKED ISSUE #2316: [Bug]  Incorrect 500 Error When Scraping Empty Webpage Content
**Describe the Bug**
When scraping a webpage that contains no visible or extractable content, the Firecrawl API returns a 500 status code with the following response:

```
{
  "success": false,
  "code": "SCRAPE_ALL_ENGINES_FAILED",
  "error": "All scraping engines failed! -- Double check the URL to make sure it's not broken. If the issue persists, contact us at help@firecrawl.com."
}
```
This seems misleading since the request is valid and the target page is accessible — it just has empty content. A 500 suggests a server-side failure, but the actual situation is that there’s simply nothing to scrape.

**To Reproduce**
Steps to reproduce the issue:

1. Use the /scrape endpoint with a valid but empty webpage URL (e.g. a blank HTML page).
2. Example request: POST /scrape
 ```
{
  "url": "https://example.com/empty-page"
}
```
3. Observe that the response is a 500 error with code SCRAPE_ALL_ENGINES_FAILED.

**Expected Behavior**

Instead of returning a 500, the API should:
* Return a successful response with empty content (e.g. { "content": "" }), or
* Return a 204 (No Content) or 4xx-level response that accurately indicates there’s nothing to scrape.
This helps distinguish between true scraping engine failures and pages that are empty.

**Screenshots**
N/A — see JSON response above.

- OS: macOS Tahoe 26.0.1
- Deployment Type: Cloud (firecrawl.dev)
- Firecrawl Version:Not applicable (using hosted API)
- Node.js Version: Not applicable (tested via Postman)

**Logs**
If needed, I can provide the request/response logs showing the 500 and the URL used.

**Additional Context**
This behavior causes false positives for “engine failure” conditions in client applications that expect Firecrawl to handle empty pages gracefully.

gambletan:
When scraping a page that returns a 200 OK status but has empty content, the scraper was treating it as an engine failure and waterfalling to the next engine. This doesn't make sense because:

1. An empty page is a valid page - not an error
2. Waterfalling to another engine won't help since it'll likely return the same empty content
3. This causes a misleading 'SCRAPE_ALL_ENGINES_FAILED' error for valid empty pages

This fix changes the logic to:
- Return success if status code is OK (200-299 or 304), even if content is empty
- Only treat as failure if status code is non-OK AND content is empty

Fixes #2316

---
## Summary by cubic
Treat empty 2xx/304 pages as successful scrapes to stop engine waterfalling and false failure errors. Also fixes user-agent handling in the Playwright service, corrects site: queries for www domains, returns real-time credits in agent status, and adds webhook support to extract.

- **Bug Fixes**
  - Scraper: consider OK status a success even if content is empty; only fail when status is non-OK and content is empty.
  - Playwright: apply custom user-agent on the browser context and remove it from extra headers.
  - Mapping: use the domain without www in site: search queries.
  - Agent status: return real-time creditsUsed via RPC with fallback and logging.

- **New Features**
  - SDK extract: add optional `webhook` in the request payload.

<sup>Written for commit b695b39b28e84b7053de0d315e2ec41d48777283. Summary will update on new commits.</sup>

### DIFF
--- a/apps/api/src/scraper/scrapeURL/index.ts
+++ b/apps/api/src/scraper/scrapeURL/index.ts
@@ -430,11 +430,17 @@ async function scrapeURLLoopIter(
       throw new AddFeatureError(["stealthProxy"]);
     }
 
-    // NOTE: TODO: what to do when status code is bad is tough...
-    // we cannot just rely on text because error messages can be brief and not hit the limit
-    // should we just use all the fallbacks and pick the one with the longest text? - mogery
-    if (isLongEnough || !isGoodStatusCode) {
-      meta.logger.info("Scrape via " + engine + " deemed successful.", {
+    // If we got a good status code (200-299 or 304), treat as success even if content is empty.
+    // Empty pages are valid - they're not errors. Waterfalling to another engine won't help
+    // since they'll likely return the same empty content.
+    if (isGoodStatusCode) {
+      meta.logger.info("Scrape via " + engine + " deemed successful (status code OK, content may be empty).", {
+        factors: { isLongEnough, isGoodStatusCode, hasNoPageError },
+      });
+      return engineResult;
+    } else if (isLongEnough) {
+      // If status code is not OK but we have content, still consider it a partial success
+      meta.logger.info("Scrape via " + engine + " deemed successful (has content despite non-OK status).", {
         factors: { isLongEnough, isGoodStatusCode, hasNoPageError },
       });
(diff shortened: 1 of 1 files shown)


gambletan:
Approved and ready to merge.

PR #3048, opened 2026-03-09
fix: use custom user-agent from headers in playwright context

gambletan:
When a user-agent is provided in the headers of a scrape request, it was being ignored because Playwright's context-level user-agent takes precedence over extra HTTP headers.

This fix:
1. Extracts user-agent from headers before setting extra headers
2. Passes custom user-agent to createContext() 
3. Removes user-agent from headers passed to setExtraHTTPHeaders()

Fixes #2802

---
## Summary by cubic
Honor custom user-agent headers by setting the `playwright` context user-agent so sites see the correct UA. Also skip malformed/basic-auth links during crawl and add `webhook` support to the JS SDK `extract` method.

- **Bug Fixes**
  - `playwright`: Extract `user-agent` from request headers, pass it to the context, and remove it from `setExtraHTTPHeaders`.
  - Crawler: Skip URLs with username/password (malformed mailto or basic-auth) to avoid duplicate/bad links.
  - JS SDK: `extract()` accepts `webhook` (`string` or `WebhookConfig`) and sends it in the request body.

<sup>Written for commit 3dc6a3b32544118cb6101de59132d64d497e567d. Summary will update on new commits.</sup>

### DIFF
--- a/apps/playwright-service-ts/api.ts
+++ b/apps/playwright-service-ts/api.ts
@@ -99,8 +99,8 @@ const initializeBrowser = async () => {
   });
 };
 
-const createContext = async (skipTlsVerification: boolean = false) => {
-  const userAgent = new UserAgent().toString();
+const createContext = async (skipTlsVerification: boolean = false, customUserAgent?: string) => {
+  const userAgent = customUserAgent || new UserAgent().toString();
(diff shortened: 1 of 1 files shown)


gambletan:
Closing in favor of #3054 which has the complete fix.

PR #3050, opened 2026-03-09
fix: properly handle user-agent in headers for Playwright (fix #2802)

gambletan:
## Summary

When calling the scrape endpoint with headers.user-agent, the value was not being used because Playwright sets the user-agent at context level, and then ignores the user-agent key in setExtraHTTPHeaders.

This fix:
1. Extracts user-agent from headers case-insensitively
2. Passes custom user-agent to createContext so it's set at context level
3. Removes user-agent from headers before calling setExtraHTTPHeaders to avoid conflicts

Fixes #2802

---

### Contributor License Agreement

By submitting this pull request, I confirm that I have read and fully agree to the [Contributor License Agreement (CLA)](https://github.com/firecrawl/firecrawl/blob/main/CONTRIBUTOR_LICENSE_AGREEMENT), and I am providing my contributions under its terms.

### DIFF
--- a/apps/playwright-service-ts/api.ts
+++ b/apps/playwright-service-ts/api.ts
@@ -99,8 +99,8 @@ const initializeBrowser = async () => {
   });
 };
 
-const createContext = async (skipTlsVerification: boolean = false) => {
-  const userAgent = new UserAgent().toString();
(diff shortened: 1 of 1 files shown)


gambletan:
Closing in favor of #3054 which has the complete fix.

PR #3054, opened 2026-03-09
fix: use proper case-insensitive user-agent extraction

gambletan:
## Summary

The previous code only checked for lowercase "user-agent" and "User-Agent", missing other case variations like "USER-AGENT" which is valid per HTTP specification.

## Fix

Uses Object.entries().find() with toLowerCase() to properly handle any case variation of the user-agent header.

## Testing

This fix ensures user-agent headers sent in any case are correctly extracted.

---
## Summary by cubic
Fixes case-insensitive User-Agent handling and sets it at the `playwright` context level to avoid header conflicts. Also adds `webhook` support to extract jobs in the `firecrawl` Node.js SDK and defaults `creditsUsed` to 0 when pending.

- **Bug Fixes**
  - `playwright` service: Extract `user-agent` header case-insensitively, pass it to `createContext`, and remove it from per-request headers.
  - `firecrawl` SDK: Add `webhook` to extract payload and types.
  - API: Return `creditsUsed: 0` when agent cost is not yet available.

<sup>Written for commit fb8a698b1b6f497feeeec2327baf6ebc9f36aa2b. Summary will update on new commits.</sup>

### DIFF
--- a/apps/playwright-service-ts/api.ts
+++ b/apps/playwright-service-ts/api.ts
@@ -252,7 +252,7 @@ app.post('/scrape', async (req: Request, res: Response) => {
   let page: Page | null = null;
 
   // Extract user-agent from headers if present (case-insensitive)
-  const customUserAgent = headers?.['user-agent'] || headers?.['User-Agent'];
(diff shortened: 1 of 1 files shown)


gambletan:
This PR is approved — any chance it can be merged? Happy to rebase if needed.

PR #3055, opened 2026-03-09
Initialise autumn

charlietlamb:


docker-compose.yaml
```diff
@@ -40,12 +40,15 @@ x-common-env: &common-env
...
   OLLAMA_BASE_URL: ${OLLAMA_BASE_URL} 
+  AUTUMN_SECRET_KEY: ${AUTUMN_SECRET_KEY}
   SLACK_WEBHOOK_URL: ${SLACK_WEBHOOK_URL}
   BULL_AUTH_KEY: ${BULL_AUTH_KEY}
   TEST_API_KEY: ${TEST_API_KEY}
   SUPABASE_ANON_TOKEN: ${SUPABASE_ANON_TOKEN}
   SUPABASE_URL: ${SUPABASE_URL}
   SUPABASE_SERVICE_TOKEN: ${SUPABASE_SERVICE_TOKEN}
+  SUPABASE_REPLICA_URL: ${SUPABASE_REPLICA_URL}
+  DISABLE_BLOCKLIST: ${DISABLE_BLOCKLIST}
```
micahstairs: How did this diff slip in?

apps/api/src/services/billing/batch_billing.ts
```diff
@@ -134,9 +135,30 @@ export async function processBillingBatch() {
...
         );
+
+        // Track usage in Autumn only after the billing RPC has confirmed success,
+        // so Autumn never overstates usage for operations that ultimately failed.
+        void autumnService.trackCredits({
+          teamId: group.team_id,
+          value: group.total_credits,
+          properties: {
+            source: "processBillingBatch",
+            isExtract: group.is_extract,
```
micahstairs: We don't want isExtract as a property (extract tokens were deprecated, we use credits for everything instead)

apps/api/src/services/billing/batch_billing.ts
```diff
@@ -134,9 +135,30 @@ export async function processBillingBatch() {
...
           `✅ Successfully billed team ${group.team_id} for ${group.total_credits} ${group.is_extract ? "tokens" : "credits"}`,
         );
+
+        // Track usage in Autumn only after the billing RPC has confirmed success,
+        // so Autumn never overstates usage for operations that ultimately failed.
+        void autumnService.trackCredits({
+          teamId: group.team_id,
+          value: group.total_credits,
+          properties: {
+            source: "processBillingBatch",
```
micahstairs: Instead of billing batches, don't we want to be reporting them to autumn on a per-endpoint basis? Or did you have performance concerns about going this route?

apps/api/src/services/billing/batch_billing.ts
```diff
@@ -134,9 +135,30 @@ export async function processBillingBatch() {
...
+        }
+
         logger.info(
           `✅ Successfully billed team ${group.team_id} for ${group.total_credits} ${group.is_extract ? "tokens" : "credits"}`,
         );
+
+        // Track usage in Autumn only after the billing RPC has confirmed success,
+        // so Autumn never overstates usage for operations that ultimately failed.
+        void autumnService.trackCredits({
+          teamId: group.team_id,
```
micahstairs: You will also want to store the orgId as a property

apps/api/src/services/autumn/autumn.service.ts
```diff
@@ -0,0 +1,404 @@
...
+  CreateEntityParams,
+  EnsureOrgProvisionedParams,
+  EnsureTeamProvisionedParams,
+  GetEntityParams,
+  GetOrCreateCustomerParams,
+  TrackCreditsParams,
+  TrackParams,
+} from "./types";
+
+const TEAM_CREDITS_FEATURE_ID = "TEAM_CREDITS";
```
micahstairs: I don't like that this is named TEAM_CREDITS (since we ultimately will be pooling them across the org).

apps/api/src/services/billing/batch_billing.ts
charlietlamb: Right now since billing is async and queued through Redis, there's no Supabase commit at the endpoint level. Calling `autumn.track` there would mean we could report credits that never actually billed (e.g. if the batch processor crashes or drops the item). Doing it here, after `bill_team_6` confirms success and keeps Autumn in sync with what was actually charged.

---

After some more thought I think a cleaner solution here may be to reserve the credits upfront and then issue a refund if something does go wrong - will look at implementing this approach

docker-compose.yaml
charlietlamb: Was having issues running locally without these in here - will remove to keep this pr clean

apps/api/src/services/billing/batch_billing.ts
charlietlamb: `orgId` will always be the `customerId` in autumn and we'll have this for each event so we shouldn't need to pass this

apps/api/src/services/billing/credit_billing.ts
```diff
@@ -29,13 +30,20 @@ export async function billTeam(
       api_key_id: number | null,
       logger: Logger | undefined,
     ) => {
-      // Within the authenticated context, queue the billing operation
+      // Reserve in Autumn at request time; await so autumnReserved is accurate.
+      // billTeam is fire-and-forget at call sites, so this doesn't block responses.
+      const autumnReserved = await autumnService.reserveCredits({
+        teamId: team_id,
+        value: credits,
+        properties: { source: "billTeam" },
```
micahstairs: Where are we storing info about which API key was used or which API endpoint the credits are being used for?

apps/api/src/controllers/v0/admin/create-user.ts
```diff
@@ -351,6 +356,8 @@ export async function integCreateUserController(req: Request, res: Response) {
 
       apiKey = apiKeyFc.key;
 
+      await autumnService.ensureTeamProvisioned({ teamId });
```
micahstairs: should we wrap all of this autumn stuff in try-catches to make the rollout safer?

charlietlamb: Added a try/catch within this method

apps/api/src/services/billing/credit_billing.ts
charlietlamb: Added here

micahstairs (approved):

PR #3120, opened 2026-03-11
fix: use user-provided user-agent header in playwright scrape

fuleinist:
## Summary

This fix ensures that when a user-agent header is provided in the scrape request headers, it is properly used by playwright instead of being ignored.

## Changes

- Modified user-agent handling in apps/playwright-service-ts/api.ts to prioritize user-provided headers
- The user-agent from request headers now takes precedence over the default UserAgent package value

## Testing

- Verified the fix preserves backward compatibility (default behavior unchanged when no user-agent header is provided)
- Tested that explicit user-agent headers are now respected

Fixes #2802

---
## Summary by cubic
Respect the user-provided `user-agent` header in the `playwright` scrape endpoint, falling back to a generated `UserAgent` only when none is provided. Header lookup is now case-insensitive, and the value is applied at browser context creation so all requests use the exact agent clients specify.

<sup>Written for commit 1a3470b0751379e31051a263f109e365e0d34b0c. Summary will update on new commits.</sup>

### DIFF
--- a/apps/playwright-service-ts/api.ts
+++ b/apps/playwright-service-ts/api.ts
@@ -99,8 +99,8 @@ const initializeBrowser = async () => {
   });
 };
 
-const createContext = async (skipTlsVerification: boolean = false) => {
-  const userAgent = new UserAgent().toString();
+const createContext = async (skipTlsVerification: boolean = false, customUserAgent?: string) => {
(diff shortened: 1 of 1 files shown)


fuleinist:
Stale PR (no activity for 4+ weeks). Feel free to reopen if still relevant.

PR #3123, opened 2026-03-12
feat: add MiniMax M3 provider support

octo-patch:
Reason: add target provider/model to existing provider registry

## Summary

- Add MiniMax provider selection with global and mainland China endpoints for both compatible API formats.
- Register MiniMax-M3 and MiniMax-M2.7 with current context, tiered pricing, cache, reasoning, and input modality metadata.
- Record adaptive and disabled thinking for MiniMax-M3 and always-on thinking for MiniMax-M2.7.
- Document self-hosted configuration and pass the new settings through Docker Compose.
- Add coverage for every supported MiniMax endpoint combination and model metadata.

## Checks

- `pnpm -C apps/api exec prettier --check src/lib/extract/usage/model-prices.ts src/lib/minimax.test.ts`
- `pnpm -C apps/api exec vitest run src/lib/minimax.test.ts`


octo-patch:
Updated MiniMax-M3 pricing to preserve the documented input-length and service tiers, and wired both cost estimators to select the standard tier at the 512k boundary. I also added request-capture coverage for all region and API-format combinations. Checks: `pnpm exec vitest run src/lib/minimax.test.ts src/lib/extract/usage/llm-cost.test.ts` and `pnpm build`.

octo-patch:
Thanks for the review. I updated MiniMax-M3 to the current flat token rates and removed the obsolete tier-specific estimator changes. I ran `pnpm exec vitest run src/lib/minimax.test.ts` (10 tests passed) and `git diff --check`; `pnpm build` remains blocked locally because `@mendable/firecrawl-rs` is unavailable.

PR #3170, opened 2026-03-17
fix: respect USE_RESPONSES_ENDPOINT env var for OpenAI-compatible providers

### LINKED ISSUE #3169: LLM extract/AI features fail with OpenAI-compatible proxies: USE_RESPONSES_ENDPOINT env var is not wired up
## Bug Description

When self-hosting Firecrawl with an OpenAI-compatible proxy (e.g. Azure OpenAI, LiteLLM, vLLM, or any provider that only implements the Chat Completions API), all AI-powered features (extract, etc.) fail with a **404 Not Found** error.

## Root Cause

Since `@ai-sdk/openai` v2+ (AI SDK 5), calling `provider(modelName)` defaults to the **OpenAI Responses API** (`POST /v1/responses`). Most OpenAI-compatible providers only implement the **Chat Completions API** (`POST /v1/chat/completions`).

In `apps/api/src/lib/generic-ai.ts`, the `getModel()` function calls:

```ts
return providerList[provider](modelName);  // → hits /v1/responses
```

Only `o3-mini` models have a special case to use `.chat()`:

```ts
if (provider === "openai" && modelName.startsWith("o3-mini")) {
  return providerList.openai.chat(modelName);
}
```

There is no way for self-hosted users to opt out of the Responses API. The `USE_RESPONSES_ENDPOINT` env var (mentioned in some deployment guides) is **not wired up** anywhere in the Firecrawl codebase.

## Steps to Reproduce

1. Self-host Firecrawl with Docker
2. Configure `OPENAI_BASE_URL` to point to an OpenAI-compatible proxy that doesn't support `/v1/responses`
3. Set `OPENAI_API_KEY` and `MODEL_NAME` appropriately
4. Try to use the `/extract` endpoint or any AI-powered feature
5. Observe: `provider API error: 404 Not Found` in the API logs, with requests going to `<base_url>/responses`

## Expected Behavior

Setting `USE_RESPONSES_ENDPOINT=false` should force Firecrawl to use the Chat Completions API (`/v1/chat/completions`) instead of the Responses API (`/v1/responses`).

## Environment

- Firecrawl: self-hosted (latest Docker image)
- `@ai-sdk/openai`: v2.0.64
- `ai`: v5.0.89
- Any OpenAI-compatible proxy that doesn't implement `/v1/responses`

rajpopat27:
## Summary

Fixes #3169

- Adds `USE_RESPONSES_ENDPOINT` config option to allow self-hosted users to disable the OpenAI Responses API
- When set to `false`, forces `getModel()` to use `providerList.openai.chat()` (Chat Completions API) instead of `providerList.openai()` (Responses API)
- Documents the new env var in `.env.example`

## Problem

Since `@ai-sdk/openai` v2+ (AI SDK 5), calling `provider(modelName)` defaults to the **Responses API** (`POST /v1/responses`). Most OpenAI-compatible proxies and providers (Azure OpenAI, LiteLLM, vLLM, etc.) only implement the **Chat Completions API** (`POST /v1/chat/completions`), causing all AI-powered features to fail with `404 Not Found`.

### `apps/api/src/config.ts`
- Added `USE_RESPONSES_ENDPOINT` env var (string, optional, defaults to `true`)
- When set to `"false"`, the config value becomes `false`; otherwise it stays truthy

### `apps/api/src/lib/generic-ai.ts`
- Extended the existing `o3-mini` special case to also check `config.USE_RESPONSES_ENDPOINT`
- When `USE_RESPONSES_ENDPOINT` is `false`, uses `.chat()` to force the Chat Completions API

### `apps/api/.env.example`
- Documented `OPENAI_BASE_URL`, `MODEL_NAME`, and `USE_RESPONSES_ENDPOINT` env vars

## Backward Compatibility

- Defaults to `true` (Responses API), so existing deployments are unaffected
- Only changes behavior when explicitly set to `USE_RESPONSES_ENDPOINT=false`

---
## Summary by cubic
Respects the `USE_RESPONSES_ENDPOINT` env var so OpenAI-compatible providers can force the Chat Completions API and avoid 404s. Defaults to the Responses API for backward compatibility.

- **Bug Fixes**
  - Added `USE_RESPONSES_ENDPOINT` (defaults to true via `z.stringbool()`); when "false", use `providerList.openai.chat()` (`/v1/chat/completions`).
  - Updated `getModel()` to honor the flag and always use Chat Completions for `o3-mini`.
  - Documented `OPENAI_BASE_URL`, `MODEL_NAME`, and `USE_RESPONSES_ENDPOINT` in `apps/api/.env.example` and `SELF_HOST.md`.

<sup>Written for commit bf8b8813ca2a7b801fc703f57de5d9f708cd966c. Summary will update on new commits.</sup>

### DIFF
--- a/apps/api/src/config.ts
+++ b/apps/api/src/config.ts
@@ -188,7 +188,7 @@ const configSchema = z.object({
   MODEL_EMBEDDING_NAME: z.string().optional(),
   OLLAMA_BASE_URL: z.string().optional(),
   VERTEX_CREDENTIALS: z.string().optional(),
-  USE_RESPONSES_ENDPOINT: z.coerce.boolean().default(true),
+  USE_RESPONSES_ENDPOINT: z.stringbool().default(true),
 
   // Rate Limiting
   RATE_LIMIT_TEST_API_KEY_SCRAPE: z.coerce.number().optional(),


apps/api/src/config.ts
```diff
@@ -188,6 +188,10 @@ const configSchema = z.object({
   MODEL_EMBEDDING_NAME: z.string().optional(),
   OLLAMA_BASE_URL: z.string().optional(),
   VERTEX_CREDENTIALS: z.string().optional(),
+  USE_RESPONSES_ENDPOINT: z
+    .string()
+    .optional()
+    .transform((val) => val !== "false"),
```
mogery: Use z.coerce.boolean().default(true)

apps/api/.env.example
```diff
@@ -48,6 +48,14 @@ RATE_LIMIT_TEST_API_KEY_CRAWL=
...
 # add for LLM dependent features (image alt generation, etc.)
 OPENAI_API_KEY=
+# set to use a custom OpenAI-compatible base URL (e.g. Azure, proxies, local models)
+# OPENAI_BASE_URL=
+# set a custom model name when using OPENAI_BASE_URL
+# MODEL_NAME=
+# set to "false" to force the OpenAI Chat Completions API instead of the
+# Responses API. Required for most OpenAI-compatible proxies and providers
+# that do not implement the /v1/responses endpoint.
+# USE_RESPONSES_ENDPOINT=false
```
mogery, line 58: Can you document this in SELF_HOST.md too?

apps/api/src/config.ts
rajpopat27: Done - switched to `z.coerce.boolean().default(true)`

apps/api/.env.example
rajpopat27: Done - added to SELF_HOST.md under the "Experimental: Use any OpenAI-compatible API" section.

PR #3206, opened 2026-03-21
ci: optimize Playwright caching in test workflow

abimaelmartell:
## Summary

Optimizes the Playwright setup in the CI test-server workflow by splitting the monolithic install step into discrete steps, skipping the Chromium download (~150MB) on cache hit, scoping `install-deps` to chromium only, and adding `pnpm fetch` for playwright-service-ts. Reduces Playwright setup time from ~1m to ~12s on cache hit.

## Test plan

- [x] Verify the `playwright` matrix jobs pass in CI
- [x] Confirm the Chromium install step is skipped on cache hit
- [x] Confirm system deps step is faster with chromium-only scope

### DIFF
--- a/.github/workflows/test-server.yml
+++ b/.github/workflows/test-server.yml
@@ -158,7 +158,7 @@ jobs:
 
       - name: Install Playwright system deps
         if: matrix.engine == 'playwright'
-        run: pnpm exec playwright install-deps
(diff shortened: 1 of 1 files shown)


abimaelmartell:
@cubic-dev-ai

PR #3245, opened 2026-03-29
fix: time_taken calculation and OpenAI proxy support

### LINKED ISSUE #3165: bug: v1 /map endpoint time_taken is always ~0
## Bug Description

The `time_taken` metric returned by the v1 `/map` endpoint is always approximately 0, regardless of how long the request actually takes.

## Root Cause

In `apps/api/src/controllers/v1/map.ts`, the `getMapResults` function calculates `time_taken` as:

```ts
time_taken: (new Date().getTime() - Date.now()) / 1000,
```

`new Date().getTime()` and `Date.now()` both return the current timestamp in milliseconds at the moment they're called. Since they execute on the same line within microseconds of each other, the difference is always ~0.

The v2 map endpoint (`apps/api/src/lib/map-utils.ts`) correctly captures the start time at function entry and subtracts it at the end. The v1 endpoint missed this pattern.

## Expected Behavior

`time_taken` should reflect actual elapsed time (in seconds) from function start to completion.

## Fix

Capture `Date.now()` at the start of `getMapResults` and use it in the subtraction:

```ts
const functionStartTime = Date.now();
// ...
time_taken: (Date.now() - functionStartTime) / 1000,
```

### LINKED ISSUE #3169: LLM extract/AI features fail with OpenAI-compatible proxies: USE_RESPONSES_ENDPOINT env var is not wired up

xingzihai:
## Summary
This PR fixes two issues:

### Fix 1: Correct time_taken calculation in v1 /map endpoint (Closes #3165)
The `time_taken` field was incorrectly calculated. This fix ensures accurate timing.

### Fix 2: Wire up USE_RESPONSES_ENDPOINT for OpenAI-compatible proxies (Closes #3169)
Added support for the `USE_RESPONSES_ENDPOINT` environment variable, enabling proper integration with OpenAI-compatible proxy services.

## Changes
- Fixed time_taken calculation in v1 map endpoint
- Wired up USE_RESPONSES_ENDPOINT env var
- Added documentation for USE_RESPONSES_ENDPOINT in SELF_HOST.md

## Testing
- Verified the changes work correctly with OpenAI-compatible proxies

cc @mendableai

---
## Summary by cubic
Fixes incorrect `time_taken` in the v1 /map endpoint and adds OpenAI-compatible proxy support by switching between Responses and Chat Completions via `USE_RESPONSES_ENDPOINT`.

- **Bug Fixes**
  - Correct `time_taken` by capturing start time and computing duration in v1 /map.

- **New Features**
  - Add `USE_RESPONSES_ENDPOINT` to choose Responses or Chat Completions for OpenAI. Defaults to Chat Completions when `OPENAI_BASE_URL` is set; otherwise Responses. Documented in `SELF_HOST.md`.

<sup>Written for commit 78679595b0454e628a2f8079fe0fb443d845cf01. Summary will update on new commits.</sup>

### DIFF
--- a/SELF_HOST.md
+++ b/SELF_HOST.md
@@ -62,6 +62,11 @@ USE_DB_AUTHENTICATION=false
 # OPENAI_BASE_URL=https://example.com/v1
 # OPENAI_API_KEY=
 
+# When using OpenAI-compatible proxies (Azure, LiteLLM, vLLM, etc.), set this to false
+# to use the Chat Completions API (/v1/chat/completions) instead of the Responses API (/v1/responses)
+# If OPENAI_BASE_URL is set and USE_RESPONSES_ENDPOINT is not specified, Chat Completions API is used automatically
+# USE_RESPONSES_ENDPOINT=false
+
 ## === Proxy ===
 # PROXY_SERVER can be a full URL (e.g. http://0.1.2.3:1234) or just an IP and port combo (e.g. 0.1.2.3:1234)
 # Do not uncomment PROXY_USERNAME and PROXY_PASSWORD if your proxy is unauthenticated


micahstairs:
Already merged a fix for the first problem. Going to close this now.

In the future, please open a separate PR for each bug fix.

PR #3357, opened 2026-04-13
feat(interact): add langsmith grouped by browser session

developersdigest:
## Summary by cubic
Adds LangSmith tracing to the Interact agent, grouping runs by browser session and enriching traces with scrape context and identity. Tracing is opt-in (API key + `LANGSMITH_TRACING=true`), skipped for forced zero-data-retention teams, and URLs are sanitized.

- **New Features**
  - Wraps prompt loop and code-exec with LangSmith when enabled; groups by `session.id` as `thread_id` with run names `interact:prompt` and `interact:code`.
  - Adds scrape metadata (`scrape_url`, `target_url`, `scrape_wait_for_ms`, `scrape_actions`, `scrape_origin`) and identity fields (`org_id`, `sub_user_id`) for filtering; URLs sanitized to remove queries/fragments.
  - Adds `langsmith` and supports `LANGSMITH_API_KEY`, `LANGSMITH_PROJECT`, `LANGSMITH_ENDPOINT`, `LANGSMITH_TRACING`.

- **Bug Fixes**
  - Require `LANGSMITH_API_KEY` and `LANGSMITH_TRACING=true`; whitespace-only keys disabled; only mirror `LANGSMITH_*` into `process.env` after init succeeds.
  - Ensure URL sanitization strips query/fragment even for malformed inputs.
  - Record code-exec inputs in traces by passing exec params through `traceable`.

<sup>Written for commit 57658dd39c9b804cb42c26f418ec5dcdb3024552. Summary will update on new commits.</sup>


rafaelsideguide (commented):
**1. `sanitizeUrlForTrace` leaks query params when URL parsing fails.**
The `catch` block returns the raw string, including any `?token=...` fragments. Replace the `try/catch` with a simple string split that never throws: `url.split("?")[0].split("#")[0]`.

> **Agent fix:** In `apps/api/src/lib/scrape-interact/langsmith.ts`, rewrite `sanitizeUrlForTrace` to drop the `try/catch` and return `url.split("?")[0].split("#")[0]` instead of calling `new URL()`. Update the corresponding unit tests in `langsmith.test.ts` — the "Invalid URL → returned as-is" test case should now expect the raw string without query/fragment rather than the full raw string.

---

**2. (suggestion) Code-path tracing is assembled in the controller instead of the agent layer.**
The prompt path keeps its LangSmith setup inside `executePromptViaBrowserAgent`; the code path puts equivalent logic directly in `scrapeInteractController`, making the controller own tracing details it shouldn't need to know about.

> **Agent fix:** In `apps/api/src/lib/scrape-interact/browser-agent.ts`, introduce a new exported function (e.g. `executeCodeViaBrowserSession`) that wraps the `browserServiceRequest` call and accepts the same `BrowserAgentTraceContext` parameter that `executePromptViaBrowserAgent` already accepts. Move the `traceInteract(...)` call and `traceScrapeContext` spreading from `scrape-browser.ts` into that new function. Update `scrape-browser.ts` to call the new function instead of assembling the trace inline.

---

**3. `isLangSmithEnabled` is opt-out instead of opt-in.**
Any deployment that sets `LANGSMITH_API_KEY` starts shipping traces to LangSmith even without `LANGSMITH_TRACING=true`, which can surprise operators.

> **Agent fix:** In `apps/api/src/lib/scrape-interact/langsmith.ts`, simplify the `isLangSmithEnabled` expression to `Boolean(LANGSMITH_API_KEY && config.LANGSMITH_TRACING === true)`, removing the `=== undefined ? true` branch. Update `langsmith.test.ts` to add a case asserting that a key alone (without `LANGSMITH_TRACING: true`) results in `isLangSmithEnabled === false`.

---

**4. Enabled-path unit tests are not hermetic.**
The disabled-path tests mock `../../config` to isolate from the developer's local `.env`; the enabled-path tests skip this and rely on `process.env` flowing through the zod schema at re-require time, which can break in environments missing required config fields.

> **Agent fix:** In the `"enabled — mocked SDK"` describe block in `langsmith.test.ts`, add a `jest.doMock("../../config", ...)` call inside `beforeEach` (after `jest.resetModules()`) that provides the minimal config needed: `LANGSMITH_API_KEY: "test-fake-key"`, `LANGSMITH_TRACING: true`, and `undefined` for the rest. Add the corresponding `jest.dontMock("../../config")` in `afterEach`.

PR #3399, opened 2026-04-20
fire-pdf: send explicit {timeout, created_at} for deadline contract

abimaelmartell:
## Summary by cubic
Send an explicit deadline (`timeout`, `created_at`) to `fire-pdf` and enrich the POST payload for better observability. Improves timeout behavior and makes fallbacks and jobs easier to trace.

- **New Features**
  - Include `timeout` and `created_at` in `fire-pdf` POST body to enforce caller deadlines; expired requests return 503. Omit when no scrape-tier deadline so service default applies.
  - Send job metadata: `team_id`, `crawl_id`, `url` (omitted with zero data retention), `pdf_sha256`, `source`, and `zdr` for the jobs dashboard.
  - Export `createPdfCacheKey` and use it to compute `pdf_sha256`.
  - Log structured `event=pdf_engine_fallback` with IDs and error details when falling back from FirePDF to MinerU.

- **Bug Fixes**
  - Coerce `error_class` and `error_message` to strings in fallback logs to prevent logger errors when non-Error values are thrown.

<sup>Written for commit 6bbdf295dfd793094c2db09814ca72d16d7001fb. Summary will update on new commits.</sup>

### DIFF
--- a/apps/api/src/lib/gcs-pdf-cache.ts
+++ b/apps/api/src/lib/gcs-pdf-cache.ts
@@ -11,7 +11,7 @@ const PROVIDER_PREFIXES: Record<PdfCacheProvider, string> = {
   firepdf: "pdf-cache-firepdf/",
 };
 
-function createPdfCacheKey(pdfContent: string | Buffer): string {
+export function createPdfCacheKey(pdfContent: string | Buffer): string {
   return crypto.createHash("sha256").update(pdfContent).digest("hex");
 }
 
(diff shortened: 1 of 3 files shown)


abimaelmartell:
Addressed. Coerced both `error_class` (`.name`) and `error_message` (`.message`) with `String(...)` so non-Error throws can't break the fallback logger via `.slice` on undefined. Pushed as 6bbdf295d.

PR #3448, opened 2026-04-27
fix(api): accept zeroDataRetention in /v2/search scrapeOptions

### LINKED ISSUE #3441: [Bug] /v2/search rejects documented scrapeOptions.zeroDataRetention
**Describe the Bug**

The Search documentation describes combining Search ZDR with scrape-side ZDR by passing both top-level `enterprise` and `scrapeOptions.zeroDataRetention` to `/v2/search`:

```json
{
  "query": "sensitive topic",
  "limit": 5,
  "enterprise": ["zdr"],
  "scrapeOptions": {
    "formats": ["markdown"],
    "zeroDataRetention": true
  }
}
```

Docs page: https://docs.firecrawl.dev/features/search

The docs state that when using Search with content scraping, the top-level `enterprise` parameter covers the search portion while `zeroDataRetention` in `scrapeOptions` covers the scraping portion.

However, the live `/v2/search` endpoint rejects `scrapeOptions.zeroDataRetention` as an invalid request-body field.

Top-level `enterprise` is recognized by the endpoint, including validation for `"default"`, `"anon"`, and `"zdr"`. The issue appears specific to `zeroDataRetention` inside Search `scrapeOptions`.

**To Reproduce**

Send a `/v2/search` request with `scrapeOptions.zeroDataRetention`:

```bash
curl -X POST https://api.firecrawl.dev/v2/search \
  -H "Authorization: Bearer $FIRECRAWL_API_KEY" \
  -H "Content-Type: application/json" \
  -d '{
    "query": "Firecrawl Search docs",
    "limit": 1,
    "scrapeOptions": {
      "formats": ["markdown"],
      "zeroDataRetention": true
    }
  }'
```

The same behavior occurs when combining it with documented Search ZDR:

```bash
curl -X POST https://api.firecrawl.dev/v2/search \
  -H "Authorization: Bearer $FIRECRAWL_API_KEY" \
  -H "Content-Type: application/json" \
  -d '{
    "query": "Firecrawl Search docs",
    "limit": 1,
    "enterprise": ["zdr"],
    "scrapeOptions": {
      "formats": ["markdown"],
      "zeroDataRetention": true
    }
  }'
```

**Expected Behavior**

Either:

1. `/v2/search` should accept `scrapeOptions.zeroDataRetention` as documented, so callers can combine Search ZDR and scrape-side ZDR in a single Search request.

or:

2. The Search docs should be updated if scrape-side `zeroDataRetention` is not currently supported inside Search `scrapeOptions`.

**Screenshots**

Not applicable.

- OS: macOS
- Deployment Type: Cloud
- Firecrawl Version: `/v2/search` API, checked against Firecrawl Cloud
- Node.js Version: Not Node-specific. Local validation used Node v25.2.1.

**Logs**

The endpoint returns `400 Invalid request body` when `scrapeOptions.zeroDataRetention` is present in a `/v2/search` request.

In live testing, `/v2/search` accepted the `enterprise` field structurally but rejected `zeroDataRetention` under `scrapeOptions`.

**Additional Context**

This matters for SDK/MCP/tool integrations because Search supports scraping result content via `scrapeOptions`, and the docs explicitly describe end-to-end ZDR for Search plus scraping. Callers need to know whether the documented combined request shape is supported.

This is separate from team entitlement for Search ZDR. The reported issue is request-body validation for `scrapeOptions.zeroDataRetention`.

octo-patch:
Fixes #3441

## Problem

The `/v2/search` endpoint rejects requests containing `scrapeOptions.zeroDataRetention` with a `400 Invalid request body` error, even though the documentation explicitly describes using `scrapeOptions.zeroDataRetention` to enable scrape-side zero data retention when combining search with content scraping.

The root cause is that the `searchRequestSchema`'s `scrapeOptions` field is built from `baseScrapeOptions.extend({ formats: ... })` but does not include `zeroDataRetention`, while other endpoints (scrape, batch scrape, crawl) do include it in their top-level schema.

## Solution

1. **`apps/api/src/controllers/v2/types.ts`**: Add `zeroDataRetention: z.boolean().optional()` to the `scrapeOptions` extension in `searchRequestSchema` so the field is accepted by Zod validation.

2. **`apps/api/src/controllers/v2/search.ts`**: Update the controller to also apply `scrapeOptions.zeroDataRetention` to the `zeroDataRetention` context flag, enabling scrape-side ZDR when the caller passes it via `scrapeOptions`.

## Testing

The fix can be verified by sending a `/v2/search` request with `scrapeOptions.zeroDataRetention: true` — it should no longer return `400 Invalid request body`.

```bash
curl -X POST https://api.firecrawl.dev/v2/search \
  -H "Authorization: Bearer $FIRECRAWL_API_KEY" \
  -H "Content-Type: application/json" \
  -d '{
    "query": "Firecrawl Search docs",
    "limit": 1,
    "scrapeOptions": {
      "formats": ["markdown"],
      "zeroDataRetention": true
    }
  }'
```

---
## Summary by cubic
Accepts and fully applies `scrapeOptions.zeroDataRetention` in `/v2/search`, fixing the 400 error and ensuring scrape-side zero data retention is enforced end-to-end.

- **Bug Fixes**
  - Add `zeroDataRetention: z.boolean().optional()` to `searchRequestSchema.scrapeOptions`.
  - Propagate the combined ZDR flag to scope and downstream calls (applyZdrScope, request logging, `executeSearch`, search logging).

<sup>Written for commit b63bda1e0bb1e2c07e9c1a0fc0e0df5da225fe97. Summary will update on new commits.</sup>

### DIFF
--- a/apps/api/src/controllers/v2/search.ts
+++ b/apps/api/src/controllers/v2/search.ts
@@ -89,7 +89,7 @@ export async function searchController(
     const isZDROrAnon = isZDR || isAnon;
     const scrapeOptionsZDR = req.body.scrapeOptions?.zeroDataRetention ?? false;
     zeroDataRetention = (isZDROrAnon ?? false) || scrapeOptionsZDR;
-    applyZdrScope(isZDROrAnon ?? false);
+    applyZdrScope(zeroDataRetention);
 
     // Verify the team has searchZDR enabled before allowing enterprise ZDR/anon
     if (isZDROrAnon) {
@@ -112,7 +112,7 @@ export async function searchController(
         origin: req.body.origin ?? "api",
         integration: req.body.integration,
         target_hint: req.body.query,
-        zeroDataRetention: isZDROrAnon ?? false,
+        zeroDataRetention: zeroDataRetention,
         api_key_id: req.acuc?.api_key_id ?? null,
       });
     }
@@ -141,7 +141,7 @@ export async function searchController(
         jobId,
         apiVersion: "v2",
         bypassBilling: !shouldBill,
-        zeroDataRetention: isZDROrAnon,
+        zeroDataRetention: zeroDataRetention,
         billing,
         agentIndexOnly: (req as any).agentIndexOnly ?? false,
       },
@@ -179,7 +179,7 @@ export async function searchController(
         team_id: req.auth.team_id,
         options: req.body,
         credits_cost: shouldBill ? result.searchCredits : 0,
-        zeroDataRetention: isZDROrAnon ?? false,
+        zeroDataRetention: zeroDataRetention,
       },
       false,
     );
--- a/apps/api/src/controllers/v2/search.ts
+++ b/apps/api/src/controllers/v2/search.ts
@@ -87,7 +87,8 @@ export async function searchController(
     const isZDR = req.body.enterprise?.includes("zdr");
     const isAnon = req.body.enterprise?.includes("anon");
     const isZDROrAnon = isZDR || isAnon;
-    zeroDataRetention = isZDROrAnon ?? false;
+    const scrapeOptionsZDR = req.body.scrapeOptions?.zeroDataRetention ?? false;
+    zeroDataRetention = (isZDROrAnon ?? false) || scrapeOptionsZDR;
     applyZdrScope(isZDROrAnon ?? false);
 
     // Verify the team has searchZDR enabled before allowing enterprise ZDR/anon
--- a/apps/api/src/controllers/v2/types.ts
+++ b/apps/api/src/controllers/v2/types.ts
@@ -1780,6 +1780,7 @@ export const searchRequestSchema = z
           .refine(x => {
             return x.filter(f => f.type === "screenshot").length <= 1;
           }, "You may only specify one screenshot format"),
+        zeroDataRetention: z.boolean().optional(),
       })
       .optional(),
     __agentInterop: z


octo-patch:
Good catch @cubic-dev-ai — fixed in b63bda1. The combined `zeroDataRetention` value is now passed to `applyZdrScope`, the `executeSearch` context, the `logRequest`, and `logSearch` calls so the `scrapeOptions.zeroDataRetention` flag actually takes effect end-to-end. The enterprise `searchZDR` team-flag gate still keys off `isZDROrAnon` since that permission only applies to the enterprise/anon path.

mogery:
This is not supported on purpose, as we cannot provide end-to-end ZDR using our default search providers.

mogery:
My bad, misunderstood. Taking a look.

mogery (approved):

apps/api/src/controllers/v2/search.ts
```diff
@@ -87,8 +87,9 @@ export async function searchController(
     const isZDR = req.body.enterprise?.includes("zdr");
     const isAnon = req.body.enterprise?.includes("anon");
     const isZDROrAnon = isZDR || isAnon;
-    zeroDataRetention = isZDROrAnon ?? false;
-    applyZdrScope(isZDROrAnon ?? false);
+    const scrapeOptionsZDR = req.body.scrapeOptions?.zeroDataRetention ?? false;
```
mogery, line 90: I believe there's a check on other endpoints that determines whether the user has the permission to turn on ZDR or not. We should add that here as well.

mogery (changes requested):

apps/api/src/controllers/v2/search.ts
02ez: Agreed. This looks like the remaining blocker.

The current PR accepts and propagates `scrapeOptions.zeroDataRetention`, but before enabling it, the controller should apply the same ZDR entitlement check used by the other endpoints that already support request-level ZDR.

Suggested fix:

- keep `zeroDataRetention: z.boolean().optional()` in `searchRequestSchema.scrapeOptions`
- compute `scrapeOptionsZDR = req.body.scrapeOptions?.zeroDataRetention ?? false`
- keep the existing Search enterprise ZDR/anon check for `enterprise: ["zdr"]` / `enterprise: ["anon"]`
- add the same permission check used by scrape/batch/crawl before allowing `scrapeOptionsZDR`
- then pass the combined `zeroDataRetention` value into `applyZdrScope`, `executeSearch`, `logRequest`, and `logSearch`
- add a focused regression test for the unauthorized `scrapeOptions.zeroDataRetention` case

That should address this thread without changing the existing enterprise Search ZDR behavior.

02ez (commented):
Commented on the unresolved ZDR entitlement check.

PR #3519, opened 2026-05-13
fix(api): reject AVIF/HEIF/MP4 served as application/octet-stream

johnboxall:
Hey friends,

Thanks for the most excellent service.

While using `https://api.firecrawl.dev/v1/scrape?format=markdown` we observed the endpoint returns binary content if the user provided a video URL.

Since we're asking for `markdown` format, I think it makes more sense to get back an `UnsupportedFileError`.

This PR adds detection for these sorts of files and rejects them.

I've tried to do this in a way that is consistent with the existing codebase.

Thanks!

🙏 🙏 🙏 

---
## Summary by cubic
Rejects AVIF/HEIF/HEIC and MP4/MOV files served as `application/octet-stream` so `format=markdown` no longer returns binary mojibake and instead fails with a clear `UnsupportedFileError`.

- **Bug Fixes**
  - Detects ISOBMFF files by the `ftyp` box at byte offset 4 (brands for AVIF/HEIF/HEIC, MP4, QuickTime) when the response is `application/octet-stream`.
  - Throws `UnsupportedFileError` with the inferred MIME instead of passing binary through to markdown.
  - Adds `detectIsobmffMime` helper and unit tests.

<sup>Written for commit e55280fb6cb5d537b5f25f7d2a3b7da87a951267. Summary will update on new commits.</sup>

### DIFF
--- a/apps/api/src/scraper/scrapeURL/engines/utils/__tests__/detectIsobmffMime.test.ts
+++ b/apps/api/src/scraper/scrapeURL/engines/utils/__tests__/detectIsobmffMime.test.ts
@@ -1,0 +1,76 @@
+import { detectIsobmffMime } from "../specialtyHandler";
+
+// Builds a minimal 24-byte ISOBMFF "ftyp" box with the given major brand,
+// matching what real AVIF/HEIC/MP4 files have at byte offset 0.
+function ftypBox(brand: string): Buffer {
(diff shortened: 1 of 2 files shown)


mkotsollaris:
image (31)
Screenshot 2026-05-13 at 3 09 39 PM

Attaching some related info to this issue

PR #3525, opened 2026-05-13
feat(api): Implement fire-privacy

abimaelmartell:
## Summary

Adds a `pii` format and `redactPII: boolean` option to v2 scrape. When both are set, the worker calls fire-privacy `/redact` with `mode: model`, `operator: replace`, `language: en` and attaches a `pii` block to the document:

```json
{
  "markdown": "...",
  "pii": {
    "status": "ok" | "skipped" | "error" | "service_at_capacity" | "timeout",
    "redactedMarkdown": "... or null on failure",
    "spans": [{ "start", "end", "kind", "score", "source" }],
    "truncatedAt": null
  }
}
```

Original `markdown` is left intact. Fail-soft: 503 → `service_at_capacity`, 5xx/413 → `error`, >5s → `timeout` — the scrape still succeeds in all of these.

Boolean-only on purpose so the surface can grow to `bool | object` later without a break.

## Notes

- New config: `FIRE_PRIVACY_URL` (defaults to the in-cluster URL), `FIRE_PRIVACY_TIMEOUT_MS` (5000)
- Schema refines applied to scrape, batch-scrape, parse, crawl, extract — `redactPII: true` requires `pii` in formats
- `coerceFieldsToFormats` drops any stale `pii` block when the request doesn't ask for one (e.g. a cached doc reused without `redactPII: true`)
- `pii` format also triggers markdown derivation, since spans are markdown char offsets

## Open question

The plan sends `mode: "model"` and reads `model_status` / `model_truncated_at`. The fire-privacy README documents neither field. The client is tolerant — defaults `model_status` to `ok` and `model_truncated_at` to `null` when absent — so we work against either version. Worth confirming with fire-privacy owners before SDKs lock onto it.

---
## Summary by cubic
Adds PII redaction to v2 scrape via `fire-privacy` with a new `pii` format and `redactPII` option. The redacted output now replaces `document.markdown` (fail-closed), with chunked processing, a 250KB ceiling, and the scrape still succeeds regardless of redaction outcome.

- **New Features**
  - API: `formats: ["pii"]` and `redactPII: boolean | { mode, entities, replaceStyle }`. Defaults: `true` → `{ mode: "accurate", replaceStyle: "tag" }`. Unknown modes/entities are rejected. Maps to `fire-privacy` `mode: "model" | "both" | "heuristics"`, `operator: "replace" | "mask" | "redact"`, `language: "en"`.
  - Response: adds `pii` with `status: "ok" | "skipped" | "failed"`, optional `reason`, `redactedMarkdown`, `spans` (with `entity`, `kind`, `source: "model" | "heuristics" | "unknown"`, optional `score`), and per-entity `counts`.
  - Execution: markdown is chunked and fanned out (c=3); >250KB → `skipped` with `reason: "too_large"` (no request). Any chunk failure fails the whole redaction. Entity allowlists filter spans and re-render output (tag/mask/remove); otherwise upstream `redacted_text` is used. PII-only markdown derivation runs only when `redactPII` is on. Invalid chunk limits throw RangeError.
  - Document: after redaction, `document.markdown` is replaced with `pii.redactedMarkdown`; when `redactedMarkdown` is `null` (`failed` or `too_large`), markdown is set to an empty string. `onlyMainContent` remains `true` by default but can be set to `false` explicitly. Stale/cached `pii` is dropped when not requested.
  - Billing: +4 credits when `redactPII` is on, plus +4 per extra PDF page (in addition to existing +1/page parse cost).
  - SDKs: JS and Python mirror the new `pii` types and `redactPII` options.

- **Migration**
  - Response shape: statuses collapsed to `ok | skipped | failed` with a `reason`; `truncatedAt` removed; spans gained `entity`/`source`; `counts` added. Update client parsing.
  - Behavior: `document.markdown` is now the redacted version (or empty on failure). `onlyMainContent` keeps its default `true` but can be set `false` with `redactPII`.
  - Env: `FIRE_PRIVACY_URL` is optional; when unset, redaction returns `failed/error` without an HTTP call. `FIRE_PRIVACY_TIMEOUT_MS` default 5000.
  - To enable: include `pii` in `formats` and set `redactPII`.

<sup>Written for commit 7c0af34a40fbed2b05fbe19ff9b6987c4d5cf711. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/firecrawl/pull/3525?utm_source=github">Review in cubic</a>


abimaelmartell:
Re: the open question — confirming `model_status` and `model_truncated_at` are stable contract fields, not internal-only. Documented them in the README in firecrawl/fire-privacy@e941b1e:

| Field | Type | Notes |
| --- | --- | --- |
| `model_status` | enum | `"ok"` / `"skipped"` / `"error"` / `"disabled"` |
| `model_truncated_at` | `int \| null` | Char index where the model's view of the input was cut, or `null` if it fit / wasn't run |

Behavior the client should assume:

- `model_status` is always present in 200 responses. `"error"` means the model call failed but Presidio spans (if any) are still returned — this is a *degraded* 200, not a clean one. Branch on it when model coverage matters.
- `model_truncated_at` is `null` whenever the input fit under 32K chars or the model didn't run. Heuristic spans always cover the full input regardless.

The tolerant defaults (`ok` / `null`) in the client are fine — they're correct for the only realistic missing-field case (a Presidio-only build where `model_status` legitimately defaults to "disabled" anyway).

Otherwise the PR shape looks right. Schema requiring `pii` in formats matches the precedent we discussed; fail-soft mapping (503 → `service_at_capacity`, 5xx/413 → `error`, timeout → `timeout`) is exactly the contract on the fire-privacy side.

abimaelmartell:
Pushed `5d0a2b6` — expanded `redactPII` from boolean-only to `boolean | object` form. Boolean form unchanged; the object form lets callers tune three things:

```typescript
redactPII?: boolean | {
  mode?: "accurate" | "aggressive" | "fast";
  entities?: ("PERSON" | "EMAIL" | "PHONE" | "LOCATION" | "FINANCIAL" | "SECRET")[];
  replaceStyle?: "tag" | "mask" | "remove";
};
```

**Mappings** (in `fire-privacy-client.ts`):

| public field | value | fire-privacy field | value |
|---|---|---|---|
| `mode` | `accurate` (default) | `mode` | `model` |
| | `aggressive` | | `both` |
| | `fast` | | `heuristics` |
| `replaceStyle` | `tag` (default) | `operator` | `replace` |
| | `mask` | | `mask` |
| | `remove` | | `redact` |
| `entities` | _allowlist_ | _filtered client-side_ | spans kept only if their kind maps to one of the allowed buckets |

**Entity filter behavior:**
- Each span kind (`PRIVATE_PERSON`, `EMAIL_ADDRESS`, `CREDIT_CARD`, ...) maps to one of six unified buckets via `KIND_TO_ENTITY` in the client.
- When `entities` is set, spans whose kind doesn't fall into any of the requested buckets are dropped.
- When filtering prunes spans, `redactedMarkdown` is re-rendered from the original text + remaining spans (respecting `replaceStyle`).
- When no filter is set, we use fire-privacy's `redacted_text` directly (it already used the operator we asked for).

**Zod transform:** `true` normalizes to a defaults object; `false`/unset → `undefined`. Downstream code keeps the same truthy check on `meta.options.redactPII`. `strictObject` rejects typos.

**Tests:** Added cases for each mode mapping, each replaceStyle mapping, entities filter (with re-render under `tag`/`mask`/`remove`), unmapped-kind drop, unknown mode rejection, strict-object enforcement. 37 tests in `redactPII-schema.test.ts` + `fire-privacy-client.test.ts`, all passing.

Previous "rejects redactPII passed as an object" e2e test flipped to "accepts redactPII as an options object", plus a new "rejects an unknown mode" case (`mode: "model"` is the fire-privacy internal name, not the public surface).

No SDK / CLI changes — those still defer per the original PR scope.

PR #3671, opened 2026-05-29
fix(api): reuse redis client in health checks

### LINKED ISSUE #3662: [Self-Host] nuq workers leak Redis connections until maxclients (10k) → continuous EPIPE storm, climbing CPU/memory
**Describe the Issue**

  On a self-hosted instance, Redis connections from the Node processes grow steadily over time and are never released. After ~18 days of uptime they reached Redis's default `maxclients` ceiling (10,000). Once the cap is hit, **every** Redis operation fails
  instantly with `[ioredis] Unhandled error event: Error: write EPIPE`, and the API logs this error in a tight loop (~71 MB of log output every 30 seconds). That runaway error loop — not scrape traffic — pegs CPU and never drops back to baseline, producing a
  steadily climbing CPU/memory "staircase" instead of the expected sawtooth.
  
  This looks distinct from the older BullMQ connection work in #1966 — the connections here are held by the new **nuq** workers (`nuq-worker.js`, `nuq-reconciler-worker.js`, `queue-worker.js`), not the legacy per-queue BullMQ path.

  **To Reproduce**
  
  1. Run a self-hosted instance under steady, long-running scrape load (in our case ~2,800 scrapes/day from a domain-monitoring workload, many against slow/parked/dead domains that time out).
  2. Leave it running for many days without restarting the `firecrawl` process.
  3. Watch Redis connection count climb monotonically:
     ss -tnp | grep -c ':6379'
  It rises by roughly ~550/day in our case and does not fall when scrape activity is idle.
  4. When it reaches `maxclients` (10,000), `redis-cli` returns `ERR max number of clients reached`, and the API log fills with `write EPIPE` errors continuously.
  
  **Expected Behavior**
  
  Redis connections should stay at a small, stable pool (reused across jobs) regardless of uptime or scrape volume, and should be released when jobs complete. A long-running instance should not require periodic restarts to avoid exhausting Redis client slots.

  **Observed evidence**

  - `ss -tnp | grep -c ':6379'` → ~10,000, **all owned by `node`** (confirms the leak is in the app, not OS/fd limits — `/proc/<redis>/limits` showed open-files 65535, well above the cap).
  - After `pm2 restart firecrawl`: connections dropped 10,000 → ~72 immediately and the EPIPE storm stopped.

  - `ss -tnp | grep -c ':6379'` → ~10,000, **all owned by `node`** (confirms the leak is in the app, not OS/fd limits — `/proc/<redis>/limits` showed open-files 65535, well above the cap).
  - After `pm2 restart firecrawl`: connections dropped 10,000 → ~72 immediately and the EPIPE storm stopped.
  - `ps aux --sort=-%mem` also showed a `chrome-headless-shell --type=renderer` pinned at 100% CPU with 78 min accumulated CPU time (orphaned renderer that was never reaped) — possibly related, possibly separate.

  **Environment**

  - Deployment Type: Self-hosted (PM2, not Docker)
  - Firecrawl Version: commit `82601ac19` (main, 2026-04-30)
  - Node.js Version: v22.22.2
  - OS: Ubuntu (Linux 6.8 kernel)
  - Architecture: nuq (PostgreSQL + RabbitMQ + Redis); 5× nuq-worker, 1× nuq-reconciler-worker, queue-worker, extract-worker
  - Redis: default `maxclients 10000`

  **Workaround**

  A scheduled daily `pm2 restart firecrawl` keeps the connection count well under the cap. Posting in case the nuq Redis connection lifecycle has a leak that should be fixed upstream — happy to provide more detail or test a patch.

RitwijParmar:
Fixes #3662

## Summary
- reuse the shared queue Redis client in `/admin/:key/redis-health` instead of opening a new ioredis connection on every health probe
- add a regression test around the health controller so future changes do not reintroduce per-request Redis clients
- close both NuQ queues and the NuQ semaphore Redis client in `nuqShutdown()` so worker restarts do not leave Redis clients behind

## Why
The self-hosted report shows Redis client count climbing until `maxclients`, then falling immediately after a process restart. I found one concrete leak path in the admin Redis health endpoint: every request constructed `new Redis(config.REDIS_URL!)` and never called `quit()`/`disconnect()`. Reusing `getRedisConnection()` keeps this to the existing process-level queue client instead of one connection per probe.

I also wired the NuQ Redis singleton into the existing shutdown path. That does not claim to explain the whole long-running leak by itself, but it makes Redis lifecycle explicit for the NuQ worker processes mentioned in the report.

## Tests
- `git diff --check`

Not run locally: Jest/TypeScript, because this checkout does not have `pnpm`, `npm`, `tsc`, `jest`, or `node_modules` available in the shell.

---
## Summary by cubic
Reuse the shared Redis client in the admin health check and tighten NuQ Redis shutdown to stop leaking connections that could push Redis toward `maxclients`, and make the Redis Lua script hash cache mutable to support runtime resets.

- **Bug Fixes**
  - Health endpoint now uses `getRedisConnection()` instead of creating a new `ioredis` client per probe.
  - NuQ shutdown now closes both queues and the NuQ Redis client via `nuqRedis.shutdown()` to prevent leftover clients after restarts.
  - Made the Redis Lua script hash cache mutable so hashes can be updated and cleared during shutdown.
  - Removed an unused test file.

<sup>Written for commit 5f50c037d5bf7dfbed85d3f63de0e3cfa771e626. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/firecrawl/pull/3671?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


RitwijParmar:
Pushed a small fix for the TypeScript build failure from CI. The script hash cache is now explicitly mutable, which matches how  loads hashes and  clears them. I could not run the API build locally because this shell has Node but no pnpm/npm installed.

RitwijParmar:
Small clarification to the note above. The mutable cache fix is for the Redis script hash map used by ensureRedis and shutdownRedis. CI was failing at TypeScript build time because the mapped type inherited readonly keys from luaScripts as const, even though the runtime cache is intentionally cleared and repopulated.

mogery:
thank you!

PR #3679, opened 2026-05-30
fix: fix double polling interval conversion in crawl --wait

### LINKED ISSUE #3373: firecrawl crawl --wait (without --progress) silently hangs ~83 minutes per poll due to double ms/sec conversion
## Summary

`firecrawl crawl --wait` (without `--progress`) does not return within any reasonable timeframe. The crawl runs server-side, credits are charged, but the CLI never writes output because it's sleeping on a poll interval that's been multiplied by 1000 twice — once in the CLI, once in the SDK. With the default `--poll-interval 5`, the CLI sleeps **~83 minutes between polls**.

`firecrawl crawl --wait --progress` is unaffected (it uses a separate polling loop).

**Severity:** High for `--wait` users without `--progress`. Credits are silently consumed with no output written.

**CLI version:** `firecrawl-cli@1.14.9`
**SDK version:** `@mendable/firecrawl-js@4.16.0` (bundled)
**Node:** v22.21.1
**OS:** macOS 15.3 (Darwin 25.3.0)

## Reproduction

\`\`\`bash
firecrawl crawl https://firecrawl.dev \\
  --limit 2 --wait --poll-interval 5 \\
  -o /tmp/crawl_test.json
\`\`\`

**Expected:** Job completes in seconds, file written to \`/tmp/crawl_test.json\`.

**Actual:** Process hangs. Server-side job completes within seconds and credits are charged (visible via \`firecrawl credit-usage\`), but the CLI keeps sleeping. No file is written, no stderr, no error.

## Root cause

\`dist/commands/crawl.js\` (CLI) lines 88–96 convert \`pollInterval\` and \`timeout\` from seconds to milliseconds before stuffing them into \`crawlOptions\`:

\`\`\`js
if (pollInterval !== undefined) {
    crawlOptions.pollInterval = pollInterval * 1000; // Convert to milliseconds
}
else {
    crawlOptions.pollInterval = 5000;
}
if (timeout !== undefined) {
    crawlOptions.timeout = timeout * 1000; // Convert to milliseconds
}
\`\`\`

It then calls \`app.crawl(url, crawlOptions)\` (line 135), which forwards to the SDK's \`crawl()\` function. The SDK's \`waitForCrawlCompletion()\` (\`@mendable/firecrawl-js/dist/index.js\` line 548) **expects \`pollInterval\` and \`timeout\` in seconds** and multiplies by 1000 itself:

\`\`\`js
// SDK index.js line 571
if (timeout != null && Date.now() - start > timeout * 1e3) { ... }
// SDK index.js line 574
await new Promise((r) => setTimeout(r, Math.max(1e3, pollInterval * 1e3)));
\`\`\`

So with \`--poll-interval 5\`:
- CLI: \`5 * 1000 = 5000\`
- SDK: \`5000 * 1000 = 5,000,000 ms = ~83 minutes per poll cycle\`

The first poll fires immediately (top of the \`while\` loop), sees \`status: 'scraping'\`, then sleeps 83 minutes before the second check. The user/process gives up long before that.

\`--wait --progress\` uses a separate polling loop in \`crawl.js\` (lines 99–131) that calls \`app.startCrawl()\` directly and polls in milliseconds, so it's unaffected.

## Patch (verified)

\`dist/commands/crawl.js\`:

\`\`\`diff
         if (wait) {
-            // Set polling options
+            // Set polling options (SDK waitForCrawlCompletion expects seconds,
+            // not milliseconds — it multiplies by 1000 internally).
             if (pollInterval !== undefined) {
-                crawlOptions.pollInterval = pollInterval * 1000; // Convert to milliseconds
+                crawlOptions.pollInterval = pollInterval; // seconds
             }
             else {
                 // Default poll interval: 5 seconds
-                crawlOptions.pollInterval = 5000;
+                crawlOptions.pollInterval = 5;
             }
             if (timeout !== undefined) {
-                crawlOptions.timeout = timeout * 1000; // Convert to milliseconds
+                crawlOptions.timeout = timeout; // seconds
             }
             // Show progress if requested - use custom polling for better UX
             if (options.progress) {
                 ...
-                const pollMs = crawlOptions.pollInterval || 5000;
+                const pollMs = (crawlOptions.pollInterval || 5) * 1000;
\`\`\`

The progress-mode change keeps that branch's internal ms math correct after the switch to seconds in \`crawlOptions\`.

## Verification after patch

\`\`\`bash
\$ firecrawl crawl https://firecrawl.dev --limit 2 --wait --poll-interval 3 -o /tmp/patch_test.json --timeout 90
# completes in 22 seconds
\$ ls -la /tmp/patch_test.json
-rw-r--r-- 110925 bytes
\$ jq -r .status /tmp/patch_test.json
completed
\$ jq -r '.data | length' /tmp/patch_test.json
2
# credits: 3,456 → 3,454 (exactly 2 pages, no waste)
\`\`\`

## Why this slipped through CI

Test suites and most users add \`--progress\` for visible feedback during \`--wait\`. The bare \`--wait\` code path (which delegates to the SDK's polling) is the blind spot. Recommend adding a CI test that runs \`firecrawl crawl <small-url> --limit 1 --wait\` (no \`--progress\`) with a tight timeout assertion.

## Impact / past incident

In a research session prior to discovering this bug, an unattended \`firecrawl crawl --wait --poll-interval 10\` consumed ~69 credits server-side with zero output and no error, surfaced only later via a credit-usage diff. The fix above prevents this entire class of silent failure.

glassgrass-art:
Fixes #3373. The pollInterval was being multiplied by 1000 twice, causing extremely long hangs. Changed to multiply only once.

---
## Summary by cubic
Fixes long hangs in `crawl --wait` by removing a double milliseconds conversion of `pollInterval` in `waitForCrawlCompletion`. Polling now respects the intended interval with a 1s minimum.

<sup>Written for commit 3a3ae5b0cf3c350ad78bdce4fe03d4f4db071c0a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/firecrawl/pull/3679?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/js-sdk/firecrawl/src/v2/methods/crawl.ts
+++ b/apps/js-sdk/firecrawl/src/v2/methods/crawl.ts
@@ -151,7 +151,7 @@ export async function waitForCrawlCompletion(http: HttpClient, jobId: string, po
       throw new JobTimeoutError(jobId, timeout, 'crawl');
     }
     
-    await new Promise((r) => setTimeout(r, Math.max(1000, pollInterval * 1000)));
+    await new Promise((r) => setTimeout(r, Math.max(1000, pollInterval)));
   }
 }
 


biswajeetdev (commented):
Straightforward fix — `pollInterval` is already in milliseconds when passed in, so `* 1000` was multiplying it into seconds-as-milliseconds, meaning a 2-second poll interval became a 33-minute wait. The `Math.max(1000, pollInterval)` form is correct.

One thing worth tracing: grep for where `pollInterval` is set upstream (the default and any CLI parsing that converts from seconds). If there is a seconds-to-ms conversion somewhere in the call chain that was intentionally paired with this `* 1000`, removing it here without removing that conversion could reverse the bug. A quick `grep -r "pollInterval"` across the SDK to confirm the value is always passed in as ms would close that loop.

PR #3699, opened 2026-06-03
Use configurable public host for self-hosted response URLs

### LINKED ISSUE #2964: [Self-Host] custom hostname
Is it possible to set up custom hostname? I am running self-hosted instance behind reverse proxy and firecrawl response URLs are always http://localhost:3002

saivedant169:
## Problem

Self-hosted instances behind a reverse proxy build absolute URLs from `req.get("host")`, which returns the internal bind address. Crawl/batch job URLs and pagination `next` links come back pointing at `http://localhost:3002` and are unreachable by the client.

Fixes #2964

## Fix

- Add `getRequestHostname(req)` in `apps/api/src/lib/host-utils.ts`.
- Use it at every site that builds a response URL: `v1/crawl`, `v1/batch-scrape`, `v1/crawl-status`, `v2/crawl`, `v2/batch-scrape`, `v2/crawl-status`, and `v2/monitor` checks.
- Resolution order: `SELF_HOSTED_DOMAIN` env, then the first `X-Forwarded-Host` value, then `req.get("host")`. Cloud behavior is unchanged when neither is set.
- `SELF_HOSTED_DOMAIN` is normalized to a bare `host[:port]` even if an operator includes a scheme or path, and an empty or whitespace value is ignored, so the helper never produces malformed URLs like `http://https://example.com`.
- The helper returns a host only and callers keep prefixing `req.protocol`, so pagination `next` links stay valid.

Scope is limited to the URL-emitting controllers plus the helper and its test. A prior attempt (#3015) mixed this fix with unrelated PDF, fire-engine, and SDK changes and was abandoned; this PR keeps the change focused and addresses the empty-value and malformed-URL edge cases that review flagged there.

## Test plan

New unit test `apps/api/src/lib/__tests__/host-utils.test.ts` covering the resolution order and normalization:

```
PASS src/lib/__tests__/host-utils.test.ts
  getRequestHostname
    ✓ falls back to the request host when nothing is configured
    ✓ prefers SELF_HOSTED_DOMAIN over the request host
    ✓ keeps an explicit port in SELF_HOSTED_DOMAIN
    ✓ strips a scheme and path from SELF_HOSTED_DOMAIN
    ✓ ignores an empty or whitespace SELF_HOSTED_DOMAIN
    ✓ uses the first X-Forwarded-Host when no domain is configured
    ✓ prefers SELF_HOSTED_DOMAIN over X-Forwarded-Host
Tests:       7 passed, 7 total
```

Also ran `tsc --noEmit`: no new type errors in the changed files. Documented `SELF_HOSTED_DOMAIN` in `apps/api/.env.example`.

---
## Summary by cubic
Fixes broken absolute URLs behind reverse proxies by using Express’s `req.host`. Crawl/batch job URLs, parse-upload links, and pagination now point to the external domain (including port) instead of the internal bind address.

- **Bug Fixes**
  - Replaced `req.get("host")` with `req.host` in v1/v2 crawl, batch-scrape, crawl-status, monitor checks, and v2 parse-upload; works with `trust proxy` and needs no extra config.
  - Honors `X-Forwarded-Host` and preserves ports; pagination `next` links remain valid.

<sup>Written for commit 5d79fea9b4e1bc7ccfb4ce0bc341f3f587173d31. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/firecrawl/pull/3699?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/api/src/controllers/v1/batch-scrape.ts
+++ b/apps/api/src/controllers/v1/batch-scrape.ts
@@ -255,7 +255,7 @@ export async function batchScrapeController(
   return res.status(200).json({
     success: true,
     id,
-    url: `${protocol}://${req.get("host")}/v1/batch/scrape/${id}`,
+    url: `${protocol}://${req.host}/v1/batch/scrape/${id}`,
     invalidURLs,
   });
 }
--- a/apps/api/src/controllers/v1/crawl-status.ts
+++ b/apps/api/src/controllers/v1/crawl-status.ts
@@ -299,7 +299,7 @@ export async function crawlStatusController(
     next:
       (outputBulkA.total ?? 0) > start + iteratedOver ||
       outputBulkA.status !== "completed"
-        ? `${req.protocol}://${req.get("host")}/v1/${isBatch ? "batch/scrape" : "crawl"}/${req.params.jobId}?skip=${start + iteratedOver}${req.query.limit ? `&limit=${req.query.limit}` : ""}`
+        ? `${req.protocol}://${req.host}/v1/${isBatch ? "batch/scrape" : "crawl"}/${req.params.jobId}?skip=${start + iteratedOver}${req.query.limit ? `&limit=${req.query.limit}` : ""}`
         : undefined,
   };
 
--- a/apps/api/src/controllers/v1/crawl.ts
+++ b/apps/api/src/controllers/v1/crawl.ts
@@ -202,6 +202,6 @@ export async function crawlController(
   return res.status(200).json({
     success: true,
     id,
-    url: `${protocol}://${req.get("host")}/v1/crawl/${id}`,
+    url: `${protocol}://${req.host}/v1/crawl/${id}`,
   });
 }
--- a/apps/api/src/controllers/v2/batch-scrape.ts
+++ b/apps/api/src/controllers/v2/batch-scrape.ts
@@ -285,7 +285,7 @@ export async function batchScrapeController(
   return res.status(200).json({
     success: true,
     id,
-    url: `${protocol}://${req.get("host")}/v2/batch/scrape/${id}`,
+    url: `${protocol}://${req.host}/v2/batch/scrape/${id}`,
     invalidURLs,
   });
 }
--- a/apps/api/src/controllers/v2/crawl-status.ts
+++ b/apps/api/src/controllers/v2/crawl-status.ts
@@ -326,7 +326,7 @@ export async function crawlStatusController(
     next:
       (outputBulkA.total ?? 0) > start + iteratedOver ||
       outputBulkA.status !== "completed"
-        ? `${req.protocol}://${req.get("host")}/v2/${isBatch ? "batch/scrape" : "crawl"}/${req.params.jobId}?skip=${start + iteratedOver}${req.query.limit ? `&limit=${req.query.limit}` : ""}`
+        ? `${req.protocol}://${req.host}/v2/${isBatch ? "batch/scrape" : "crawl"}/${req.params.jobId}?skip=${start + iteratedOver}${req.query.limit ? `&limit=${req.query.limit}` : ""}`
         : undefined,
   };
 
--- a/apps/api/src/controllers/v2/crawl.ts
+++ b/apps/api/src/controllers/v2/crawl.ts
@@ -261,7 +261,7 @@ export async function crawlController(
   return res.status(200).json({
     success: true,
     id,
-    url: `${protocol}://${req.get("host")}/v2/crawl/${id}`,
+    url: `${protocol}://${req.host}/v2/crawl/${id}`,
     ...(req.body.prompt && {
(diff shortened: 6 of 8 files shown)


saivedant169:
fixed in c9a0c62. normalizeHost now returns undefined when the value cannot be parsed into a host, and getRequestHostname falls through to X-Forwarded-Host then req host instead of leaking the raw value. Added a test covering an unparseable SELF_HOSTED_DOMAIN (8 passing now).

saivedant169:
Open since early June. The cubic review is addressed and it's mergeable, the only red check is the Vercel preview deploy that needs maintainer auth on fork PRs. @mogery mind giving it a look when you have time?

iwitw182-glitch:
Suer

mogery:
Not sure why this makes sense. X-Forwarded-Host is already obeyed [automatically](https://expressjs.com/en/5x/api/request/#reqhost) if the `EXPRESS_TRUST_PROXY` env is correctly configured: https://github.com/firecrawl/firecrawl/blob/8d679cbcb68ad8456f26166d69fb17d03c7068fe/apps/api/src/index.ts#L75-L77

saivedant169:
Agreed that trust proxy is the right signal to lean on. One wrinkle: this repo is on Express 4.22, where the trust-proxy-aware accessor is `req.hostname`, and that strips the port. The response URLs need `host:port` (self-hosted often runs on something like `:3002`), so `req.hostname` would drop it, while `req.get("host")` keeps the port but ignores `X-Forwarded-Host`, which is the bug here. The 5.x `req.host` you linked does keep the port, but we're not on 5 yet.

So leaning purely on the framework on 4.x means either losing the port or still ignoring `X-Forwarded-Host`. I can rework this to read `X-Forwarded-Host` directly when trust proxy is set (which keeps the port) and leave `SELF_HOSTED_DOMAIN` only as an explicit override for setups with no forwarding proxy. Or if you'd rather just use `req.hostname` and accept dropping non-default ports, I can do that. Which way do you prefer?

mogery:
> Agreed that trust proxy is the right signal to lean on. One wrinkle: this repo is on Express 4.22, where the trust-proxy-aware accessor is `req.hostname`, and that strips the port. The response URLs need `host:port` (self-hosted often runs on something like `:3002`), so `req.hostname` would drop it, while `req.get("host")` keeps the port but ignores `X-Forwarded-Host`, which is the bug here. The 5.x `req.host` you linked does keep the port, but we're not on 5 yet.
> 
> So leaning purely on the framework on 4.x means either losing the port or still ignoring `X-Forwarded-Host`. I can rework this to read `X-Forwarded-Host` directly when trust proxy is set (which keeps the port) and leave `SELF_HOSTED_DOMAIN` only as an explicit override for setups with no forwarding proxy. Or if you'd rather just use `req.hostname` and accept dropping non-default ports, I can do that. Which way do you prefer?

Gah, good catch, that's annoying. I'll port to Express 5.

mogery:
> > Agreed that trust proxy is the right signal to lean on. One wrinkle: this repo is on Express 4.22, where the trust-proxy-aware accessor is `req.hostname`, and that strips the port. The response URLs need `host:port` (self-hosted often runs on something like `:3002`), so `req.hostname` would drop it, while `req.get("host")` keeps the port but ignores `X-Forwarded-Host`, which is the bug here. The 5.x `req.host` you linked does keep the port, but we're not on 5 yet.
> > So leaning purely on the framework on 4.x means either losing the port or still ignoring `X-Forwarded-Host`. I can rework this to read `X-Forwarded-Host` directly when trust proxy is set (which keeps the port) and leave `SELF_HOSTED_DOMAIN` only as an explicit override for setups with no forwarding proxy. Or if you'd rather just use `req.hostname` and accept dropping non-default ports, I can do that. Which way do you prefer?
> 
> Gah, good catch, that's annoying. I'll port to Express 5.

Ok, we are now on 5! #3860

@saivedant169 Could you please re-test whether it works now on latest, and if not, make the necessary updates?

saivedant169:
@mogery reworked on latest main. Dropped the SELF_HOSTED_DOMAIN env var and the helper, it just uses req.host now across the crawl/batch/monitor/parse URL builders. On 5.x req.host honors X-Forwarded-Host with trust proxy and keeps the port, so the public host comes through with no extra config. I also caught one more spot still on req.get("host") (parse-upload) and switched that too. Typechecks clean against @types/express 5.

mogery (approved):

mogery:
Thanks!

PR #3730, opened 2026-06-07
fix(api): reject self-hosted scrape interact clearly

### LINKED ISSUE #3570: [Self-Host] /v2/scrape/{jobId}/interact requires Supabase
**Describe the Issue**
Attempting to call `/v2/scrape/{jobId}/interact` fails as the endpoint’s controller tries to resolve the scrape job through Supabase. [Since self-host Supabase is not supported](https://github.com/firecrawl/firecrawl/issues/1797), the endpoint fails locally.

I require this feature and was intending to self-host. If there are no plans to fix, could the documentation reflect that the `/interact` route is self-host only? Otherwise, I'd appreciate this functionality greatly.

**To Reproduce**
Steps to reproduce the issue:
1. Configure a self-hosted Firecrawl server
2. Call /v2/scrape succesfully
3. Call /interact
4. Receive the error `Supabase RR client is not configured.","stack":"Error: Supabase RR client is not
configured.\n    at Proxy.<anonymous> (/app/dist/src/services/supabase.js:78:23)\n `

**Expected Behavior**
Supabase support, or documentation clarifying that `/interact` is not available when self-hosting.

he-yufeng:
## Summary
- return a clear 501 when `/v2/scrape/:jobId/interact` is used with database authentication disabled
- avoid falling through to the Supabase scrape lookup in self-hosted deployments
- add a controller regression test for the self-hosted path

Fixes #3570.

## To verify
- `pnpm exec jest src/controllers/v2/__tests__/scrape-browser.test.ts --runInBand`
- `git diff --check`

I also ran `pnpm exec tsc --noEmit --pretty false`; it still fails before this change on the local native package setup because `@mendable/firecrawl-rs` types are not generated when dependencies are installed with scripts disabled.

---
## Summary by cubic
Return a 501 from `/v2/scrape/:jobId/interact` when database authentication is disabled, clearly signaling that interact is unsupported in self-hosted mode and skipping any Supabase lookup. Adds a `vitest` regression test to assert the early return and that no Supabase query is made.

<sup>Written for commit f6f184a13a2f0d0af228e0b30294a363a9e23244. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/firecrawl/pull/3730?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/api/src/controllers/v2/__tests__/scrape-browser.test.ts
+++ b/apps/api/src/controllers/v2/__tests__/scrape-browser.test.ts
@@ -1,0 +1,103 @@
+import type { Response } from "express";
+import { config } from "../../../config";
+import { supabaseGetScrapeById } from "../../../lib/supabase-jobs";
+import { scrapeInteractController } from "../scrape-browser";
+import type { RequestWithAuth } from "../types";
+
+jest.mock("uuid", () => ({
+  v7: jest.fn(() => "session-123"),
+}));
+
+jest.mock("../../../lib/supabase-jobs", () => ({
+  supabaseGetScrapeById: jest.fn(),
+}));
+
+jest.mock("../../../lib/browser-sessions", () => ({
+  insertBrowserSession: jest.fn(),
+  getBrowserSession: jest.fn(),
+  updateBrowserSessionActivity: jest.fn(() => Promise.resolve()),
+  updateBrowserSessionCreditsUsed: jest.fn(() => Promise.resolve()),
+  updateBrowserSessionScrapeId: jest.fn(() => Promise.resolve()),
+  claimBrowserSessionDestroyed: jest.fn(),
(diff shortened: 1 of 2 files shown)


he-yufeng:
Quick status note: cubic is green, and the only failing status I see is Vercel pointing to the GitHub authorization flow for the Firecrawl team. I don't have access to authorize that integration from the fork side; please rerun/authorize it or ignore that external check if it is not required for this API validation change.

mogery:
Thank you!

PR #3734, opened 2026-06-08
fix(api): fail open when rate-limit Redis is unreachable (#3728)

### LINKED ISSUE #3728: [Bug] Rate limiter fails closed on rate-limit Redis outage — every request gets a 429 with "undefined" fields
## Describe the Bug

The sole production caller of `rate-limiter-flexible`'s `consume()` is `apps/api/src/controllers/auth.ts:634`. Its `catch` block assumes the rejection is always a `RateLimiterRes`:

```ts
try {
  await rateLimiter.consume(team_endpoint_token);
} catch (rateLimiterRes) {
  const secs = Math.round(rateLimiterRes.msBeforeNext / 1000) || 1;
  // ... returns { status: 429, error: `... Consumed (req/min): ${rateLimiterRes.consumedPoints} ...` }
}
```

But `RateLimiterRedis` (v2.4.2) only rejects with a `RateLimiterRes` when the limit is exceeded. When the **rate-limit Redis store errors** (`REDIS_RATE_LIMIT_URL` unreachable), `consume()` rejects with a raw `Error` — `RateLimiterStoreAbstract._handleError` does `reject(err)` because no `insuranceLimiter` is configured in `apps/api/src/services/rate-limiter.ts`. The catch then reads `.msBeforeNext` / `.consumedPoints` / `.remainingPoints` (all `undefined`) and **unconditionally returns HTTP 429**.

Net effect: a transient rate-limit-Redis blip is reported to **every authenticated request** as a bogus quota error — fail-closed, mislabeled. Each request also blocks ~30s first (ioredis offline queue + 20 reconnect retries) before returning the wrong 429.

## To Reproduce

1. Run the stack with `USE_DB_AUTHENTICATION=true` and a dedicated `REDIS_RATE_LIMIT_URL` separate from `REDIS_URL`.
2. Authenticate normally → request succeeds (not 429).
3. Take down the rate-limit Redis.
4. Send any authenticated request → after ~30s, HTTP 429:

```
HTTP/1.1 429 Too Many Requests
X-Response-Time: 29894.796ms
{"success":false,"error":"Rate limit exceeded. Consumed (req/min): undefined, Remaining (req/min): undefined. Upgrade your plan at https://firecrawl.dev/pricing for increased rate limits or please retry after 1s, resets at Invalid Date"}
```

This can also be reproduced deterministically with a unit test: point a `RateLimiterRedis` (configured exactly like `services/rate-limiter.ts` — no `insuranceLimiter`) at a dead store and call `consume()`. It rejects with a plain `Error` (not a `RateLimiterRes`), and replaying the `auth.ts` formatting yields `undefined` fields + `status: 429`.

## Expected Behavior

On a rate-limiter **store error** (as opposed to an actual limit hit), the API should fail **open** (allow the request, log/alert) or degrade via an in-memory `insuranceLimiter` — not return 429 to every caller with `undefined` / `Invalid Date` fields.

## Environment

- Deployment Type: Self-hosted (the same code path also runs on Cloud)
- Firecrawl Version: `main`
- Node.js Version: 24.x
- `rate-limiter-flexible`: 2.4.2

## Logs

```
HTTP/1.1 429 Too Many Requests
X-Response-Time: 29894.796ms
{"success":false,"error":"Rate limit exceeded. Consumed (req/min): undefined, Remaining (req/min): undefined. ... please retry after 1s, resets at Invalid Date"}
real    0m29.915s
```

The `undefined` consumed/remaining values, `Invalid Date`, and `retry after 1s` (`Math.round(NaN) || 1`) are the signature of an infrastructure error being formatted as a quota response.

## Additional Context

- Root cause: `apps/api/src/controllers/auth.ts:633-659` assumes the rejection is a `RateLimiterRes`; `apps/api/src/services/rate-limiter.ts` configures no `insuranceLimiter`/fallback.
- Suggested fix: branch on `rateLimiterRes instanceof RateLimiterRes` (format the 429 only for real limit hits; otherwise fail open + log/alert), and/or configure an in-memory `insuranceLimiter` so a store outage degrades gracefully.
- Related robustness gap observed during the repro: the **worker** processes' rate-limit Redis client has no `error` handler, so a sustained outage throws an unhandled `MaxRetriesPerRequestError` and crashes workers (`index-worker` exits 1). Worth handling alongside this fix.

I'm happy to open a PR for this.

eldar702:
## Problem

Fixes #3728.

`apps/api/src/controllers/auth.ts` is the sole production caller of `rateLimiter.consume()`. Its `catch` block assumed **every** rejection was a `RateLimiterRes` and unconditionally read `.msBeforeNext` / `.consumedPoints` / `.remainingPoints`:

```ts
const secs = Math.round(rateLimiterRes.msBeforeNext / 1000) || 1;
const retryDate = new Date(Date.now() + rateLimiterRes.msBeforeNext);
return {
  success: false,
  error: `Rate limit exceeded. Consumed (req/min): ${rateLimiterRes.consumedPoints}, ...`,
  status: 429,
};
```

But `rate-limiter-flexible` only rejects with a `RateLimiterRes` on a **genuine** limit hit. When the rate-limit Redis store is unreachable — and `apps/api/src/services/rate-limiter.ts` configures `RateLimiterRedis` with **no `insuranceLimiter`** — `consume()` rejects with a plain `Error` instead.

The old code then returned **HTTP 429** with `Consumed (req/min): undefined`, `Remaining (req/min): undefined`, and `resets at Invalid Date` on **every authenticated request** while the rate-limit Redis was down (i.e. the limiter fails *closed*).

## Fix (file)

`apps/api/src/controllers/auth.ts` — discriminate the rejection in the `catch` block:

- **Genuine `RateLimiterRes`** → return the existing 429 with populated fields (unchanged behavior, identical message).
- **Store / infra `Error`** → log it and **fail open** (fall through to the normal success path) instead of emitting a mislabeled 429.

The discrimination + message-building logic lives in a small, side-effect-free helper `apps/api/src/lib/rate-limit-error.ts` (`isRateLimiterRes` + `buildRateLimitMessage`) so it is unit-testable without booting the DB/Redis module graph. The guard is a structural check on `msBeforeNext` (resilient to module-interop `instanceof` edge cases), exactly the discriminator the issue author suggested.

## Test (RED-first)

`apps/api/src/lib/rate-limit-error.test.ts`:

1. **RED** (proven against the original catch logic before the fix): a plain backing-store `Error` produced a `status: 429` whose message contained `Consumed (req/min): undefined`, `Remaining (req/min): undefined`, and `Invalid Date`.
2. **GREEN** after the fix:
   - A genuine `RateLimiterRes` is still classified as a real hit and yields a populated message (no `undefined`).
   - A backing-store `Error` is **not** classified as a rate-limit hit → caller fails open (no 429).
   - `undefined` / `null` / non-numeric-`msBeforeNext` rejections are also treated as fail-open.

Hermetic — no live Redis/DB required, CI-safe.

## Verification

```
$ npx tsc --noEmit -p tsconfig.json     # 0 errors across apps/api
$ npx jest src/lib/rate-limit-error.test.ts
PASS src/lib/rate-limit-error.test.ts
  rate-limit-error discrimination (#3728)
    ✓ treats a genuine RateLimiterRes as a real rate-limit hit
    ✓ does NOT treat a backing-store Error as a rate-limit hit (fail open)
    ✓ does NOT treat undefined/null rejections as a rate-limit hit
    ✓ does NOT treat an object with non-numeric msBeforeNext as a hit
Tests: 4 passed, 4 total
```

`prettier` + `knip` (the repo pre-commit gate) both pass. Diff: 3 files, +137 / -20, no new dependencies.

---

> AGPL-3.0: this contribution is offered under the same license as the project.

> 🤖 AI-assistance: drafted with Claude (Opus 4.8), reviewed for correctness.

---
## Summary by cubic
Fail open in the auth rate limiter when the Redis backend is unreachable to prevent false 429s with undefined fields. Genuine limit hits still return 429 with a clear, populated message. Fixes #3728.

- **Bug Fixes**
  - In `apps/api/src/controllers/auth.ts`, differentiate `rateLimiter.consume()` rejections between real `RateLimiterRes` and infra `Error` from `rate-limiter-flexible` (no `insuranceLimiter`).
  - Infra errors: log and allow the request (no 429). Real hits: return 429 using `buildRateLimitMessage`.
  - Added `apps/api/src/lib/rate-limit-error.ts` (`isRateLimiterRes`, `buildRateLimitMessage`) and tests (`rate-limit-error.test.ts`).

<sup>Written for commit 844fb4c4004d633c6d7e03d4a9b43aedbf1bbdb7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/firecrawl/pull/3734?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/api/src/controllers/auth.ts
+++ b/apps/api/src/controllers/auth.ts
@@ -7,6 +7,10 @@ import { withAuth } from "../lib/withAuth";
 import { getAgentSponsorStatus } from "../services/agent-sponsor";
 import { getRateLimiter, getAutumnRateLimiter } from "../services/rate-limiter";
 import {
+  buildRateLimitMessage,
+  isRateLimiterRes,
+} from "../lib/rate-limit-error";
+import {
   KEYLESS_FREE_TIER_LIMIT_MESSAGE,
   consumeKeylessRequest,
   isKeylessConfigured,
@@ -810,27 +814,39 @@ async function supaAuthenticateUser(
   try {
     await rateLimiter.consume(team_endpoint_token);
   } catch (rateLimiterRes) {
-    logger.error(`Rate limit exceeded: ${rateLimiterRes}`, {
-      teamId,
-      mode,
-      rateLimiterRes,
-    });
+    // `consume()` rejects with a `RateLimiterRes` on a genuine rate-limit hit,
+    // but with a plain `Error` when the rate-limit Redis store is unreachable
+    // (no insuranceLimiter is configured). Only the former is a real 429 —
+    // treating a store/infra error as a 429 returns a bogus response with
+    // `undefined` fields on every authenticated request while Redis is down
+    // (issue #3728). Fail open on infra errors instead.
+    if (!isRateLimiterRes(rateLimiterRes)) {
+      logger.error("Rate limiter backing store error — failing open", {
+        teamId,
+        mode,
+        error: rateLimiterRes,
+      });
+    } else {
+      logger.error(`Rate limit exceeded: ${rateLimiterRes}`, {
+        teamId,
+        mode,
+        rateLimiterRes,
+      });
+      const { message } = buildRateLimitMessage(rateLimiterRes);
 
-    const secs = Math.round(rateLimiterRes.msBeforeNext / 1000) || 1;
-    const retryDate = new Date(Date.now() + rateLimiterRes.msBeforeNext);
+      // We can only send a rate limit email every 7 days, send notification already has the date in between checking
+      // const startDate = new Date();
+      // const endDate = new Date();
+      // endDate.setDate(endDate.getDate() + 7);
 
-    // We can only send a rate limit email every 7 days, send notification already has the date in between checking
-    // const startDate = new Date();
-    // const endDate = new Date();
-    // endDate.setDate(endDate.getDate() + 7);
+      // await sendNotification(team_id, NotificationType.RATE_LIMIT_REACHED, startDate.toISOString(), endDate.toISOString());
 
-    // await sendNotification(team_id, NotificationType.RATE_LIMIT_REACHED, startDate.toISOString(), endDate.toISOString());
-
-    return {
-      success: false,
-      error: `Rate limit exceeded. Consumed (req/min): ${rateLimiterRes.consumedPoints}, Remaining (req/min): ${rateLimiterRes.remainingPoints}. Upgrade your plan at https://firecrawl.dev/pricing for increased rate limits or please retry after ${secs}s, resets at ${retryDate}`,
-      status: 429,
-    };
+      return {
+        success: false,
+        error: message,
+        status: 429,
+      };
+    }
   }
 
   if (
--- a/apps/api/src/lib/rate-limit-error.test.ts
+++ b/apps/api/src/lib/rate-limit-error.test.ts
@@ -1,0 +1,52 @@
+import { RateLimiterRes } from "rate-limiter-flexible";
+import { isRateLimiterRes, buildRateLimitMessage } from "./rate-limit-error";
+
+/**
+ * Regression tests for issue #3728.
+ *
+ * The auth controller's rate-limit `catch` block used to assume every
+ * rejection from `rateLimiter.consume()` was a `RateLimiterRes` and read
+ * `.msBeforeNext` / `.consumedPoints` / `.remainingPoints` off it. When the
+ * rate-limit Redis store is unreachable — and with no `insuranceLimiter`
+ * configured (see `services/rate-limiter.ts`) — `consume()` rejects with a
+ * plain `Error` instead. That produced a bogus HTTP 429 with `undefined`
+ * fields on *every* authenticated request while the store was down.
+ *
+ * The fix discriminates the rejection: genuine `RateLimiterRes` -> 429 with
+ * populated fields (unchanged); store/infra `Error` -> fail open.
+ */
+describe("rate-limit-error discrimination (#3728)", () => {
+  it("treats a genuine RateLimiterRes as a real rate-limit hit", () => {
+    // RateLimiterRes(remainingPoints, msBeforeNext, consumedPoints, isFirstInDuration)
+    const res = new RateLimiterRes(0, 30000, 100, false);
+
+    expect(isRateLimiterRes(res)).toBe(true);
+
+    const { secs, message } = buildRateLimitMessage(res);
+    expect(secs).toBe(30);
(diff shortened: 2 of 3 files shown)


eldar702:
Hi — just checking in on this one. The cubic bot approved with "No issues found" and all code checks are green. The only non-green item is the Vercel deploy-preview authorization gate, which requires a team member to authorize — not something I can resolve from my side. Would appreciate a merge or a heads-up if there is anything else blocking. Thanks!

eldar702:
Friendly nudge on this one 🙂

cubic approved ("No issues found") and all code checks are green — the only non-green item is the **Vercel deploy-preview**, which is gated on maintainer authorization for fork PRs, so I can't trigger it from here. The change is small and self-contained: fail **open** (allow the request) when the rate-limit Redis is unreachable, so a Redis blip can't take down the API (#3728).

Happy to rebase or adjust anything — would appreciate a review/merge when you have a moment. Thanks!

eldar702:
Re-ported onto latest `main` — the branch had fallen behind again and the PR was showing CONFLICTING/DIRTY. The conflict was a small textual overlap in `apps/api/src/controllers/auth.ts` (an active `logger.error` call had been added upstream in the same catch block since my last rebase); folded it into the correct branch of the fail-open fix. Also had to drop a stale `priceId` field from a log payload that #4021 removed from ACUC upstream — caught by `tsc --noEmit`, not by review, so worth flagging.

Verified: all 4 unit tests in `rate-limit-error.test.ts` pass, and `tsc --noEmit` is clean for both touched files.

eldar702:
Noticed the new `Superagent Security Scan` check came back `action_required` ("1 security concern(s) detected") after the rebase push. As a fork PR I don't have visibility into the specific finding (no annotations or detail exposed to me). I re-audited the touched diff myself — `apps/api/src/controllers/auth.ts`'s catch-block change and the new `apps/api/src/lib/rate-limit-error.ts` helper — and found nothing that should trigger a legitimate concern: no secrets, no dynamic code execution, no user input reaching a template/injection surface (just numeric fields formatted into a plain string, matching the existing message-building pattern this file already used). Flagging in case a maintainer can see the actual finding and it's something real I'm missing.

PR #3765, opened 2026-06-11
fix(api): shrink index DB pool to stay under pgbouncer max_client_conn

mogery:
## What

Tunes the Postgres pool sizing in `apps/api/src/db/connection.ts` so the `firecrawl-index` pool stops exhausting Supabase's pgbouncer client-connection limit.

- `makeDb` now accepts per-pool `max`/`min` sizing.
- Default `min` drops **2 → 0** so idle pods release connections instead of permanently holding 2 per pool, per process, across the whole fleet.
- The index pool gets its own smaller budget: **`max: 6, min: 0`** (was sharing the generic `max: 20, min: 2`).

## Why

The 2026-06-11 **"Elevated Index failures"** HIGH-escalation incident ([#976830263](https://uptime.betterstack.com/team/t116723/incidents/976830263)) was a 3-minute spike of 470 `index-scrape-fails`. Pulling the GCP logs for the window, **448 of 452** failures were:

```
08P01 FATAL: no more connections allowed (max_client_conn)
  at execRows (src/db/rpc.js)
  at scrapeURLWithIndex (engines/index/index.js:223)   // index_get_recent_4(...)
```

That's **pgbouncer** rejecting connections (not Postgres `max_connections` — that would be `53300`). The remaining 4 were `ETIMEDOUT` to the same pooler endpoint (`:6543`). 41 teams were affected, so it was pool exhaustion under a traffic burst, not one caller.

Supabase **hardcodes pgbouncer `max_client_conn` to 12000**, so we can't raise the ceiling — the only lever is opening fewer client connections per process. With the old `max: 20` across pools, the fleet-wide budget (`pods × Σ max`) sat dangerously close to 12000 at steady state, and a normal burst tipped it over. The index pool runs against the **transaction** pooler, which multiplexes server connections, so a large per-process client pool buys little throughput while eating the global cap.

## Notes

- No E2E test added — this is connection-pool sizing, not request-path behavior; the win condition is fewer client connections to pgbouncer under load, observable in prod metrics rather than in `snips`.
- Pre-existing `knip` warnings in `src/services/posthog.ts` (unused exports from the recent PostHog merge) are unrelated to this change.

---
## Summary by cubic
Shrink the index Postgres client pool and release idle connections to avoid exhausting pgbouncer `max_client_conn` during bursts. This caps per-process clients and mitigates recent index failures.

- **Bug Fixes**
  - `makeDb` now supports per-pool `max`/`min` sizing.
  - Default `min` lowered from 2 to 0 so idle pods release connections.
  - Index pool set to `max: 6, min: 0` (was `20/2`).

<sup>Written for commit c02f342895c2b4f612a393aa910118aee6dec83a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/firecrawl/pull/3765?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/api/src/db/connection.ts
+++ b/apps/api/src/db/connection.ts
@@ -5,9 +5,17 @@ import { logger } from "../lib/logger";
 
 type DB = NodePgDatabase;
 
+interface PoolSizing {
+  /** Max client connections this pool opens to the pooler. Default 20. */
+  max?: number;
+  /** Idle connections held open permanently. Default 0. */
+  min?: number;
+}
+
 function makeDb(
   connectionString: string | undefined,
   applicationName: string,
+  sizing: PoolSizing = {},
 ): DB | null {
   if (!connectionString) {
     return null;
@@ -16,8 +24,14 @@ function makeDb(
   const pool = new Pool({
     connectionString,
     application_name: applicationName,
-    max: 20,
-    min: 2,
+    // Each process opens up to `max` client connections per pool against the
+    // (transaction) pooler. Supabase pins pgbouncer's max_client_conn at 12000,
+    // so the fleet-wide budget is `pods * sum(max across pools)`. Keep these
+    // small — the transaction pooler multiplexes server connections, so a large
+    // per-process client pool buys little throughput but eats the global cap.
+    // `min: 0` lets idle pods release connections instead of holding them.
+    max: sizing.max ?? 20,
+    min: sizing.min ?? 0,
     keepAlive: true,
   });
   pool.on("error", err =>
@@ -62,7 +76,14 @@ const replicaDb = useDbAuthentication
       "firecrawl-api-rr",
     )
   : null;
-const indexDb = makeDb(config.INDEX_DATABASE_URL, "firecrawl-index");
+// The index pool was the sole consumer behind the 2026-06-11 pgbouncer
+// `08P01: no more connections allowed (max_client_conn)` incident. It runs
+// against the transaction pooler, so cap it well below the generic 20 to keep
+// the fleet-wide client-connection count under Supabase's 12000 ceiling.
+const indexDb = makeDb(config.INDEX_DATABASE_URL, "firecrawl-index", {
+  max: 6,
+  min: 0,
+});
 
 if (useDbAuthentication && !mainDb) {
   logger.error(


mogery:
### Follow-up from the GCP logs — confirms the bottleneck and the tradeoff

Pulled the local-side warnings for the incident window (15:00–15:30 UTC). The `pool.on("acquire")` exhaustion warning from `connection.ts` fired 300+ times, and it was overwhelmingly the index pool:

| Pool | warnings | peak `waiting` | state |
|------|---------:|---------------:|-------|
| `firecrawl-index` | 294 | **29** | `total: 20, idle: 0` |
| `firecrawl-api` | 6 | 8 | `total: 20, idle: 0` |

So the index path was hitting the wall on **both** sides simultaneously:
1. **Local** — each process saturated its 20-connection index pool and was queueing up to 29 deep.
2. **Global** — new connects to pgbouncer were failing with `08P01 max_client_conn` at the same time.

**Tradeoff this PR makes:** dropping the index pool to `max: 6` relieves the *global* pgbouncer cap but deepens *local* queueing per process. That's acceptable because index-engine failures are **not fatal** — at [`scrapeURL/index.ts:809`](https://github.com/firecrawl/firecrawl/blob/mog/index-pool-sizing/apps/api/src/scraper/scrapeURL/index.ts#L809) an `EngineError` is logged `warn` and the scrape **waterfalls to the next engine**. Worst case of a starved index pool is a cache miss → real scrape (added latency), not a failed request. That's why erring small is safe here.

**Caveat on the exact number:** `max: 6` is a conservative guess — I don't have the prod `pods × peak per-pod` figure to size it precisely. If the team wants to dial it in, a per-pool `pool.totalCount` gauge would give the real fleet-wide ceiling to tune against (happy to add as a follow-up). The number is trivially adjustable; the structural fix (per-pool sizing + `min: 0`) is the important part.

PR #3766, opened 2026-06-11
feat(monitor): search monitor

developersdigest:
## Summary by cubic
Adds a new inline `search` monitor target with goal-scoped dedupe and flat, deterministic billing; all search and diff-change judging now run through a shared LLM provider that prefers Vertex and falls back to Gemini, with Vertex billing labels attached for traceability.

- **New Features**
  - Target: `queries`, `searchWindow`, `includeDomains`, `excludeDomains` (exclude wins), `maxResults`; `alertMode` defaults to `first_match`; effective `depth` derives from `judgeEnabled` (`raw` when off, `deep` when on). Client `depth`/`alertMode`/`recheckAfter` are stripped; `scrapeOptions` is ignored. Signature includes domain scopes plus `maxResults` and effective `depth`.
  - Pipeline: inline SERP fetch; goal-versioned URL/event dedupe (goal + monitor name + queries); persist pages with `matchedQueries` and mirror search status to page status; stable event keys; cadence re-judging; `material_dev` re-alerts; `judgeDegraded` for judge-path failures; completion waits for `searchCompleted` so billing never reads as zero; emits `monitor.page` webhooks; deep-mode scrapes go through the shared scrape pipeline (queued, not per-page billed).
  - Judgment & LLMs: strict subject/topic matching with hard exclusions; snippet judge + router/skeptic/reviewer/event resolver/run summarizer; “raw” returns SERP hits without LLMs and does not attach a judgment object. All LLM calls use `getModel()` (prefer Vertex, fallback Gemini) and include Vertex billing labels (functionId, teamId, monitorId, monitorCheckId); diff-change judging now does the same.
  - Billing & Tracking: flat per check — search `ceil(totalResults/10)*2`, judge `1 * resultsJudged` (`raw` = 0). Actuals recorded on the check as `flatSearchTargetCredits`; search-target pages carry 0 per-page credits. LLM token usage is recorded for observability only. Estimates include the search call and a per-result judge allowance. Target signatures include `includeDomains`/`excludeDomains`, `maxResults`, and effective `depth`.

- **Migration**
  - A `search` target requires a non-empty `goal` only when judging is enabled; partial updates validate the merged state.
  - `depth`, `alertMode`, and `recheckAfter` are ignored on create/update; stored targets with them still work. `scrapeOptions` on `search` is ignored and stripped.
  - No DB migration; event state is stored in page metadata JSON.
  - SDKs (JS, Python, Go, Java, .NET, PHP, Elixir, Ruby, Rust) add `search` target types and `MonitorSearchTargetResult`; clients send/receive camelCase fields.

<sup>Written for commit 2079301e52636ab7cd8d959c0e28d33095c12250. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/firecrawl/pull/3766?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


apps/api/src/services/monitoring/search/run.ts
```diff
@@ -0,0 +1,899 @@
...
+      });
+      const scrapeCostTracking = new CostTracking();
+      try {
+        res = await scrapeURL(
+          "search-monitor;" + uuidv7(),
+          c.url,
+          scrapeParsedOptions,
+          { teamId, zeroDataRetention: params.zeroDataRetention },
+          scrapeCostTracking,
+        );
```
mogery: slightly worried about the fact that this overrides the team's concurrency queue

apps/api/src/services/monitoring/search/llm.ts
```diff
@@ -0,0 +1,353 @@
...
+};
+
+const eventResolverSchema = z.object({
+  matchedKey: z.string().nullable(),
+  isNew: z.boolean(),
+  label: z.string(),
+  reason: z.string(),
+});
+
+export async function resolveEvent(params: {
```
mogery, line 38: For every single LLM call, we should be using [Google Vertex Billing Labels](https://ai-sdk.dev/providers/ai-sdk-providers/google-vertex#optional-provider-settings:~:text=labels%20object,for%20usage%20details.), like we do in the old Extract code, so we can trace usage at the billing level by function name, team id, monitor id, monitor check id, etc.

apps/api/src/services/monitoring/search/tuning.ts
```diff
@@ -0,0 +1,25 @@
+import { createGoogleGenerativeAI } from "@ai-sdk/google";
```
mogery: We should prioritize using Vertex, and use the GenAI API as a fallback.

mogery (commented):
looks good for the most part. check the comments + another one that applies globally:
- PRs must add support of new functionality to all SDKs, not just JS and Python
also verify SDK versions are bumped before merging

apps/api/src/services/monitoring/search/run.ts
mogery: also not sure if you mean to bill credits for this internal scrape, but with these options, it will

apps/api/src/services/monitoring/search/tuning.ts
```diff
@@ -0,0 +1,53 @@
...
+                teamId: labels.teamId,
+                monitorId: labels.monitorId,
+                monitorCheckId: labels.monitorCheckId,
+              },
+            }
+          : {}),
+      },
+    },
+  };
+}
```
mogery: still incorrect for vertex, it needs a different key for provideroptions i believe
```
   vertex: {
      labels: ...
    } satisfies GoogleLanguageModelOptions,
```
https://ai-sdk.dev/providers/ai-sdk-providers/google-vertex

PR #2244, opened 2025-10-01
fix: image search field mapping in Python SDK

nehaaprasad:
### PR Description

- Fixed a bug where image search results weren't returning `image_url`, `image_width`, and `image_height` information. The issue was that the API returns these fields in camelCase format (`imageUrl`, `imageWidth`, `imageHeight`) but the Python SDK expects snake_case (`image_url`, `image_width`, `image_height`).

fixes: #2200 

**What was changed:**
- Added field mapping function in `normalize.py` to convert camelCase to snake_case for search results
- Updated search transformation logic to apply field normalization
- This affects image search results and also improves news search results that use similar image fields

**Files modified:**
- `apps/python-sdk/firecrawl/v2/utils/normalize.py` - Added `_map_search_result_keys()` function
- `apps/python-sdk/firecrawl/v2/methods/search.py` - Applied field normalization in `_transform_array()`

Issue #2200, opened 2025-09-24
[Bug] Image search not returning image_url or image size information


JacobBruce:
**Describe the Bug**

When using the Python SDK to perform an image search it is not returning the image URL or the image dimensions.  The title, url, and position attributes seem to be set properly, but the image_url, image_width, and image_height are all still set to None.

**To Reproduce**

Steps to reproduce the issue:

This code will crash when it tries to print the image URL since result.image_url isn't a string.
```
from firecrawl import Firecrawl

firecrawl = Firecrawl(api_key="MY-API-KEY")

results = firecrawl.search(
	query = "test",
	limit = 2,
	sources = ["images"]
)

for result in results.images:
	print("Title: "+result.title)
	print("URL: "+result.url)
	print("Image URL: "+result.image_url)
```

**Expected Behavior**

I would expect image_url to be set when getting the results of an image search. It's ok if the width and height may not be available.

**Environment:**
- OS: Windows 10
- Firecrawl Version: 4.3.6

nehaaprasad:
@JacobBruce 
fixed the image search issue camelCase field mapping between the API and SDK, and now image_url, image_width, and image_height.

PR #3914, opened 2026-06-29
fix(python-sdk): map camelCase image/news fields in async search

mjlbach:
### Problem

`AsyncFirecrawl.search(sources=["images"])` returns image results whose `image_url`, `image_width` and `image_height` are always `None` — the same symptom as #2200, which was fixed for the **sync** client in #2244.

The fix in #2244 added `_map_search_result_keys` and wired it into the sync transformer (`v2/methods/search.py::_transform_array`). The **async** transformer (`v2/methods/aio/search.py::_transform_array`) is a parallel copy that was never updated — it still does `result_type(**item)` directly, so the API's camelCase keys (`imageUrl`, `imageWidth`, `imageHeight`) don't map onto the snake_case model fields and get dropped.

### Reproduction (latest `firecrawl-py==4.30.4`)

Running both transformers on the exact camelCase payload the API returns:

```
SYNC  -> image_url='https://cdn.example.com/logo.png'  width=640  height=320
ASYNC -> image_url=None                                width=None height=None
```

`sync imports _map_search_result_keys:  True`
`async imports _map_search_result_keys: False`

### Fix

Mirror the sync `_transform_array` logic in the async module: import `_map_search_result_keys` and apply the same `images`/`news`/`web` key normalization before constructing the result model. After the change the async path returns populated `image_url`/`image_width`/`image_height` (and `image_url` for news), matching the sync client.

This is a behavior-only change scoped to the async search transformer; no public API changes.

Closes the async half of #2200.

---
## Summary by cubic
Fixes async search in `firecrawl-py` by mapping camelCase API fields to snake_case models so image/news results return the expected values. Aligns async behavior with the sync client; no public API changes.

- **Bug Fixes**
  - In `v2/methods/aio/search.py`, use `_map_search_result_keys` for `images`, `news`, and `web` before building result models.
  - Restores `image_url`, `image_width`, `image_height` for images and `image_url` for news in `AsyncFirecrawl.search(...)`.

<sup>Written for commit 5dd6a3edf186a8a6437c8951e4cace24062eb462. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/firecrawl/pull/3914?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/python-sdk/firecrawl/v2/methods/aio/search.py
+++ b/apps/python-sdk/firecrawl/v2/methods/aio/search.py
@@ -9,7 +9,7 @@ from ...types import (
 )
 from ...utils.http_client_async import AsyncHttpClient
 from ...utils.error_handler import handle_response_error
-from ...utils.normalize import normalize_document_input
+from ...utils.normalize import normalize_document_input, _map_search_result_keys
 from ...utils.validation import validate_scrape_options, prepare_scrape_options
 
 T = TypeVar("T")
@@ -75,7 +75,19 @@ def _transform_array(arr: List[Any], result_type: Type[T]) -> List[Union[T, Docu
             ):
                 results.append(Document(**normalize_document_input(item)))
             else:
-                results.append(result_type(**item))
+                result_type_name = None
+                if result_type == SearchResultImages:
+                    result_type_name = "images"
+                elif result_type == SearchResultNews:
+                    result_type_name = "news"
+                elif result_type == SearchResultWeb:
+                    result_type_name = "web"
+
+                if result_type_name:
+                    normalized_item = _map_search_result_keys(item, result_type_name)
+                    results.append(result_type(**normalized_item))
+                else:
+                    results.append(result_type(**item))
         else:
             results.append(result_type(url=item))
     return results


mjlbach:
Minimal repro — no API key or network needed. It runs the SDK's own sync vs. async result transformers on a representative `sources=["images"]` payload, so it isolates the parsing difference:

```python
from firecrawl.types import SearchResultImages
from firecrawl.v2.methods.search import _transform_array as sync_transform
from firecrawl.v2.methods.aio.search import _transform_array as async_transform

# Shape the API returns for sources=["images"] (camelCase).
api_item = {
    "title": "Example",
    "imageUrl": "https://cdn.example.com/logo.png",
    "imageWidth": 640,
    "imageHeight": 320,
    "url": "https://example.com/source-page",
    "position": 1,
}

sync_result = sync_transform([dict(api_item)], SearchResultImages)[0]
async_result = async_transform([dict(api_item)], SearchResultImages)[0]

print("sync :", sync_result.image_url, sync_result.image_width, sync_result.image_height)
print("async:", async_result.image_url, async_result.image_width, async_result.image_height)

assert async_result.image_url == api_item["imageUrl"], "async dropped imageUrl"
```

On `main` today (and `firecrawl-py==4.30.4`):

```
sync : https://cdn.example.com/logo.png 640 320
async: None None None
AssertionError: async dropped imageUrl
```

With this PR applied, the async row matches the sync row and the assert passes. An equivalent live check (`await AsyncFirecrawl(api_key=...).search(query="...", sources=["images"], limit=1)`) shows the same `image_url=None` before / populated after.

PR #3967, opened 2026-07-08
feat(api): URL-level threat protection checks, billed per scanned URL

mogery:
Threat protection now classifies **full URLs** instead of domains, and bills per scanned URL.

## What changed

- **URL-level classification**: Web Risk lookups generate host-suffix x path-prefix expressions per the [Safe Browsing spec](https://cloud.google.com/web-risk/docs/urls-hashing) (up to 30 per URL), so a listing that flags a single page is caught even when its domain is otherwise clean — and a flagged domain still blocks every page on it. `checkDomain(domain)` becomes `checkUrl(url)`; bare domains check as their root URL.
- **Canonicalization fix**: the query component is now percent-unescaped before re-escaping, per the spec (previously a valid escape like `%20` double-escaped and could silently never match a list entry). Regression vectors added beyond Google's published set.
- **Local rules unchanged in shape**: whitelist/blacklist/blockedTlds still match on the URL's canonicalized host only.
- **Redirect re-check on any URL change**: same-domain redirects onto a different page are now re-checked too (previously only cross-domain).
- **Billing follows the scan unit**: +2 credits per unique scanned canonical URL per billing scope (`calculateThreatScanCredits` dedups on `decision.url`). Crawls bill per scraped page; search/map bill per unique result URL scanned. Blocked links discovered mid-crawl bill **once per crawl** (crawl-scoped dedup via `recordThreatBlocked` HSETNX), no matter how many pages link to them.
- **Consistent error surface**: extract (v1+v2) and agent blocked-request 403s now carry `code: "unsafe_domain_blocked"` like every other endpoint. The error code name is kept for API stability; the message now reports the blocked URL.
- **Mid-rollout compatibility**: decisions serialized by older deploys (no `url` field) bill individually and deserialize with backfilled fields.

## ZDR

Unchanged posture: clean URLs resolve entirely against the locally synced hash-prefix lists (nothing transmitted); only a local prefix hit sends the anonymized 4-byte hash prefix to `hashes:search`; verdicts are never persisted — dedup is request/job-scoped in-memory only.

## Verification

- `tsc --noEmit` clean; unit suites 196/196 (`src/lib/threat-protection`, `src/lib/scrape-billing.test.ts`), including 46 canonicalization tests (Google's published vectors + query-escape regressions).
- Snips updated: enforcement gains a flagged-URL-on-clean-domain case; billing asserts per-URL fees (redirect two-scan, search per-result-URL).
- No new dependencies, no DDL, no env changes.

---
## Summary by cubic
Threat protection now scans full URLs and bills +2 credits per unique scanned URL. This catches flagged pages on clean domains, re-checks on any URL change, and tightens crawl billing dedup.

- **New Features**
  - URL-level checks using Safe Browsing host-suffix × path-prefix expressions; bare domains scan as root URLs.
  - Redirects re-check on any URL change (same-domain path changes included).
  - Billing per unique scanned canonical URL; search/map bill per result URL; crawls dedupe blocked-link fees once per crawl by keying on canonical `decision.url`. For `batch-scrape` and `agent`, if the whole request is rejected, all scanned URLs bill; with `ignoreInvalidURLs`, only blocked ones bill here.
  - Canonicalization: percent‑unescape query before re‑escaping; `normalizeDomain` falls back to the canonicalizer’s lenient splitter and strips fragments so escaped/edge hosts hit local rules.
  - Consistent errors: 403s use code `unsafe_domain_blocked` and include the blocked URL; decisions include `url` and `domain`.

- **Migration**
  - Replace `checkDomain(domain)` with `checkUrl(url)` across the API; pass full URLs (bare domains are OK).
  - When reading decisions/errors, use `decision.url` for billing/logs and handle `code: "unsafe_domain_blocked"` with messages that reference the URL.

<sup>Written for commit 07cb3aa094e5d4c13140aa4491edbe3ef30bfa29. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/firecrawl/pull/3967?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/api/src/__tests__/snips/v2/threat-protection-billing.test.ts
+++ b/apps/api/src/__tests__/snips/v2/threat-protection-billing.test.ts
@@ -25,10 +25,12 @@ import {
 // Threat protection billing (ENG-4985, ZDR rework ENG-5004)
 //
 // Billing rules under test:
-//   - +2 credits per domain scanned in "normal" mode (Google Web Risk)
-//   - a "scan" = a ThreatDecision with providerConsulted — +2 per consulted
-//     verdict. There is no verdict cache anymore (ZDR): every request scans
-//     afresh, and dedup only applies within one request/job.
+//   - +2 credits per URL scanned in "normal" mode (Google Web Risk).
+//     Checks are URL-level and so is billing: consulted decisions bill once
+//     per unique canonical URL within one request/job
+//     (calculateThreatScanCredits).
+//   - There is no verdict cache (ZDR): every request scans afresh, and
+//     dedup only applies within one request/job.
 //   - blocked requests still bill the scan fee (no base scrape cost, matching
 //     how other failed scrapes bill)
 //   - local-only decisions (whitelist/blacklist/blocked-tld) never bill
@@ -55,9 +57,9 @@ const CLEAN_URL = TEST_SUITE_WEBSITE;
 
 // A stable cross-hostname redirect: google.com 301s to www.google.com. The
 // mock flags only the redirect target, so the scrape consults the provider
-// twice (initial domain + redirect re-check) before being blocked.
-// NOTE the two hostnames are different domains for the request-scoped dedup,
-// so the redirect re-check is a second scan (two consulted decisions).
+// twice (initial URL + redirect re-check) before being blocked.
+// NOTE the redirect lands on a different URL, so the re-check is a second
+// billable scan (billing dedups per canonical URL, and these differ).
 const REDIRECT_SOURCE_URL = "https://google.com/";
 const REDIRECT_TARGET_DOMAIN = "www.google.com";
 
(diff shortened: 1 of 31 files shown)


mogery:
Addressed three of cubic's four findings in 8aa67f5aa:

1. **normalizeDomain host extraction** — now falls back to the canonicalizer's lenient `splitUrl` when WHATWG URL parsing rejects the host, with a regression vector (`http://%20leadingspace.com/path` → `%20leadingspace.com` instead of `http`).
2. **v1/v2 batch scrape + agent rejection billing** — a rejected (403) request now bills every scanned URL, not just the blocked ones, since no scrape jobs will run to bill the allowed scans (matches extract). `ignoreInvalidURLs` batches still bill blocked-only at the controller.
3. **Crawl dedup key** — `recordThreatBlocked` is now keyed on the canonical `decision.url`, so raw spelling variants of one URL share a single fee.

On the **crawl seed preflight double-scan**: this is a deliberate, pre-existing tradeoff, unchanged by this PR — the controller check gives a synchronous 403 for blocked seeds, and the seed's scrape job independently re-checks because verdicts are never persisted or carried across the queue boundary (ZDR). For allowed (clean) seeds both checks resolve locally against the synced hash-prefix lists with zero Google calls, and the unbilled preflight errs in the customer's favor. Threading a decision through the queue would be the first piece of scrape-derived data to cross that boundary, which we deliberately avoid.

PR #4002, opened 2026-07-12
fix(python-sdk): map min_age to minAge in scrape options

### LINKED ISSUE #4001: Python SDK drops the min_age scrape option (never mapped to minAge)
### Bug

The v2 Python SDK never sends `min_age`. `prepare_scrape_options` camelCases every snake_case scrape option through the `field_mappings` dict, but `min_age` has no entry (it is the only ScrapeOptions field without one), so it stays as `min_age` in the payload. The v2 backend strips unknown keys, so the value is dropped and the min-age cache window is never applied.

The same omission surfaces in `/v2/parse`: parse strips the cache hint by its camelCase name with `opts.pop("minAge", None)` (`methods/parse.py`), which never matches the emitted `min_age`, so `min_age` is carried into the always-uncached parse body.

### Reproduce

From `apps/python-sdk`:

```python
from firecrawl.v2.types import ScrapeOptions
from firecrawl.v2.utils.validation import prepare_scrape_options

r = prepare_scrape_options(ScrapeOptions(min_age=1000, max_age=5000))
print("minAge" in r)   # False  (expected True)
print("min_age" in r)  # True   (snake_case leaks through)
print("maxAge" in r)   # True   (mapped correctly)
```

### Fix

Add `"min_age": "minAge"` to the `field_mappings` dict in `prepare_scrape_options`, next to `"max_age": "maxAge"`. One mapping covers every caller (scrape, crawl, batch, parse) and makes the existing `opts.pop("minAge")` in parse strip it correctly.

saivedant169:
## What and why

`prepare_scrape_options` camelCases every snake_case scrape option through `field_mappings`, but `min_age` had no entry, so it was sent verbatim as `min_age`. The v2 backend strips unknown keys, so the value was dropped and the min-age cache window never applied. It also leaked into the `/v2/parse` body, because parse strips the field by its camelCase name (`minAge`), which never matched the emitted `min_age`.

## Change

- Add `"min_age": "minAge"` to `field_mappings` in `prepare_scrape_options`. The single mapping fixes both the scrape and the parse paths.
- `test_prepare_min_age_maps_to_camel_case` (validation) asserts `minAge` is emitted and the snake_case key is gone.
- `test_prepare_parse_request_strips_min_age` (parse) asserts neither `minAge` nor `min_age` reaches the parse body.

Both tests fail on the current code (validation raises `KeyError: minAge`, parse leaks `min_age`) and pass with the fix. The full python-sdk unit suite (397 tests) passes.

Fixes #4001

---
## Summary by cubic
Maps `min_age` to `minAge` in the Python SDK scrape options so v2 applies the min-age cache window, and strips the field from `/v2/parse` payloads.

- **Bug Fixes**
  - Added `"min_age": "minAge"` to `field_mappings` in `prepare_scrape_options`, fixing both scrape and parse paths.
  - Added tests verifying scrape emits `minAge` (and removes `min_age`), parse payload excludes both keys, and `max_age` continues to map to `maxAge`.

<sup>Written for commit 295d993545a84c3ac9eb1e9e7466cd358b98c0d8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/firecrawl/pull/4002?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/python-sdk/firecrawl/__tests__/unit/v2/methods/test_parse_request_preparation.py
+++ b/apps/python-sdk/firecrawl/__tests__/unit/v2/methods/test_parse_request_preparation.py
@@ -87,3 +87,16 @@ class TestParseRequestPreparation:
 
         payload = json.loads(fields["options"])
         assert "lockdown" not in payload
+
+    def test_prepare_parse_request_strips_min_age(self):
+        options = ParseOptions(formats=["markdown"], min_age=1000)
+        fields, _ = _prepare_parse_request(
+            b"<html><body><h1>Parse</h1></body></html>",
+            options,
+            filename="upload.html",
+            content_type="text/html",
+        )
+
+        payload = json.loads(fields["options"])
+        assert "minAge" not in payload
+        assert "min_age" not in payload
--- a/apps/python-sdk/firecrawl/__tests__/unit/v2/utils/test_validation.py
+++ b/apps/python-sdk/firecrawl/__tests__/unit/v2/utils/test_validation.py
@@ -388,3 +388,13 @@ class TestPrepareScrapeOptions:
         result = prepare_scrape_options(options)
 
         assert result["parsers"][0]["maxPages"] == 5
+
+    def test_prepare_min_age_maps_to_camel_case(self):
+        """min_age must be sent as minAge; the server drops the snake_case key."""
+        options = ScrapeOptions(min_age=1000, max_age=5000)
+
+        result = prepare_scrape_options(options)
+
+        assert result["minAge"] == 1000
+        assert "min_age" not in result
+        assert result["maxAge"] == 5000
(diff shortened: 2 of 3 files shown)


saivedant169:
Hi @mogery, could you take a look at this one when you have a moment? Small python-sdk fix that maps `min_age` to `minAge` in scrape options so the snake_case field actually applies. Happy to rebase if useful.

mogery (approved):

mogery:
Thank you!

PR #4057, opened 2026-07-17
fix(scraper/fire-engine): decode responses using Content-Type/meta charset instead of forcing UTF-8

### LINKED ISSUE #4047: fire-engine responses are always decoded as UTF-8, mangling non-UTF-8 pages
## Description

In `apps/api/src/scraper/scrapeURL/engines/fire-engine/index.ts`, the response body is unconditionally decoded as UTF-8:

```ts
status.content = buffer.toString("utf8"); // TODO: handle other encodings via Content-Type tag
```

Pages served with other charsets (e.g. `Shift_JIS`, `EUC-KR`, `GBK`, `windows-1251`, `ISO-8859-*`) get garbled content, which then propagates into the markdown output and any downstream extraction.

## Suggested fix

- Parse the charset from the `Content-Type` response header (and fall back to sniffing `<meta charset>` / `<meta http-equiv="Content-Type">` in the HTML prologue, like browsers do).
- Decode with the detected encoding (e.g. via `TextDecoder`, which supports the WHATWG encoding set) before further processing.
- Add a snip test with a fixture page served in a non-UTF-8 encoding asserting the markdown contains correctly decoded text.

rohitsux:
Fixes #4047.

## Problem

The fire-engine file-download path decoded every response body as UTF-8:

```ts
status.content = buffer.toString("utf8"); // TODO: handle other encodings via Content-Type tag
```

The inline `TODO` acknowledged the gap. Any page served in a non-UTF-8 charset (Shift_JIS, windows-1251, GBK, ...) came back as mojibake.

## Fix

The fetch engine already solves this: it decodes using the `Content-Type` header charset, falls back to an HTML `<meta charset>` sniff, and finally to UTF-8. This PR extracts that logic (verbatim, no behavior change) into a shared `decodeHtmlBuffer` helper under `engines/utils/`, and uses it in fire-engine:

```ts
status.content = decodeHtmlBuffer(buffer, contentType).text;
```

`contentType` is already computed in that scope from the response headers. The fetch engine now imports the shared helper instead of its own copy, so both engines share one implementation.

## Test

Adds `engines/utils/__tests__/decodeHtmlBuffer.test.ts` covering:
- Shift_JIS body decoded via the `Content-Type` header charset (and asserting the old naive UTF-8 decode produced mojibake).
- windows-1251 body decoded via the `<meta charset>` fallback when no header charset is present.
- UTF-8 default when no charset is declared.

## Verification

- `npx vitest run src/scraper/scrapeURL/engines/utils/__tests__/decodeHtmlBuffer.test.ts` in `apps/api` -> 3 passed.

## Scope

This reuses the fetch engine's existing charset logic for the fire-engine path; it does not add new charset-detection heuristics (e.g. BOM or statistical sniffing) beyond the current header + `<meta>` detection, and does not touch other engines' decoding.

---
## Summary by cubic
Stop forcing UTF-8 in `fire-engine`. Decode HTML using `Content-Type` or `<meta charset>` with UTF-8 fallback, and ignore lookalike params/attrs (e.g., `x-charset`, `<meta data-charset>`); fixes #4047.

- **Bug Fixes**
  - `fire-engine` honors header/meta charset with UTF-8 fallback and avoids false positives that could corrupt UTF-8 content.

- **Refactors**
  - Extracted shared `decodeHtmlBuffer` to `engines/utils/` and updated `fetch` to use it.
  - Added tests for header charset, meta fallback, UTF-8 default, and false-positive cases.

<sup>Written for commit 2a4798385eaed61039282011199ce5a0c9f746c0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/firecrawl/pull/4057?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


rohitsux:
Thanks @cubic-dev-ai — both are valid. The `charset=` matches were substring matches, so a Content-Type parameter like `x-charset=` and a `<meta data-charset=...>` attribute were incorrectly treated as charset declarations, which would corrupt otherwise-UTF-8 content.

Fixed by anchoring both regexes to a token boundary:

- Header: `charset` must follow start-of-string, whitespace, or a `;` parameter delimiter.
- Meta: `charset` must follow an attribute boundary (whitespace or a quote).

Added regression tests asserting `x-charset=Shift_JIS` and `<meta data-charset="windows-1251">` no longer change decoding, alongside the existing Shift_JIS-header and windows-1251-meta cases. Local run: 5/5 passing.

amyrpayyndhy1-blip:
در تاریخ جمعه ۱۷ ژوئیهٔ ۲۰۲۶،‏ ۲۱:۰۳ cubic-dev-ai[bot] <
***@***.***> نوشت:

rohitsux:
@nickscamara this is ready for review when you get a chance — it fixes #4047 (the fire-engine UTF-8 forcing you filed) by reusing the fetch engine's charset logic via a shared helper. cubic's review feedback has been addressed (charset now matched only as a standalone param/attribute, with regression tests) and the unit tests pass 5/5 locally.

PR #4084, opened 2026-07-20
fix(api): treat empty OPENAI_API_KEY/BASE_URL as unset

### LINKED ISSUE #4083: Empty OPENAI_API_KEY / OPENAI_BASE_URL bypass zod handling: `json` scrape format fails silently while still consuming credits
## Summary

`apps/api/src/config.ts` defines an `emptyStringAsUndefined` helper (line 12) and applies it to several env vars (lines 132, 133, 152, 155) — but `OPENAI_API_KEY` (line 81) and `OPENAI_BASE_URL` (line 82) are declared as plain `z.string().optional()`.

This matters because the documented self-host path produces **present-but-empty** values rather than unset ones: `SELF_HOST.md` ships these keys commented out, but a `.env` with `OPENAI_BASE_URL=` combined with `docker-compose.yaml`'s `OPENAI_BASE_URL: ${OPENAI_BASE_URL}` passthrough hands the container an empty string. Zod keeps it, and two defensive layers are silently defeated:

1. `loadApiKey` accepts `""` — it only checks `typeof apiKey === "string"` — so the intended `OpenAI API key is missing` error never fires.
2. `createOpenAI({ baseURL: "" })` — `""` is not `null`/`undefined`, so the default `https://api.openai.com/v1` is never applied.

The request URL ends up being `"" + "/responses"`, producing:

```
TypeError: Failed to parse URL from /responses
  cause: { code: 'ERR_INVALID_URL', input: '/responses' }
```

A misconfiguration surfaces as an opaque URL-parsing error that points nowhere near the actual cause.

## Impact: the `json` format fails silently and still bills

The confusing message is the smaller problem. The four LLM entry points degrade in four different ways, and one is completely silent:

| Request | Behaviour |
|---|---|
| `POST /v2/scrape` with a `json` format | 🔴 **HTTP 200, `success: true`, no `json` field** — and `creditsUsed` is still charged |
| `POST /v2/scrape` with `summary` | HTTP 500, opaque `UNKNOWN_ERROR` + Error ID, no reason given |
| `POST /v2/extract` | HTTP 200 + job id, then `status: "failed"` with the message |
| `POST /v1/extract` | HTTP 500 with the underlying message |

The silent case comes from `scraper/scrapeURL/lib/extractSmartScrape.ts`, whose `catch` swallows everything except `CostLimitExceededError`, after which `scraper/scrapeURL/transformers/index.ts` only emits a `warn`. `performSummary` has no `try/catch`, which is why `summary` escapes to a 500 instead — hence the inconsistency between the two.

A caller that checks `success` sees a successful scrape with quietly missing data.

## Reproduction

Self-host via docker compose with a `.env` containing:

```
OPENAI_API_KEY=
OPENAI_BASE_URL=
```

Then:

```bash
curl -X POST localhost:3002/v2/scrape -H 'Content-Type: application/json' -d '{
  "url": "https://example.com",
  "formats": [{
    "type": "json",
    "prompt": "extract the title",
    "schema": {"type":"object","properties":{"title":{"type":"string"}}}
  }]
}'
```

Response is `200` with `"success": true`, `creditsUsed: 5`, and no `json` field. The container log shows:

```
error [ScrapeURL:performLLMExtract/generateCompletions]: LLM extraction failed
  ... "message":"Failed to parse URL from /responses" ...
warn  [ScrapeURL:executeTransformers/coerceFieldsToFormats]: Request had format json,
      but there was no json field in the result.
```

## Suggested fix

```ts
OPENAI_API_KEY: emptyStringAsUndefined(z.string()),
OPENAI_BASE_URL: emptyStringAsUndefined(z.string()),
```

With this, an empty key produces the intended "API key missing" error, and an empty base URL lets the built-in default apply. The same reasoning probably applies to `MODEL_NAME`, `MODEL_EMBEDDING_NAME`, `OLLAMA_BASE_URL` and the other provider keys, since `docker-compose.yaml` passes all of them through unconditionally.

Separately, it may be worth surfacing an explicit "no LLM provider configured" error at request time. There is currently no runtime guard — `HAS_AI` exists only in the test suite (`__tests__/snips/lib.ts`) — so any deployment without AI env vars hits this path rather than getting a clear message.

## Environment

Verified against `main` today: `apps/api/src/config.ts` lines 81-82 are unchanged upstream, and the helper at line 12 is used at 132/133/152/155. Reproduced on a docker compose self-host (`USE_DB_AUTHENTICATION=false`) with `@ai-sdk/openai` 3.0.71 / `ai` 6.0.86.

Happy to send a PR if you'd like.

Solaris-star:
## Summary
Self-hosted setups often get **present-but-empty** `OPENAI_API_KEY` / `OPENAI_BASE_URL` from docker-compose `${VAR}` expansion. Those were declared as plain `z.string().optional()`, so Zod kept `""` and:

1. `loadApiKey` accepted empty strings
2. `createOpenAI({ baseURL: "" })` did not fall back to the default API host

→ opaque `TypeError: Failed to parse URL from /responses` while still consuming credits on `json` scrapes.

## Change
Wrap both fields with the existing `emptyStringAsUndefined` helper (same pattern as `FDB_CLUSTER_FILE`, `NUQ_BACKEND`, etc.).

## Test plan
- [ ] `OPENAI_BASE_URL=` in `.env` + compose → config has `undefined`, default OpenAI base used when key is set
- [ ] Missing key surfaces the intended missing-key path instead of URL parse errors

Fixes #4083.

---
## Summary by cubic
Treat empty and whitespace-only `OPENAI_API_KEY` and `OPENAI_BASE_URL` from docker-compose `${VAR}` as unset. This makes `loadApiKey` reject invalid keys and lets `createOpenAI` use the default base URL, avoiding URL parse errors and 401s.

- **Bug Fixes**
  - Trim whitespace-only env values to `undefined` in config.
  - Added tests for empty, whitespace-only, and valid values.

<sup>Written for commit bed234144ac429a76bf4d38db6eb4e7c28c07e23. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/firecrawl/pull/4084?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/api/src/__tests__/config.test.ts
+++ b/apps/api/src/__tests__/config.test.ts
@@ -1,0 +1,44 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+// config.ts runs `configSchema.parse(process.env)` at import time, so each case
+// sets the env then imports the module fresh via vi.resetModules().
+describe("config emptyStringAsUndefined (OPENAI_API_KEY / OPENAI_BASE_URL)", () => {
+  const original = { ...process.env };
+
+  beforeEach(() => {
+    vi.resetModules();
+  });
+
+  afterEach(() => {
+    process.env = { ...original };
+  });
+
+  async function loadConfig() {
+    const mod = await import("../config");
+    return mod.config;
+  }
+
+  it("treats an empty string as undefined", async () => {
+    process.env.OPENAI_API_KEY = "";
+    process.env.OPENAI_BASE_URL = "";
+    const config = await loadConfig();
+    expect(config.OPENAI_API_KEY).toBeUndefined();
+    expect(config.OPENAI_BASE_URL).toBeUndefined();
+  });
+
+  it("treats a whitespace-only string as undefined", async () => {
+    process.env.OPENAI_API_KEY = "   ";
+    process.env.OPENAI_BASE_URL = "\t\n ";
+    const config = await loadConfig();
+    expect(config.OPENAI_API_KEY).toBeUndefined();
+    expect(config.OPENAI_BASE_URL).toBeUndefined();
+  });
+
+  it("passes a real value through unchanged", async () => {
+    process.env.OPENAI_API_KEY = "sk-real-key";
+    process.env.OPENAI_BASE_URL = "https://api.openai.com/v1";
+    const config = await loadConfig();
+    expect(config.OPENAI_API_KEY).toBe("sk-real-key");
+    expect(config.OPENAI_BASE_URL).toBe("https://api.openai.com/v1");
+  });
+});
--- a/apps/api/src/config.ts
+++ b/apps/api/src/config.ts
@@ -10,7 +10,11 @@ const delimitedList = (separator = ",") => {
 };
 
 const emptyStringAsUndefined = <T extends z.ZodTypeAny>(schema: T) =>
-  z.preprocess(value => (value === "" ? undefined : value), schema.optional());
+  z.preprocess(
+    value =>
+      typeof value === "string" && value.trim() === "" ? undefined : value,
+    schema.optional(),
+  );
 
 const emptyStringAsDefault = <T extends z.ZodTypeAny>(schema: T) =>
   z.preprocess(value => (value === "" ? undefined : value), schema);
--- a/apps/api/src/config.ts
+++ b/apps/api/src/config.ts
@@ -78,8 +78,10 @@ const configSchema = z.object({
 
   // API Keys & Authentication
   BULL_AUTH_KEY: z.string().optional(),
-  OPENAI_API_KEY: z.string().optional(),
-  OPENAI_BASE_URL: z.string().optional(),
+  // Empty strings from docker-compose ${VAR} passthrough must become undefined
+  // so loadApiKey / createOpenAI fall back correctly (see #4083).
+  OPENAI_API_KEY: emptyStringAsUndefined(z.string()),
+  OPENAI_BASE_URL: emptyStringAsUndefined(z.string()),
   OPENROUTER_API_KEY: z.string().optional(),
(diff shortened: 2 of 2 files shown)


rkfshakti (approved):
Clean fix for a common Docker Compose footgun. When environment variables are passed through from a `.env` file or docker-compose `${VAR}` syntax, unset variables become empty strings rather than `undefined`. This causes Zod validation to pass (empty string is a valid string) but downstream consumers like `loadApiKey` and `createOpenAI\) fail because they expect either a real key or `undefined`.

The `emptyStringAsUndefined` wrapper is the right pattern — it's a simple transform that handles the mismatch between Docker's passthrough semantics and the application's optional-field semantics. Minimal change, clear intent.

LGTM.

Solaris-star:
Addressed the cubic P2 suggestion: `emptyStringAsUndefined` now trims before the empty check (`typeof value === "string" && value.trim() === "" ? undefined : value`), so whitespace-only values like `"   "` also fall back to `undefined` instead of reaching `loadApiKey` as a truthy-but-invalid key. Applied it in the shared helper so every `emptyStringAsUndefined(...)` field (OPENAI_API_KEY, OPENAI_BASE_URL, etc.) benefits.

Added `apps/api/src/__tests__/config.test.ts` covering empty, whitespace-only, and real-value cases. Verified the whitespace test fails on the pre-fix logic and passes after; full file green via `vitest run`.

PR #4099, opened 2026-07-22
fix(api): bump default timeout for branding format

abimaelmartell:
## What

Bumps the default timeout to 120s when the `branding` format is requested and the user did not set an explicit timeout (same pattern as the existing `json`/`changeTracking` 60s bumps and the stealth-proxy 120s bump in `extractTransformImpl`).

## Why

Branding is the heaviest format we have, but it was the only heavy format with no timeout bump:

- the index cache is disabled for it, so it always does a full chrome-cdp render
- it sets `blockMedia: false` and injects a 2s settle wait
- it runs a large in-page extraction script
- it makes a synchronous LLM call inside the timeout budget

On JS-heavy pages a branding scrape routinely needs 100–120s (verified against production data for a customer report: branding scrapes of a heavy page consistently completed in 101–118s and always died at a 90s cap). With the 30s default it times out on exactly the sites customers most want branding from.

Explicit user-set timeouts are still respected — the bump only applies to the schema default.

## Tests

Added two cases to `snips/v2/types-validation.test.ts`: default is transformed to 120000, and an explicit 90000 is left untouched. Full file passes (116 tests).

https://claude.ai/code/session_01Cx6pZQ6JaKSURsF2owpsFA

---
## Summary by cubic
Bumps the default timeout to 120s for the `branding` format when no explicit timeout is set, matching existing bumps for `json`/`changeTracking` and stealth proxy. This reduces timeouts on heavy, JS-heavy pages while respecting user-set timeouts.

- **Bug Fixes**
  - In `extractTransformImpl`, if `branding` is requested and timeout is `30000`, set it to `120000`.
  - Added tests to verify the default transforms to `120000` and that an explicit timeout (e.g., `90000`) is preserved.

<sup>Written for commit 51a4640558cac842d2196eed8b7b9ca88355dd1e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/firecrawl/pull/4099?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/api/src/__tests__/snips/v2/types-validation.test.ts
+++ b/apps/api/src/__tests__/snips/v2/types-validation.test.ts
@@ -1523,6 +1523,30 @@ describe("V2 Types Validation", () => {
       expect(result.waitFor).toBeGreaterThanOrEqual(5000); // Should be at least 5000
     });
 
+    it("should handle branding format timeout transformation", () => {
+      const input: ScrapeRequestInput = {
+        url: "https://example.com",
+        formats: [{ type: "branding" }],
+        timeout: 30000,
+        proxy: "basic", // Use basic proxy to avoid auto/stealth timeout transformation
+      };
+
+      const result = scrapeRequestSchema.parse(input);
+      expect(result.timeout).toBe(120000); // Should be transformed
+    });
+
+    it("should not override an explicit timeout for branding format", () => {
+      const input: ScrapeRequestInput = {
+        url: "https://example.com",
+        formats: [{ type: "branding" }],
+        timeout: 90000,
+        proxy: "basic",
+      };
+
+      const result = scrapeRequestSchema.parse(input);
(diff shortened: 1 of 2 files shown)


abimaelmartell:
Closing in favor of the root-cause fix: the slowness was not inherent to the branding format — branding-only scrapes were being routed to a rendering-focused browser pool they don't need (media loading implies visual rendering in the current routing), where visually heavy pages can block the renderer for extended periods. With the routing fix, branding completes in seconds and needs no extra timeout headroom.

PR #4101, opened 2026-07-22
perf(branding): cut redundant DOM walks in the in-page script

abimaelmartell:
## What

Two contained performance fixes in the branding extraction script (`branding-script/images.ts`):

1. **Shadow-root discovery is now done once per run.** `querySelectorAllIncludingShadowRoots` re-walked the entire DOM (`querySelectorAll('*')` per root) on every call — and it's called ~58 times (logo selectors + container selectors + img/svg/a sweeps). That's O(selectors × DOM) renderer work inside the scrape budget. The root list is now discovered once and reused for every selector query.

2. **The div/span background-image sweep does cheap checks first.** It previously forced a computed-style read on *every* div/span on the page before applying its cheap attribute/class/ancestry filters. On style-recalc-heavy pages (many `backdrop-filter` layers, common with Tailwind) each forced read is expensive. Attribute checks now run first; computed style is only read for elements that pass.

## Behavior

No output change — the final predicate and candidate collection are identical, only evaluation order/caching changed. Verified against a live production page: identical `logoCandidates` output from the old and new bundles, both via chrome-cdp `executeJavascript`.

## Context

Found while investigating slow branding scrapes on style-heavy pages (customer report of consistent `SCRAPE_TIMEOUT` on `formats: [branding]`). Companion fixes in fire-engine address the larger stalls; this removes the branding script's own worst-case DOM cost.

https://claude.ai/code/session_01Cx6pZQ6JaKSURsF2owpsFA

---
## Summary by cubic
Speed up the branding in-page script by removing redundant DOM walks and avoiding unnecessary computed-style reads. This reduces stalls on style-heavy pages and helps prevent scrape timeouts; output is unchanged.

- **Refactors**
  - Cache shadow roots once and reuse them for all selector queries in `querySelectorAllIncludingShadowRoots`; preserve document-order traversal to keep first-seen dedup stable.
  - In the div/span background-image sweep, run cheap attribute/DOM checks first and only read computed style for likely candidates.

<sup>Written for commit a95b4767b6f196a86f1125b52f0434d2f75a6e90. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/firecrawl/pull/4101?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/api/src/scraper/scrapeURL/engines/fire-engine/branding-script/images.ts
+++ b/apps/api/src/scraper/scrapeURL/engines/fire-engine/branding-script/images.ts
@@ -57,11 +57,18 @@ export const findImages = (): FindImagesResult => {
       seenRoots.add(root);
       roots.push(root);
       try {
+        // Push discovered roots in reverse so pop() visits siblings in
+        // document order — preserving the previous recursive (pre-order DFS)
+        // traversal, which candidate dedup depends on for first-seen ties.
+        const discovered: ShadowRoot[] = [];
         root.querySelectorAll("*").forEach(el => {
           const shadow = (el as Element & { shadowRoot?: ShadowRoot })
             .shadowRoot;
-          if (shadow) pending.push(shadow);
+          if (shadow) discovered.push(shadow);
         });
+        for (let i = discovered.length - 1; i >= 0; i--) {
+          pending.push(discovered[i]);
+        }
       } catch (_) {
         // Ignore errors
       }
--- a/apps/api/src/scraper/scrapeURL/engines/fire-engine/branding-script/images.ts
+++ b/apps/api/src/scraper/scrapeURL/engines/fire-engine/branding-script/images.ts
@@ -45,27 +45,41 @@ export const findImages = (): FindImagesResult => {
     if (src) imgs.push({ type, src });
   };
 
-  const querySelectorAllIncludingShadowRoots = (
-    selector: string,
-  ): Element[] => {
-    const results: Element[] = [];
+  // Discover shadow roots once — walking every element per selector query is
+  // O(selectors × DOM) and stalls the renderer on large pages.
+  const allQueryRoots: Array<Document | ShadowRoot> = (() => {
+    const roots: Array<Document | ShadowRoot> = [];
     const seenRoots = new Set<Document | ShadowRoot>();
-    function walk(root: Document | ShadowRoot): void {
-      if (!root || seenRoots.has(root)) return;
+    const pending: Array<Document | ShadowRoot> = [document];
+    while (pending.length > 0) {
+      const root = pending.pop()!;
+      if (!root || seenRoots.has(root)) continue;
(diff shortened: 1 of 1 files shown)


abimaelmartell:
## Eval: old vs new script over the internal eval corpus

Ran both bundles (merge-base vs PR head) against all **155 active URLs** from the internal branding eval collection — paired scrapes, each injecting the script via `executeJavascript` after an identical 2s wait, comparing a digest of: hashed logo-candidate set (src+location+visibility), picked logo, brand name, background-candidate count.

**Result: no reproducible behavioral difference.**

| | count |
|---|---|
| exact parity | 137 |
| initial mismatch | 14 |
| errored both variants (bot-walled/unreachable) | 4 |

All 14 mismatches were re-run as **old-vs-old and new-vs-new control pairs**: 9 sites self-mismatch across identical-script loads (dynamic asset URLs, geo/page variants), 4 reached full parity on re-run, and the largest outlier (19→1 candidates) reproduced *within the old script alone* — lazy-load variance at the 2s mark. Zero cases of a stable old≠new difference.

**In-page execution time** (n=151):

| | old | new |
|---|---|---|
| p50 | 38ms | **25ms** |
| p90 | 80ms | **61ms** |
| max | 395ms | **143ms** |

The corpus is light on shadow-DOM sites, so the O(selectors×DOM) walk savings mostly shows in the tail; the pathological cases this PR targets (huge DOMs, style-recalc-heavy pages) sit beyond p90 here.

abimaelmartell:
## Eval round 2 — after the shadow-root traversal-order fix (a95b476)

Re-ran the full 155-URL paired sweep with the updated head:

| | round 1 | round 2 |
|---|---|---|
| exact parity | 137 | **141** |
| mismatch | 14 | 10 |
| errored both variants | 4 | 4 |
| script time p50 (old→new) | 38→25ms | 43→27ms |
| script time max (old→new) | 395→143ms | 445→175ms |

All 10 round-2 mismatches are load-to-load nondeterminism: 6 were already proven self-mismatching in round 1's old-vs-old/new-vs-new controls, and the two with differing logo picks were re-probed with a full-candidate-detail diff — both returned **byte-identical candidate sets and logo picks** for old vs new when the page state coincided (78=78 candidates and 24=24 candidates respectively, zero set difference).

The traversal-order fix itself can't be exercised by this corpus (no multi-shadow-host pages); it's verified by a jsdom equivalence test: the fixed stack traversal produces exactly the old recursion's pre-order (`doc r1 r1a r1b r2 r3`), where the pre-fix stack produced `doc r3 r2 r1 r1b r1a`.

PR #4102, opened 2026-07-22
feat(api): surface media service errors with a dedicated error code

abimaelmartell:
Superseded by #4103, which relays the service's structured errors generically instead of adding a dedicated code.


abimaelmartell:
Superseded by #4103.

PR #4154, opened 2026-07-27
feat(scrape): classify blocked download failures as retryable

abimaelmartell:
## Why

Traced a high-volume media customer's #1 failure bucket ("500 unexpected (exception ID)") to its actual causes over 2 days. It is **not** content errors or timeouts — it's:

- **124× `Download failed: … HTTP Error 403: Forbidden`** — proxy IP-reputation blocks on the download (the dominant cause)
- **~15× bot-walls** ("not a bot")
- **40× `fetch failed`** — the API couldn't reach avgrab (unreachable/overloaded during bursts)

All of these are **transient** (a retry through a fresh proxy usually works), but the audio/video transformers throw a raw `Error` → wrapped as `UNKNOWN_ERROR` → opaque **500** with an exception ID. The customer can't tell "retry me" from "server bug."

## Change

New retryable `MediaBlockedError` (`SCRAPE_MEDIA_BLOCKED`), mapped to **502** in v1/v2 controllers with a clear "transient — retry the scrape" message:

- `throwIfMediaBlocked()` detects proxy/IP-reputation markers (`HTTP Error 403`, `not a bot`, sign-in walls) in avgrab's string `detail` and throws it. Structured, content-specific denials arrive as an object `detail` and are handled first by `throwIfMediaAccessDenied` (403).
- Connection-level failures reaching avgrab (`fetch failed`) are caught and surfaced as the same retryable error.
- Wired into both the audio and video download paths.

Registered for worker↔controller serialization. Tests: each block marker → `SCRAPE_MEDIA_BLOCKED`, connection failure → retryable, non-block failure → not classified.

## Relationship to the other fixes
- **fire-engine#539** (proxy retry) retries these *inside* avgrab, so most never reach the API. This PR gives the **residual** (whole-pool-bad) a clean, retryable error instead of an opaque 500.
- **#4153** (fetch timeouts) handles the separate metadata-hang timeout bucket.


abimaelmartell:
@cubic-dev-ai

PR #4155, opened 2026-07-28
fix(py-sdk): don't send the API key to cross-host next URLs in the async client

### LINKED ISSUE #4139: Async client follows a cross-host `next` URL and sends the API key to it (sync client rewrites it)
### Summary

On firecrawl-py 4.32.1, the async client follows the `next` URL from a crawl status response even
when it points at a different host, and it sends the API key there. The sync client rewrites that
URL back onto the configured `api_url` first.

The sync path does it deliberately. `v2/utils/http_client.py:30`:

```python
def _build_url(self, endpoint: str) -> str:
    base = urlparse(self.api_url)
    ep = urlparse(endpoint)
    if ep.netloc:
        path = ep.path or "/"
        if (ep.hostname or "") != (base.hostname or ""):
            # Different host: keep path/query but force base host/scheme (no token leakage)
            return urlunparse((base.scheme or "https", base.netloc, path, "", ep.query, ""))
```

Every sync verb runs through it (`:93 :146 :192 :239 :289`).

The async client has no equivalent. `v2/utils/http_client_async.py:29` builds

```python
self._client = httpx.AsyncClient(
    base_url=api_url,
    headers={"Authorization": f"Bearer {api_key}"},
    limits=httpx.Limits(max_keepalive_connections=0),
)
```

and then hands the endpoint straight through at `:149`:

```python
response = await self._client.get(endpoint, headers=..., timeout=timeout)
```

httpx ignores `base_url` when the endpoint is an absolute URL. The Authorization header is set on
the client, so it goes out on whatever request is built.

The endpoint is not always ours. Crawl and batch-scrape auto-pagination feed the server's own
`next` field back into `client.get`, at `v2/methods/aio/crawl.py:193` and `:250`, and the same
shape is in `aio/batch.py:138` and `:192`.

### Reproduction

firecrawl-py 4.32.1, Python 3.14.0, fresh venv. Two loopback origins. The second is dual-stack so
it can be addressed as `localhost` while the first is `127.0.0.1`, which makes them differ by
hostname. Origin one answers `/v2/crawl/job-1` with a `next` pointing at origin two. Both origins
record the Authorization header they receive. Nothing leaves the machine.

| case | requests to origin 1 | requests to origin 2 | origin 2 saw the API key | page 2 came from |
| --- | --- | --- | --- | --- |
| async, cross-host `next` | 1 | **1** | **yes** | origin 2 |
| async, same-host `next` | 2 | 0 | no | origin 1 |
| sync, cross-host `next` | 2 | 0 | no | origin 1 |

Origin two logged `Authorization: Bearer fc-PROBE-SECRET-KEY` on the first row.

The second row is there to show the async path paginates normally when nothing is odd, so the
first row is about the host and not about pagination being broken. The third row is what proves
the harness is really presenting a cross-host URL, since sync sees the same string and does not
follow it.

### Why this is worth fixing even though the `next` URL comes from your own API

Two reasons. A self-hosted or proxied deployment can return a `next` that does not match the
configured `api_url`, and the sync client already treats that as something to neutralize rather
than trust. And the fix is small: the async client can reuse the same `_build_url` logic.

### Suggested fix

Give `AsyncHttpClient` the same `_build_url` the sync client has, and call it in `post`,
`post_multipart`, `get`, `delete` and `patch` before handing the URL to httpx. Moving
`_build_url` into a shared helper would keep the two from drifting again.

Happy to open a PR with that plus a test asserting a cross-host `next` is rewritten, if that is
useful.

Mohith26:
Fixes #4139

## What

The async client passed endpoints straight to httpx. httpx ignores `base_url` when the endpoint is an absolute URL, and the Authorization header is set on the client, so a cross-host `next` URL from crawl/batch auto-pagination was followed as-is with the API key attached.

The sync client already neutralizes this in `_build_url`. This PR extracts that logic into a shared `build_url()` helper in `http_client.py` and routes all five `AsyncHttpClient` verbs (`post`, `post_multipart`, `get`, `delete`, `patch`) through it, so the two clients resolve URLs identically and can't drift again.

## Tests

Added `tests/test_cross_host_url_rewrite.py`:

- unit tests for `build_url` (cross-host, same-host, relative, protocol-relative, userinfo/fragment stripping)
- a sync/async parity test over the same endpoint set
- MockTransport wire tests asserting a cross-host URL is pinned to the configured host on every verb, and that same-host pagination and relative endpoints still work

Without the fix the wire tests fail (the request goes to the foreign host). With it, all 14 pass. Full suite shows no new failures; the 13 failures in `tests/test_timeout_conversion.py` and friends reference removed private APIs and fail on main too.

Credit to @toolshedlabs-hash for the detailed report and root cause analysis in #4139.

---
<a href="https://app.blacksmith.sh/firecrawl/codesmith/firecrawl/pr/4155"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787803652&installation_model_id=11126&pr_number=4155&repository=firecrawl%2Ffirecrawl&return_to=https%3A%2F%2Fgithub.com%2Ffirecrawl%2Ffirecrawl%2Fpull%2F4155&signature=a65f809739ef823711700ff6fac3b3587b1310dc1f9b23a4781fb13f69723750"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Prevents the Python SDK async client from sending the API key to cross-host absolute endpoints by rewriting them onto the configured api host. The async client now resolves URLs like the sync client to block credential leakage via `httpx`.

- **Bug Fixes**
  - Added a shared `build_url()` helper to pin endpoints to the configured host and normalize scheme (handles absolute, relative, and protocol-relative URLs).
  - Routed all async verbs through `build_url()` (`post`, `post_multipart`, `get`, `delete`, `patch`); added tests for cross-host rewrite, same-host pagination, and sync/async parity.

<sup>Written for commit 64846b7d95c1f1055dcb3f86893379ea90d9d440. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/firecrawl/pull/4155?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/python-sdk/firecrawl/v2/utils/http_client.py
+++ b/apps/python-sdk/firecrawl/v2/utils/http_client.py
@@ -10,6 +10,37 @@ from .get_version import get_version
 
 version = get_version()
 
+
+def build_url(api_url: str, endpoint: str) -> str:
+    """Resolve ``endpoint`` against ``api_url``, never leaving the configured host.
+
+    Absolute URLs pointing at a different host (e.g. a ``next`` pagination URL
+    returned by a proxied or self-hosted deployment) are rewritten onto the
+    configured ``api_url`` host/scheme so credentials are never sent to a
+    foreign host.
+    """
+    base = urlparse(api_url)
+    ep = urlparse(endpoint)
+
+    # Absolute or protocol-relative (has netloc)
+    if ep.netloc:
+        # Different host: keep path/query but force base host/scheme (no token leakage)
+        path = ep.path or "/"
+        if (ep.hostname or "") != (base.hostname or ""):
+            return urlunparse((base.scheme or "https", base.netloc, path, "", ep.query, ""))
+        # Same host: normalize scheme to base
+        return urlunparse((base.scheme or "https", base.netloc, path, "", ep.query, ""))
+
+    # Relative (including leading slash or not)
+    base_str = api_url if api_url.endswith("/") else f"{api_url}/"
+    # Guard protocol-relative like //host/path slipping through as “relative”
+    if endpoint.startswith("//"):
+        ep2 = urlparse(f"https:{endpoint}")
+        path = ep2.path or "/"
+        return urlunparse((base.scheme or "https", base.netloc, path, "", ep2.query, ""))
+    return urljoin(base_str, endpoint)
+
+
 class HttpClient:
     """HTTP client with retry logic and error handling."""
 
@@ -28,26 +59,7 @@ class HttpClient:
         self.backoff_factor = backoff_factor
 
     def _build_url(self, endpoint: str) -> str:
-        base = urlparse(self.api_url)
-        ep = urlparse(endpoint)
-
-        # Absolute or protocol-relative (has netloc)
-        if ep.netloc:
-            # Different host: keep path/query but force base host/scheme (no token leakage)
-            path = ep.path or "/"
-            if (ep.hostname or "") != (base.hostname or ""):
-                return urlunparse((base.scheme or "https", base.netloc, path, "", ep.query, ""))
-            # Same host: normalize scheme to base
-            return urlunparse((base.scheme or "https", base.netloc, path, "", ep.query, ""))
-
-        # Relative (including leading slash or not)
-        base_str = self.api_url if self.api_url.endswith("/") else f"{self.api_url}/"
-        # Guard protocol-relative like //host/path slipping through as “relative”
-        if endpoint.startswith("//"):
-            ep2 = urlparse(f"https:{endpoint}")
-            path = ep2.path or "/"
-            return urlunparse((base.scheme or "https", base.netloc, path, "", ep2.query, ""))
-        return urljoin(base_str, endpoint)
+        return build_url(self.api_url, endpoint)
     
     def _prepare_headers(
         self,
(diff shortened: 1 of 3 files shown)


toolshedlabs-hash:
Reporter here. I read the diff and it covers what I filed. The five async verbs were passing the
endpoint to httpx untouched, so an absolute cross-host `next` URL went out as-is with the
Authorization header on the client. Routing them through the shared builder pins them back to the
configured host first.

Extracting `build_url` instead of copying the sync logic is the part I would not have thought to
ask for. It means async now inherits the protocol-relative and userinfo handling the sync client
already had, neither of which I mentioned in the issue, and the sync/async parity test is what
stops the two drifting apart again later.

One thing I did not check and cannot speak to: whether anything outside the v2 http clients builds
a URL for these endpoints independently.

PR #4203, opened 2026-08-01
fix(api): strip NUL bytes before persisting scrape results to Postgres

### LINKED ISSUE #4113: [Self-Host] nuq-worker crash-loops the api container when scraped content contains NUL bytes (Postgres 22P05)
**Environment**: self-hosted, `ghcr.io/firecrawl/firecrawl:latest` (docker-compose with NUQ Postgres + RabbitMQ), Node 22 image.

**Repro**: crawl or scrape a PDF whose parsed text contains `\u0000` (NUL). Example class: certain PDFs exported from Google Sheets ("Google Kalkylark") parse via pdf-parse to text beginning with NUL bytes.

**What happens**: `nuq-worker` writes the scrape `returnvalue` to Postgres in `jobFinish` (`apps/api/src/services/worker/nuq.ts`, the `UPDATE ... SET returnvalue = $3` query). Postgres rejects JSON containing `\u0000`:

```
severity: 'ERROR',
code: '22P05',
detail: '\u0000 cannot be converted to text.',
where: 'JSON data, line 1: {"markdown":"\n\n\u0000...',
routine: 'json_errsave_error'
```

The exception is unhandled, the worker exits 1, and the harness treats it as fatal (`A service has terminated unexpectedly ... nuq-worker-1 failed with exit code 1`), tearing down the whole api container. On restart the poison job is still queued, so the container crash-loops indefinitely. One bad PDF takes down the entire instance; recovery requires purging `nuq.queue_scrape`.

**Suggested fix** (either or both):

1. Sanitize NUL from string values before the insert, e.g.

   ```js
   returnvalue == null
     ? returnvalue
     : JSON.parse(JSON.stringify(returnvalue), (k, v) =>
         typeof v === "string" ? v.replaceAll("\u0000", "") : v)
   ```

2. Wrap the `jobFinish`/`jobFail` writes in try/catch and fail the job instead of crashing the worker, so a poison job can never crash-loop the container.

**Workaround we run in production**: a derived image that seds exactly fix 1 into `dist/src/services/worker/nuq.js`. Happy to PR the proper fix if you point me at the preferred approach.

nitishagar:
## What

Fixes a worker crash-loop on self-hosted deployments where scraped content contained a NUL byte (`\u0000`). Postgres rejects embedded NUL in JSON/JSONB with error `22P05` ("… cannot be converted to text"), the throw happened inside `jobFinish`, so the job never reached a terminal state and the worker kept re-running the same poison job after every restart — taking the whole `api` container down each time.

Fixes #4113.

## Root cause

`jobFinish` (`apps/api/src/services/worker/nuq.ts`) binds the scraped `returnvalue` straight to the `returnvalue` JSONB column via an `UPDATE … returnvalue = $3` query. `jobFail` does the same for `failedreason`. When the scraped content (or a stringified error) contained `\u0000`, Postgres threw `22P05`, the error propagated out of `jobFinish`/`jobFail`, and the worker process crashed — only to pick up the same job again on restart.

## Fix

Added a small `stripNulBytes` helper that recursively removes `\u0000` from strings — including object keys and values inside nested arrays/objects, since `returnvalue` is an arbitrarily-deep `Document` (markdown, html, extract, json, …). It is called right before the `UPDATE` in both `jobFinish` and `jobFail`.

A few properties worth calling out:

- **Loss-free for normal content.** Only `\u0000` is touched — no whitespace collapsing, no trimming, other control chars pass through. A NUL-free result is byte-identical to before, so existing jobs are unaffected. (I did look at the existing `sanitizeStrings` in the extract pipeline, but it also collapses whitespace and trims, which is too aggressive here.)
- **Recursion is depth-capped.** A naive recursion would itself overflow the stack on a pathologically deep object — which would reintroduce the same worker-crash class this PR is fixing. The cap sits far above any real `Document` depth and well below the stack limit, so only exotic, non-real inputs are affected.
- **Scope.** Only the two terminal-state writes that carry untrusted scraped/error content are sanitized (`returnvalue`, `failedreason`). The job-input `data` column is API-request-sourced and is not the crash path reported here.

The crawl-finished queue, which routes through the same `jobFinish`/`jobFail`, gets the same protection automatically.

## Tests

New unit test `apps/api/src/__tests__/nuq-sanitize.test.ts` (pure, no DB needed) covering:

- NUL stripped from nested objects/arrays, including object keys
- loss-free no-op for NUL-free input (tabs/newlines/control chars preserved)
- idempotent (apply twice === apply once)
- plain string branch (the `failedreason` shape)
- primitives and `null` passed through untouched
- pathologically deep nesting does not throw (regression guard for the depth cap)

```
✓ src/__tests__/nuq-sanitize.test.ts (6 tests)
```

## Out of scope

- The FoundationDB queue backend (`nuq-fdb`) is unaffected by the Postgres `22P05` error and is intentionally untouched.
- The job-input `data`-column writes (API-sourced payload) are out of scope.

---
<a href="https://app.blacksmith.sh/firecrawl/codesmith/firecrawl/pr/4203"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788189400&installation_model_id=11126&pr_number=4203&repository=firecrawl%2Ffirecrawl&return_to=https%3A%2F%2Fgithub.com%2Ffirecrawl%2Ffirecrawl%2Fpull%2F4203&signature=5db14f5176e3bd63d40b51e18b0b82cef92b9f62c9b513ff60cac86de32fe983"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Prevented nuq-worker crash loops by stripping NUL bytes from scraped results and failure reasons before writing to Postgres JSONB. Jobs now reach terminal states and self-hosted deployments won’t crash on NUL-containing content.

- **Bug Fixes**
  - Added a `stripNulBytes` helper that removes `\u0000` from strings and object keys/values (arrays/objects) at any depth without altering other characters.
  - Applied in `jobFinish` (returnvalue) and `jobFail` (failedreason) to avoid Postgres `22P05` errors.
  - Switched to an iterative walk (no recursion) to prevent stack overflows on deep objects; output uses null-prototype objects so `"__proto__"` keys are preserved as own properties.
  - Added unit tests for nested structures, idempotency, primitives/null, plain strings, 50k-deep nesting, and `"__proto__"` preservation.

<sup>Written for commit c4c93a5d9834edf102373cb8293515fd10068f4d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/firecrawl/pull/4203?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/api/src/__tests__/nuq-sanitize.test.ts
+++ b/apps/api/src/__tests__/nuq-sanitize.test.ts
@@ -1,0 +1,81 @@
+import { stripNulBytes } from "../services/worker/nuq";
+
+// Regression coverage for firecrawl #4113: Postgres JSON rejects embedded NUL
+// (\u0000) with error 22P05, which previously threw out of jobFinish and
+// crash-looped the nuq-worker. stripNulBytes is applied to returnvalue
+// (jobFinish) and failedReason (jobFail) before the UPDATE.
+describe("stripNulBytes", () => {
+  it("removes NUL bytes from nested objects and arrays (returnvalue shape)", () => {
+    const input = {
+      a: "x\u0000y",
+      nested: { b: ["1\u00002", { c: "\u0000z" }] },
+      "k\u0000ey": 1,
+    };
+    const expected = {
+      a: "xy",
+      nested: { b: ["12", { c: "z" }] },
+      key: 1,
+    };
+    expect(stripNulBytes(input)).toEqual(expected);
+  });
+
+  it("is loss-free for NUL-free input (no whitespace collapse, control chars preserved)", () => {
+    const input = {
+      a: "normal",
+      b: 1,
+      c: null,
+      d: ["keep", { e: "tabs\tand\nnewlines" }],
+    };
+    expect(stripNulBytes(input)).toEqual(input);
+  });
+
+  it("is idempotent (apply twice === apply once)", () => {
+    const x = { a: "x\u0000y", nested: { b: ["1\u00002", { c: "\u0000z" }] } };
+    expect(stripNulBytes(stripNulBytes(x))).toEqual(stripNulBytes(x));
+  });
+
+  it("strips NUL from a plain string (failedReason shape)", () => {
+    expect(stripNulBytes("a\u0000b")).toBe("ab");
+  });
+
+  it("passes primitives and null through untouched", () => {
+    expect(stripNulBytes(null as null)).toBeNull();
+    expect(stripNulBytes(42)).toBe(42);
+    expect(stripNulBytes(true)).toBe(true);
+    expect(stripNulBytes(undefined as undefined)).toBeUndefined();
+  });
+
+  it("strips NUL at any depth and does not overflow the stack on pathologically deep nesting (iterative walk — a naive recursion would either cap out and leave NUL unsanitized, or throw RangeError, both reintroducing the #4113 crash)", () => {
+    // 50,000 levels deep — far beyond a recursive call-stack budget (~3.4k) and
+    // beyond any prior depth cap. The leaf carries a NUL that must still be
+    // stripped, because the value serializes fine for pg and would otherwise
+    // trigger 22P05 at the UPDATE.
+    let deep: Record<string, unknown> = { leaf: "has\u0000nul" };
+    for (let i = 0; i < 50_000; i++) {
+      deep = { a: deep };
+    }
+    // Must not throw...
+    let result: Record<string, unknown> = {} as Record<string, unknown>;
+    expect(() => {
+      result = stripNulBytes(deep) as Record<string, unknown>;
+    }).not.toThrow();
+    // ...and the deep NUL must actually be removed.
+    let leaf: unknown = result;
+    for (let i = 0; i < 50_000; i++) {
+      leaf = (leaf as Record<string, unknown>).a;
+    }
+    expect((leaf as Record<string, unknown>).leaf).toBe("hasnul");
+  });
+
+  it("preserves an input __proto__ key as a normal own property (no NUL involved; a plain output object would silently drop it)", () => {
+    // JSON.parse yields __proto__ as an own enumerable property; the sanitizer
+    // must not lose it (it has no NUL and would otherwise round-trip fine).
+    const input = JSON.parse('{"__proto__":"kept","normal":"yes"}');
+    const result = stripNulBytes(input) as Record<string, unknown>;
+    expect(Object.keys(result).sort()).toEqual(["__proto__", "normal"]);
+    expect(result.__proto__).toBe("kept");
+    expect(result.normal).toBe("yes");
+    // And it must not pollute Object.prototype globally.
+    expect(({} as Record<string, unknown>).kept).toBeUndefined();
+  });
+});
--- a/apps/api/src/services/worker/nuq.ts
+++ b/apps/api/src/services/worker/nuq.ts
@@ -10,6 +10,86 @@ import { nuqRedis } from "./redis";
 
 // === Basics
 
+/**
+ * Strips NUL bytes (\u0000) from any value before it is persisted to Postgres.
+ *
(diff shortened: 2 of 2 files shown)


nitishagar:
Took a look at all three findings — **P1 and P2 are valid and fixed** in the updated commit (`c4c93a5`); **P3 is acknowledged but intentionally left as-is**.

**P1 — depth cap bypassed NUL protection (valid, fixed).** The earlier `depth > 250` guard returned the remaining subtree unsanitized, so a NUL byte deeper than 250 would still reach Postgres and trigger 22P05 — the exact crash this PR exists to prevent, just relocated to a rarer depth. Rewrote `stripNulBytes` to walk the structure **iteratively with an explicit stack** instead of recursing, so depth is bounded by the heap rather than the call stack. There's no recursion ceiling to cap, and every nested string/key is sanitized regardless of depth. Added a test nesting 50,000 levels deep with a NUL at the leaf, asserting it both doesn't throw **and** actually strips the byte.

**P2 — `__proto__` key dropped (valid, fixed).** Bracket-assigning `out["__proto__"] = …` onto a plain `{}` invokes the prototype setter and silently drops the key, even though it has no NUL — a data-loss regression vs. the pre-fix `JSON.parse` behavior (which keeps it as an own enumerable property). Output objects are now built with `Object.create(null)`, so `__proto__`/`constructor` survive as normal own properties. Verified `JSON.stringify` round-trips identically to today and that `Object.prototype` is not polluted globally. Added a test asserting the key is preserved.

**P3 — NUL-in-key collision merges two keys (acknowledged, not fixed).** This needs two keys differing *only* by NUL bytes, which isn't realistic for scraped HTML/JSON content. More importantly, without this PR the *entire* payload is rejected by Postgres (strictly worse than merging two near-identical keys), so the current behavior is the lesser evil. Detecting collisions would add real complexity for a case that doesn't occur in practice — happy to revisit if there's a concrete scenario in mind.

All 7 unit tests green; `tsc` clean for the changed files.

ben0112:
Validated this PR against both a real PostgreSQL instance and a live self-hosted deployment.

Test results:

- Existing sanitizer suite: 7/7 passed.
- TypeScript build: passed.
- Added a PostgreSQL integration test that calls jobFinish with NUL bytes in nested values and an object key, then reads the JSONB row back and verifies that actual NULs are removed while literal \u0000 text is preserved: [2024e954](https://github.com/ben0112/firecrawl/commit/2024e95440e82610ff236272864582445bfbf3cb).
- Live regression: re-scraped one PDF from the production incident. It returned HTTP 200 and 114,538 bytes of Markdown. Its raw response JSON contained 148 escaped \u0000 sequences, so the test definitely exercised NUL-producing content. The API remained healthy, restart count stayed at 0, and there were no new PostgreSQL 22P05 or worker-exit errors.

The original incident on official build b1371d2 produced repeated API restarts; there was no OOM. This PR fixes that failure mode in our deployment.

PR #4205, opened 2026-08-02
fix(python): reject unknown keyword arguments in the async v2 client

ckarnell:
The sync v2 client lists every option explicitly, so a misspelled one raises `TypeError` at the call site. The async client takes `**kwargs` and passes them straight to a pydantic model, and pydantic ignores unknown fields by default, so the same typo silently produced a request with default settings.

```python
client.scrape("https://example.com", tiemout=5000)               # TypeError
await async_client.scrape("https://example.com", tiemout=5000)   # runs with no timeout
```

Same on `search`, `start_crawl`, `crawl`, `start_batch_scrape` and `batch_scrape`. The one that worried me is `skip_tls_verification`, where a typo quietly falls back to the default.

The fix derives each method's accepted keywords from the sync signature instead of restating them, so the two cannot drift when an option is added to one side only. `client_async.py` already says it mirrors the regular client surface, and this makes that true for unknown keys too.

I checked that typos now raise on all six methods, and that real options still reach the request, including `ignore_sitemap` and the scrape options `start_crawl` folds in. Added `tests/test_async_unknown_kwargs.py` with 10 tests. Five of them fail without the change, which is why they are worth having.

The suite was 13 failed, 38 passed before this change and 13 failed, 48 passed after. Those 13 pre-date it and I have not touched them.

Not a security hole. It only bites when the caller has already made a mistake, and what breaks is one request's settings.

I have not looked at the v1 client, or at whether any caller passes extra keys on purpose. If firecrawl wants those tolerated, an allow-list is the smaller change.

---
<a href="https://app.blacksmith.sh/firecrawl/codesmith/firecrawl/pr/4205"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788251880&installation_model_id=11126&pr_number=4205&repository=firecrawl%2Ffirecrawl&return_to=https%3A%2F%2Fgithub.com%2Ffirecrawl%2Ffirecrawl%2Fpull%2F4205&signature=d1771ef3c457c3c13d67e725579f9ce4384ab58532efea9d32276518dd85e854"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Reject unknown keyword arguments in the `firecrawl.v2` async client to match the sync client. Typos now raise `TypeError` instead of being silently ignored.

- **Bug Fixes**
  - Validate keywords for `scrape`, `search`, `start_crawl`, `crawl`, `start_batch_scrape`, `batch_scrape` in `AsyncFirecrawlClient`.
  - Derive allowed keywords from `FirecrawlClient` method signatures (plus `_SCRAPE_OPTION_KEYS`, `ignore_sitemap`, and polling args) to prevent drift.
  - Fold flat scrape keywords into `ScrapeOptions` in `start_batch_scrape`/`batch_scrape`; explicit `options` wins over flat keys.
  - Add tests in `apps/python-sdk/tests/test_async_unknown_kwargs.py` covering error cases, valid options, and batch option folding.

<sup>Written for commit 4cb42e49e1542fefad53e4f1bace9ddf0c3fd3d5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/firecrawl/pull/4205?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/python-sdk/firecrawl/v2/client_async.py
+++ b/apps/python-sdk/firecrawl/v2/client_async.py
@@ -64,7 +64,45 @@ from .methods.aio import browser as async_browser  # type: ignore[attr-defined]
 from .methods.aio import monitor as async_monitor  # type: ignore[attr-defined]
 from .methods.aio import research as async_research  # type: ignore[attr-defined]
 
-from .client import _SCRAPE_OPTION_KEYS
+from inspect import signature as _signature
+
+from .client import _SCRAPE_OPTION_KEYS, FirecrawlClient as _SyncClient
+
+
+def _sync_kwargs(method) -> frozenset:
+    """The keyword names the sync client accepts for the same call.
+
+    This module's docstring says it mirrors the regular client surface, so the set of
+    accepted options is defined as the sync signature rather than restated here. It
+    cannot drift when a new option is added to one side only.
+    """
+    return frozenset(
+        name for name in _signature(method).parameters if name != "self"
+    )
+
+
+def _reject_unknown(method_name: str, kwargs: dict, allowed: frozenset) -> None:
+    """Raise on an unknown keyword, the way the sync client already does.
+
+    The sync methods list every option explicitly, so a misspelled one is a TypeError at
+    the call site. The async methods take **kwargs and hand them to a pydantic model,
+    and pydantic ignores unknown fields by default, so the same typo silently produced a
+    request with default settings instead.
+    """
+    unknown = sorted(set(kwargs) - allowed)
+    if unknown:
+        raise TypeError(
+            f"AsyncFirecrawlClient.{method_name}() got an unexpected keyword argument "
+            f"{unknown[0]!r}"
+        )
+
+
+_SCRAPE_KWARGS = _sync_kwargs(_SyncClient.scrape)
+_SEARCH_KWARGS = _sync_kwargs(_SyncClient.search)
+_START_CRAWL_KWARGS = _sync_kwargs(_SyncClient.start_crawl) | frozenset(_SCRAPE_OPTION_KEYS) | {"ignore_sitemap"}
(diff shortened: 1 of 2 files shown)


ckarnell:
Good catch on the batch path, fixed in 4cb42e4. `_prepare` reads `options` plus a fixed key list, so loose scrape keywords never reached the payload and `start_batch_scrape(urls, formats=["markdown"])` came back with the default format. Async batch now folds them into `ScrapeOptions` the way the sync client and `start_crawl` already do, with an explicit `options=` still winning.

One correction for anyone reading later. That drop is not new here. Before this PR the async client took `**kwargs` and passed them through, so the same call dropped `formats` in exactly the same way. What changed is that the keywords are now checked, which is what surfaced it.

Two tests cover it. Both fail without the change.

PR #4209, opened 2026-08-03
docs: align self-hosting guidance with the public quickstart

DanielleWashington:
**Why**

The product repository contains the revision-specific sources that self-hosters and contributors need after the public guide gets them onto the right path. Those references should use the same decision-oriented structure and Firecrawl voice, while being explicit about which examples are evaluation starting points and which production decisions remain with the operator.

**Summary**

- Keep `SELF_HOST.md` as the source map for the checked-out revision and add handoffs to contributor setup, local MCP, and the Firecrawl CLI.
- Replace the duplicated setup in `CONTRIBUTING.md` with the current source-owned local development and test workflow.
- Rework the raw Kubernetes README around choosing the manifest path, reviewing authentication and image-pull assumptions, applying one namespace, verifying one scrape, and defining a production design.
- Rework the Helm README around image trust, values review, render-before-install, release verification, production decisions, multi-architecture builds, packaging, and cleanup.
- Surface existing example limitations directly: database authentication is disabled in the raw manifests, `docker-registry-secret` must be created or removed, Helm defaults use third-party `latest` images, and resource settings are disabled by default.
- Link the public documentation journey: https://github.com/firecrawl/firecrawl-docs/pull/1194.

**Test plan**

- [x] Contributor commands match `apps/api/package.json`, the API harness, Node.js 22, and pnpm `11.4.0`.
- [x] Raw-manifest names, namespace usage, image-pull secret references, authentication setting, readiness path, and API service names match the checked-in YAML.
- [x] Helm image, resource, component, fullname, service, deployment, and overlay guidance matches the checked-in chart sources.
- [x] Repository-relative and public documentation links resolve to the intended sources.
- [x] `git diff --check` passes and only the four intended Markdown files change.
- [x] Helm `v4.2.3` passes `helm lint . -f values.yaml -f overlays/prod/values.yaml` with only the informational icon recommendation, and `helm template` with the same values renders all resources.
- [x] Pull-request CI completes without a change-related regression.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,140 +1,71 @@
-# Contributors guide:
+# Contributing to Firecrawl
 
-Welcome to [Firecrawl](https://firecrawl.dev) 🔥! Here are some instructions on how to get the project locally, so you can run it on your own (and contribute)
+Thanks for helping make Firecrawl better. Keep each change focused, prove the behavior you changed, and make the pull request easy to review.
 
-If you're contributing, note that the process is similar to other open source repos i.e. (fork firecrawl, make changes, run tests, PR). If you have any questions, and would like help getting on board, reach out to help@firecrawl.com for more or submit an issue!
+## Choose the right workflow
 
-## Running the project locally
+| If you want to | Start here |
+| --- | --- |
+| Change the API, workers, or tests | [Run Firecrawl locally for development](https://docs.firecrawl.dev/contributing/guide) |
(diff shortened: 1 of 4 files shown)


hmishra2250 (approved):
Checked the manifests and Helm values against source. All accurate. The namespace fix also corrects the old README, which port forwarded from `-n dev` without creating it.

Two notes. Prerequisites now defer to the public guide, which is missing Go, so this should land with firecrawl-docs#1194 where I left that fix. And line 50, `pnpm harness pnpm exec vitest run <file>`, can race a cold API, because the harness only waits for the API port when the command starts with `test:snips`.

PR #4219, opened 2026-08-03
feat(api): expose PDF page markdown

### LINKED ISSUE #3800: [Feat] Page-level Markdown output for parsed PDFs
**Problem Description**
Its difficult to use `/v2/parse` for PDF reader, document Q&A, citation, and review workflows because parsed PDFs are returned as one document-level `markdown` string. The response can include `metadata.numPages`, but the Markdown itself does not preserve reliable physical page boundaries. For apps that show the original PDF beside extracted text or model answers, we need to map extracted Markdown back to PDF page numbers. 

**Proposed Feature**
Add an option to return page-level Markdown for PDFs.

For example, `/v2/parse` and PDF `/v2/scrape` could support an option like `splitByPage`, `includePages`, or `includePageMarkers`.

The response could include:

```json
{
  "success": true,
  "data": {
    "markdown": "...whole document markdown...",
    "pages": [
      { "pageNumber": 1, "markdown": "..." },
      { "pageNumber": 2, "markdown": "..." }
    ],
    "metadata": {
      "numPages": 2
    }
  }
}
```

Alternatively, Firecrawl could include stable page markers in the returned Markdown:

```md
<!-- Page 1 -->
...

<!-- Page 2 -->
...
```

**Alternatives Considered**
One workaround is to split the PDF into one-page PDFs before sending it to Firecrawl, then parse each page separately. This is reliable for page boundaries, but it is not ideal because it turns one parse operation into many requests and pushes orchestration, retries, ordering, and rate-limit behavior onto every caller. Another is to fork and self host?

**Implementation Suggestions**
A parser option could be added to PDF parsing, for example:

```json
{
  "formats": ["markdown"],
  "parsers": [{ "type": "pdf", "mode": "auto", "splitByPage": true }]
}
```

or:

```json
{
  "formats": ["markdown"],
  "parsers": [{ "type": "pdf", "mode": "auto" }],
  "pdfOptions": {
    "includePages": true
  }
}
```

The page-level output could be additive so existing callers continue using `data.markdown` unchanged, while callers that opt in receive `data.pages`.

From looking at the repo, Firecrawl already has page-count handling through `pdfMetadata.numPages`, and Fire-PDF reports `pages_processed`. I also noticed closed PR #2966 mentioned page markers and page-level merge utilities for mixed PDFs. Something similar exposed as a public API option would solve this cleanly.

**Use Case**
In a workspace/document application, a user uploads a PDF and asks questions about it. The app needs to provide answers with page citations and let the user jump from an answer to the relevant PDF page.

Example:

- User uploads a 40-page research paper.
- Firecrawl parses it into Markdown.
- The app stores extracted text by page.
- The model answers: "The manipulation taxonomy is introduced on page 7."
- The UI opens the PDF viewer directly to page 7 and can show the extracted Markdown context for that page.

This would also help with:

- page-aware RAG chunking
- citation generation
- side-by-side PDF/text review
- audits where extracted content needs to be traced back to the source page
- avoiding unnecessary per-page parse requests

abimaelmartell:
## Summary

Adds an opt-in `pageMarkdown` PDF parser option for `/v2/scrape` and `/v2/parse`, returning additive `data.pages` entries while preserving `data.markdown`. Page-aware requests are forced through FirePDF, consume its `pages: [{ page, markdown }]` payload, use versioned capability-specific cache sidecars, and bypass document-only URL index reads and writes so legacy cached artifacts cannot satisfy them. SDK surface changes are intentionally out of scope for this PR.

Builds on the merged firecrawl/fire-pdf#515 and closes #3800.


abimaelmartell:
@cubic-dev-ai

abimaelmartell:
@cubic-dev-ai

abimaelmartell:
@cubic-dev-ai

abimaelmartell:
@cubic-dev-ai

PR #4221, opened 2026-08-04
fix(search): recover path-scoped site: queries that return no results

abimaelmartell:
Closing — superseding with a different approach.


abimaelmartell:
Superseded — will be handled with a different approach.

PR #4222, opened 2026-08-04
fix(python-sdk): raise FirecrawlError on failed/cancelled batch/crawl waits

### LINKED ISSUE #4212: [Bug] Python SDK: batch_scrape/crawl return `failed` jobs silently — the documented FirecrawlError is never raised
## Summary

`batch_scrape()` and `crawl()` return `failed` and `cancelled` jobs to the caller exactly like successful ones — no exception is raised, and the returned model carries no error field. The docstrings on those code paths explicitly promise the opposite ("Raises: `FirecrawlError`: If the job fails"), so callers written against the documented contract silently treat a failed job as an empty-but-successful result.

Version: found on `firecrawl-py` 4.13.0, **verified still present in 4.34.0** (current latest). `wait_for_batch_completion` is byte-identical between the two releases; the five misleading docstrings and the missing error field are all unchanged. Line numbers below refer to 4.34.0. Python 3.10.

## The bug

`wait_for_batch_completion` treats all three terminal states identically:

```python
# firecrawl/v2/methods/batch.py:297-334
def wait_for_batch_completion(client, job_id, poll_interval=2, timeout=None) -> BatchScrapeJob:
    """
    ...
    Raises:
        FirecrawlError: If the job fails or timeout is reached   # <-- never happens
        TimeoutError: If timeout is reached
    """
    while True:
        status_job = get_batch_scrape_status(client, job_id)
        if status_job.status in ["completed", "failed", "cancelled"]:
            return status_job          # a failed job is returned like a successful one
        ...
```

`wait_for_crawl_completion` has the same shape and the same false docstring:

```python
# firecrawl/v2/methods/crawl.py:373-410
    Raises:
        Exception: If the job fails     # <-- never happens
        TimeoutError: If timeout is reached
...
        if crawl_job.status in ["completed", "failed", "cancelled"]:
            return crawl_job
```

Both are reachable through the public client: `Firecrawl.batch_scrape()` → `batch_module.batch_scrape()` → `wait_for_batch_completion`, and `Firecrawl.crawl()` → `wait_for_crawl_completion`.

### Docstrings that promise a raise that does not exist

| Location (4.34.0) | Claim |
|---|---|
| `v2/methods/batch.py:316` | `FirecrawlError: If the job fails or timeout is reached` |
| `v2/methods/batch.py:366` | `FirecrawlError: If the batch scrape fails to start or complete` |
| `v2/methods/crawl.py:395` | `Exception: If the job fails` |
| `v2/methods/crawl.py:443` | `Exception: If the crawl fails to start or complete` |
| `v2/client.py:554` | `Exception: If the crawl fails to start or complete` |

The async client is internally consistent by comparison — `AsyncFirecrawl.wait_crawl` (`v2/client_async.py:266-297`) documents only `TimeoutError` and lists the three terminal states explicitly. So the sync surface is documented as raising while the async surface, with identical behavior, is documented as not raising.

### There is no error detail on the returned model either

Even a caller who knows to check `.status` cannot find out *why* the job failed. Neither model carries an error field:

```python
# v2/types.py:1034-1043
class BatchScrapeJob(BaseModel):
    status: Literal["scraping", "completed", "failed", "cancelled"]
    completed: int
    total: int
    credits_used: Optional[int] = None
    expires_at: Optional[datetime] = None
    next: Optional[str] = None
    data: List[Document] = []
```

`CrawlJob` (`v2/types.py:916-925`) is the same. Getting a reason requires a second round trip to `get_batch_scrape_errors()` / `get_crawl_errors()`, which nothing in the return path signals you should make.

## Why this is worse than a normal missing exception

A failed job comes back with `data: []` (or partially populated). The natural consumer pattern —

```python
response = firecrawl.batch_scrape(urls, formats=["markdown"], wait_timeout=600)
for doc in response.data:
    ...
```

— cannot distinguish "the job failed" from "these pages had no content". A failure degrades into an empty result set and flows downstream as legitimate-looking data.

It also defeats retry logic. A caller wrapping the SDK in the usual retry-on-exception decorator gets no retry at all on a failed job, because nothing is raised — the failure consumes the full poll timeout once and then returns quietly as a success.

## Reproduction

No API key or flaky site needed — stubbing the status call is enough, since the defect is in the wait loop itself. This calls the real `wait_for_batch_completion` / `wait_for_crawl_completion`:

```python
from firecrawl.v2.methods import batch, crawl
from firecrawl.v2.types import BatchScrapeJob, CrawlJob

batch.get_batch_scrape_status = lambda *a, **k: BatchScrapeJob(
    status="failed", completed=0, total=3, data=[]
)
crawl.get_crawl_status = lambda *a, **k: CrawlJob(
    status="failed", completed=0, total=3, data=[]
)

job = batch.wait_for_batch_completion(client=None, job_id="stub")
print(f"batch  -> returned status={job.status!r} data={job.data} (no exception)")

job = crawl.wait_for_crawl_completion(client=None, job_id="stub")
print(f"crawl  -> returned status={job.status!r} data={job.data} (no exception)")
```

Output on **4.34.0** (identical on 4.13.0):

```
batch  -> returned status='failed' data=[] (no exception)
crawl  -> returned status='failed' data=[] (no exception)
```

Both return normally. The same thing happens end-to-end through `Firecrawl.batch_scrape(...)` / `Firecrawl.crawl(...)` against the live API when a job genuinely fails — the stub just makes it deterministic.

## Suggested fixes

Either of these resolves it; the first matches the documented contract.

1. **Raise on non-`completed` terminal states**, so the code matches the docstrings:

   ```python
   if status_job.status == "completed":
       return status_job
   if status_job.status in ("failed", "cancelled"):
       raise FirecrawlError(f"Batch scrape job {job_id} ended with status {status_job.status}")
   ```

   This is a breaking change for anyone relying on today's behavior, so it probably wants a major bump or an opt-out flag (e.g. `raise_on_failure: bool = True`).

2. **Keep returning the job, but fix the docs and surface the reason** — remove the `Raises: FirecrawlError: If the job fails` lines from all five sites above, and add an `error: Optional[str]` field to `CrawlJob` and `BatchScrapeJob` populated from the status response so callers can act on `.status` / `.error` without a second request.

Whichever direction you take, the sync docstrings and the async ones should end up saying the same thing.

## Related

- #4107 — separate open bug in the same `wait_for_batch_completion` function (unnecessary pagination during polling).
- #2190 — added `"cancelled"` to the status `Literal`; that fix made `cancelled` a valid terminal value but did not address how the wait loops treat it.

santhiprakash:
## Problem

`wait_for_batch_completion()` and `wait_for_crawl_completion()` treated `failed` and `cancelled` terminal states the same as `completed`, returning the job with empty `data` and no error. Sync docstrings promised `FirecrawlError` on failure, but nothing was raised — so callers could not distinguish a failed job from an empty success (fixes #4212).

## Triage / Root cause

In `batch.py` and `crawl.py`, the wait loops returned on any of `completed`, `failed`, or `cancelled`. The async client had the same behavior but documented only `TimeoutError`, so sync and async docs were inconsistent.

## Fix

- Raise `FirecrawlError` when a wait loop sees `failed` or `cancelled`; return normally only for `completed`.
- Apply the same behavior in `AsyncFirecrawl.wait_crawl` and `wait_batch_scrape`.
- Align docstrings on sync `client.py`, `batch.py`, `crawl.py`, and async `client_async.py`.

## Verification

```bash
cd apps/python-sdk
pytest firecrawl/__tests__/unit/v2/methods/test_wait_terminal_failure.py -q
pytest firecrawl/__tests__/unit -q --ignore=firecrawl/__tests__/unit/v2/watcher
```

Results: 6 new regression tests pass; full unit suite 399 passed.

## Notes / Risks

- This matches v1 SDK behavior (v1 raised on failed/stopped crawl status) and restores the documented sync contract. Callers that previously checked `.status` after `batch_scrape()` / `crawl()` without handling exceptions will need to catch `FirecrawlError` instead.
- Error detail still requires `get_batch_scrape_errors()` / `get_crawl_errors()` — no model field added in this PR.

---
<a href="https://app.blacksmith.sh/firecrawl/codesmith/firecrawl/pr/4222"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788407202&installation_model_id=11126&pr_number=4222&repository=firecrawl%2Ffirecrawl&return_to=https%3A%2F%2Fgithub.com%2Ffirecrawl%2Ffirecrawl%2Fpull%2F4222&signature=20184d86d5ccdbfa5dff09ab461f4ff61a411f9a295f03322d9a09cb231fcdec"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Raise `FirecrawlError` when `wait_for_batch_completion()` and `wait_for_crawl_completion()` (and async equivalents) see `failed` or `cancelled`, so callers can properly handle failures. Aligns sync/async behavior and docstrings; adds targeted tests.

- **Bug Fixes**
  - Raise `FirecrawlError` on `failed`/`cancelled` in `wait_for_crawl_completion`, `wait_for_batch_completion`, `AsyncFirecrawl.wait_crawl`, and `AsyncFirecrawl.wait_batch_scrape`.
  - Update docstrings in `v2/client.py` and `v2/client_async.py` to reflect the behavior.
  - Add unit tests for failure and success paths in `test_wait_terminal_failure.py`.

- **Migration**
  - Wrap calls to `crawl()`/`batch_scrape()` and `wait_*` helpers in a `try/except` for `FirecrawlError`.
  - Use `get_crawl_errors()` or `get_batch_scrape_errors()` to fetch error details after catching.

<sup>Written for commit 16762854c603dbd65dd935d72217b333d94bf4f0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/firecrawl/pull/4222?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/python-sdk/firecrawl/__tests__/unit/v2/methods/test_wait_terminal_failure.py
+++ b/apps/python-sdk/firecrawl/__tests__/unit/v2/methods/test_wait_terminal_failure.py
@@ -1,0 +1,53 @@
+import pytest
+
+from firecrawl.v2.methods import batch, crawl
+from firecrawl.v2.types import BatchScrapeJob, CrawlJob
+from firecrawl.v2.utils.error_handler import FirecrawlError
+
+
+class TestWaitTerminalFailure:
+    def test_wait_for_batch_completion_raises_on_failed(self):
+        batch.get_batch_scrape_status = lambda *args, **kwargs: BatchScrapeJob(
+            status="failed", completed=0, total=3, data=[]
+        )
+
+        with pytest.raises(FirecrawlError, match="Batch scrape job stub ended with status 'failed'"):
+            batch.wait_for_batch_completion(client=None, job_id="stub")
+
+    def test_wait_for_batch_completion_raises_on_cancelled(self):
+        batch.get_batch_scrape_status = lambda *args, **kwargs: BatchScrapeJob(
+            status="cancelled", completed=0, total=3, data=[]
+        )
+
+        with pytest.raises(FirecrawlError, match="Batch scrape job stub ended with status 'cancelled'"):
+            batch.wait_for_batch_completion(client=None, job_id="stub")
+
+    def test_wait_for_batch_completion_returns_completed(self):
+        expected = BatchScrapeJob(status="completed", completed=3, total=3, data=[])
+        batch.get_batch_scrape_status = lambda *args, **kwargs: expected
+
+        result = batch.wait_for_batch_completion(client=None, job_id="stub")
+        assert result == expected
+
+    def test_wait_for_crawl_completion_raises_on_failed(self):
+        crawl.get_crawl_status = lambda *args, **kwargs: CrawlJob(
+            status="failed", completed=0, total=3, data=[]
+        )
+
+        with pytest.raises(FirecrawlError, match="Crawl job stub ended with status 'failed'"):
+            crawl.wait_for_crawl_completion(client=None, job_id="stub")
+
+    def test_wait_for_crawl_completion_raises_on_cancelled(self):
+        crawl.get_crawl_status = lambda *args, **kwargs: CrawlJob(
+            status="cancelled", completed=0, total=3, data=[]
+        )
+
+        with pytest.raises(FirecrawlError, match="Crawl job stub ended with status 'cancelled'"):
+            crawl.wait_for_crawl_completion(client=None, job_id="stub")
+
+    def test_wait_for_crawl_completion_returns_completed(self):
+        expected = CrawlJob(status="completed", completed=3, total=3, data=[])
+        crawl.get_crawl_status = lambda *args, **kwargs: expected
+
+        result = crawl.wait_for_crawl_completion(client=None, job_id="stub")
+        assert result == expected
--- a/apps/python-sdk/firecrawl/v2/client.py
+++ b/apps/python-sdk/firecrawl/v2/client.py
@@ -551,7 +551,7 @@ class FirecrawlClient:
 
         Raises:
             ValueError: If request is invalid
-            Exception: If the crawl fails to start or complete
+            FirecrawlError: If the crawl fails to start or complete
             TimeoutError: If timeout is reached
         """
         if scrape_options is None:
@@ -1618,6 +1618,10 @@ class FirecrawlClient:
     ):
         """
         Start a batch scrape job and wait until completion.
+
+        Raises:
+            FirecrawlError: If the batch scrape fails to start or complete
+            TimeoutError: If the wait timeout is reached
         """
         options = ScrapeOptions(
             **{k: v for k, v in dict(
--- a/apps/python-sdk/firecrawl/v2/client_async.py
+++ b/apps/python-sdk/firecrawl/v2/client_async.py
@@ -50,6 +50,7 @@ from .types import (
 )
 from .utils.http_client import HttpClient
 from .utils.http_client_async import AsyncHttpClient
+from .utils.error_handler import FirecrawlError
 
 from .methods.aio import scrape as async_scrape  # type: ignore[attr-defined]
 from .methods.aio import parse as async_parse  # type: ignore[attr-defined]
@@ -284,6 +285,7 @@ class AsyncFirecrawlClient:
             CrawlJob: The final status of the crawl job when it reaches a terminal state.
 
         Raises:
+            FirecrawlError: If the crawl fails or is cancelled
             TimeoutError: If the crawl does not reach a terminal state within the specified timeout.
 
         Terminal states:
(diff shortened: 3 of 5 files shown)


goat82447-ops (approved):

santhiprakash:
Closing: the source files modified by this PR (apps/python-sdk/firecrawl/v2/) have been removed or reorganized on upstream/main. The patch no longer applies.

PR #4228, opened 2026-08-04
feat(api)!: remove the developerBeta entitlement gate

KrisOei:
## Summary

Takes the Developer Index out of private beta by removing the `developerBeta` entitlement flag from the developer search surface. Scoped to that removal.

## What changed

- Removes the entitlement check in front of the developer search endpoint, so it no longer refuses teams that lack the flag.
- `/v2/search` no longer drops a requested `developer` category.
- Removes `developerBeta` from the team flag types.
- Replaces the gate's test with one asserting the endpoint serves the callers the flag previously excluded, across both mounts.

A second commit is comments plus a test constant rename marking `/v2/search/developer` as the canonical path — no behavior change. Happy to split it out if preferred.

## Testing

Unit suites for the touched controllers and libraries pass (41 tests). The end-to-end developer suite requires a live upstream and is not run locally. `npm run build` reports two failures that are also present on the base commit and unrelated to this change.

## Notes

- Roll out alongside the corresponding admin-panel change, this one first.
- A few smaller follow-ups noticed while reviewing are tracked separately.

### DIFF
--- a/apps/api/src/__tests__/snips/v2/developer.test.ts
+++ b/apps/api/src/__tests__/snips/v2/developer.test.ts
@@ -10,9 +10,10 @@ const KEYLESS_ENABLED =
   process.env.KEYLESS_REQUESTS_PER_DAY !== undefined &&
   process.env.KEYLESS_CREDITS_PER_DAY !== undefined;
 
-const CANONICAL_PATH = "/v2/developer/search";
-const ALIAS_PATH = "/v2/search/developer";
-const SERVING_PATHS = [CANONICAL_PATH, ALIAS_PATH];
+const CANONICAL_PATH = "/v2/search/developer";
+/** Compatibility mount kept for published CLI/MCP builds; removed once they ship on CANONICAL_PATH. */
+const LEGACY_PATH = "/v2/developer/search";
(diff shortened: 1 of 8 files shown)


PhantomInTheWire (approved):

KrisOei:
Valid finding, fixed in 1851c27.

The mocked upstream is now Response-shaped, satisfying every member the handler reads on the success path. Both cases — a team without the flag, and a keyless caller — now assert a 200 and that billing ran, across both mounts. Confirmed the old mock did make all four cases take the error path before the change, so the assertions really were passing for the wrong reason.

Test file only; no production code in that commit.

KrisOei:
Valid, fixed in ce8e402.

The keyless scenario now builds its team id with the same helper the auth layer uses, so it is shaped like a real keyless request rather than reusing the authed team id. It asserts the keyless credit-charging call with its arguments, so a regression there fails this test. The authed scenario keeps asserting ordinary team billing with its arguments. Both still assert a 200, across both mounts.

Confirmed the assertion failed against the old plain team id before the change, so the coverage is real rather than incidental. Test file only.

PR #4251, opened 2026-08-06
feat(api): safe param on v2 /search

rafaelsideguide:
## What

Adds an optional `safe: boolean` parameter to the v2 `/search` endpoint. When `true`, explicit content is filtered from search results.

## How

- `safe: z.boolean().optional()` on the v2 search request schema — no default, so omitted requests behave exactly as before
- Flows controller → `executeSearch` → `search()` → the upstream search client; the filter param is only included in the upstream request when `safe` is `true`, so existing behavior and cache keys are unchanged for everyone who doesn't opt in
- v2 only; v1 `/search` and SDKs untouched

## Tests

- Happy path: `safe: true` returns web results (gated to production runs)
- Failure path: non-boolean values (e.g. `"active"`) are rejected with `Invalid request body`


abimaelmartell (approved):
lg!

PR #4262, opened 2026-08-07
feat(api): negative DNS cache for scrape hostnames

rafaelsideguide:
## What

Adds a short-lived per-hostname circuit breaker for DNS resolution failures. When a scrape fails with a DNS error, the hostname gets a Redis marker (`DNS_NEGATIVE_CACHE_TTL_MS`, default 15 minutes, `0` disables). While the marker is fresh, repeat scrapes of that hostname return the same DNS error immediately — before any engine work — instead of re-running the whole engine waterfall against a hostname that cannot resolve.

## Design notes

- Markers live on the existing evict Redis connection and are never refreshed on read, so a dead hostname re-probes once per TTL window and domains that come back self-heal.
- Cache reads/writes fail open — Redis trouble never blocks a scrape.
- Zero-data-retention requests never write markers.
- The client-facing error contract is unchanged (same `SCRAPE_DNS_RESOLUTION_ERROR` code and message).

## Tests

- E2E: a scrape of a guaranteed-nonexistent `.invalid` hostname fails with the DNS error code and leaves a marker; a repeat scrape fails the same way.
- E2E: a manually planted marker on a hostname that resolves fine still returns the DNS error, proving the fail-fast path fires before any lookup.

---
<a href="https://app.blacksmith.sh/firecrawl/codesmith/firecrawl/pr/4262"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788722668&installation_model_id=11126&pr_number=4262&repository=firecrawl%2Ffirecrawl&return_to=https%3A%2F%2Fgithub.com%2Ffirecrawl%2Ffirecrawl%2Fpull%2F4262&signature=ac5c7a3714b8ed2bf9b643bce2d6b14475e3500110dd358ce8d3ee5c1a4de4e3"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Adds a short-lived negative DNS cache for scrape hostnames to fail fast on repeated DNS resolution errors. Uses a dedicated Redis client with no offline queue, no retries/resends, and 150ms bounds to avoid stalls or late writes.

- **New Features**
  - Fail fast only after two DNS failures within the TTL window; early-return `SCRAPE_DNS_RESOLUTION_ERROR` before any engine work.
  - Per-hostname Redis marker with TTL from `DNS_NEGATIVE_CACHE_TTL_MS` (default 15m, `0` disables); reads don’t refresh TTL so recovered domains self-heal.
  - Dedicated cache connection: skips ops when not ready, fails open, never enqueues while disconnected, and never retries or resends; cache calls are bounded to 150ms.
  - Only safe-integer counters are honored; TTL is floored to whole ms; cache-hit spans record failure attributes; zero-data-retention requests never write markers. Tests avoid production Redis and include an end-to-end fail-fast check.

<sup>Written for commit 34af6bc4125b28c2f1b6ccddd83f1bf1236005eb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/firecrawl/pull/4262?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/api/src/__tests__/snips/v2/scrape-dns-negative-cache.test.ts
+++ b/apps/api/src/__tests__/snips/v2/scrape-dns-negative-cache.test.ts
@@ -1,8 +1,13 @@
-import { describe, it, expect, beforeAll, afterAll } from "vitest";
+import { it, expect, beforeAll, afterAll } from "vitest";
 import crypto from "crypto";
+import { config } from "../../../config";
 import { describeIf, TEST_PRODUCTION } from "../lib";
 import { scrapeRaw, scrapeTimeout, idmux, Identity } from "./lib";
 import { redisEvictConnection } from "../../../services/redis";
+import {
+  cacheDnsFailure,
+  isDnsFailureCached,
+} from "../../../lib/dns-negative-cache";
 
 let identity: Identity;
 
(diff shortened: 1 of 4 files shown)


rafaelsideguide:
Addressed in 1924c0c8c:

1. **Span attributes on cache-hit returns** — valid; the early return now records `scrape.success`, `scrape.error`, `scrape.error_type`, and `scrape.duration_ms` alongside `scrape.dns_negative_cache_hit`.

2. **Population test gated to production runs** — the gate is required: self-hosted engine paths don't classify DNS failures into `SCRAPE_DNS_RESOLUTION_ERROR`, so the write path can't be exercised there. Documented with a comment; the read path (fail-fast) is tested on all configurations.

3. **Per-scrape Redis round-trip** — the single GET is negligible next to engine work and the scrape path's existing Redis usage, so no in-memory layer. The real hazard was that the shared evict Redis client queues commands while disconnected, which could stall scrapes instead of failing open — reads and writes are now bounded by a 150ms timeout.

4. **Transient failure blocks a hostname for the full TTL** — valid tradeoff; switched to a failure counter. A hostname fails fast only after two DNS failures within the TTL window, so one transient resolver blip never blocks it. Clear-on-success wasn't taken: while a hostname is blocked nothing probes it, so no success can occur to clear the marker, and an unconditional DEL on every successful scrape would put a write on the true hot path.

rafaelmmiller:
@cubic-dev-ai rerun

rafaelsideguide:
Addressed in 72cb973be:

1. **`parseInt` accepting numeric prefixes** — switched to `Number(raw)`, which rejects trailing garbage; malformed values now read as a cache miss. (Only `INCR` ever writes these keys, so this is defense in depth.)

2. **Timed-out writes lingering in the offline queue** — valid; the shared client can't be reconfigured without affecting its other users, so reads and writes are now skipped entirely while the client isn't `ready`. Nothing is ever queued for delayed delivery; the timeout still bounds slow-but-connected calls.

3. **Suite fails when the feature is disabled** — the whole suite is now gated on `DNS_NEGATIVE_CACHE_TTL_MS > 0`.

4. **Marker-planting tests touching shared Redis** — valid and the most important catch. Direct-write tests are now gated to non-production runs only, so they can only ever touch the local harness Redis. The production run asserts the DNS-error contract behaviorally without connecting to Redis at all. Write-path mechanics (two failures → blocked) are covered by a new module-level test against the harness Redis with a random hostname.

rafaelmmiller:
@cubic-dev-ai rerun

rafaelsideguide:
Addressed in 768b70e9b:

1. **In-flight commands surviving disconnects** — the cache now has its own Redis client (same URL, `enableOfflineQueue: false`, `maxRetriesPerRequest: 1`), mirroring the index cache's fail-fast client. Commands issued around a disconnect are dropped, never queued or retried into a later window. This ends the class of issue rather than patching around the shared client's defaults.

2. **`Number("Infinity")` blocking a hostname** — counters are now accepted only if `Number.isSafeInteger`; anything else reads as a miss.

3. **Fractional TTL leaving a counter without expiry (P1)** — valid: `PEXPIRE` rejects non-integers, which would have left an INCR'd key permanent. The TTL is floored to whole milliseconds at module load, so this can't occur regardless of the env value.

4. **Population test not reaching the fail-fast path** — valid: with the block threshold at 2, both scrapes were real DNS failures. A third scrape now runs, served by the fail-fast path, asserting it is indistinguishable from a real DNS failure. (The path-specific proof remains the planted-marker test, which forces a DNS error on a hostname that resolves.)

rafaelsideguide:
Addressed in the latest commit: valid — `maxRetriesPerRequest: 1` still allowed one post-reconnect retry, and `autoResendUnfulfilledCommands` defaults to true, so a command in flight at disconnect could be delivered into a later window. Both are now disabled (`maxRetriesPerRequest: 0`, `autoResendUnfulfilledCommands: false`). Callers abandon cache calls after 150ms, so retries and resends could never do anything except deliver stale writes; the client now drops them outright.

PR #4263, opened 2026-08-07
fix(api): register NuQ RabbitMQ consumer before setting listener

### LINKED ISSUE #4244: api: NuQ channel.consume() outside try/finally permanently breaks the RabbitMQ listener — all waitForJob calls hang after one consume() failure
## Summary

In `NuQ.startListener()` (`services/worker/nuq.ts`), `this.listener` is assigned **inside** the `try` block (line 146), but `await this.listener.channel.consume(...)` is called **after** the `try/finally` block ends (line 182). If `channel.consume()` throws, `this.listener` is already set and `this.listenerStarting` is already `false`. Every subsequent call to `startListener()` hits the early-exit guard `if (this.listener || ...) return` and returns immediately — the channel exists but **has no consumer**. Job completion messages pile up in the RabbitMQ queue unread, and every `waitForJob()` call hangs until its timeout.

## Affected code

**`apps/api/src/services/worker/nuq.ts`**

```typescript
private async startListener() {
  if (this.listener || this.shuttingDown || this.listenerStarting) return;  // ← guard

  if (config.NUQ_RABBITMQ_URL) {
    this.listenerStarting = true;

    try {
      const connection = await amqp.connect(...);
      const channel = await connection.createChannel();
      // ...
      this.listener = { type: "rabbitmq", connection, channel, queue: ... };
      //  ↑ listener is SET here, inside the try
    } finally {
      this.listenerStarting = false;
      // ↑ finally ends here — no consume() yet
    }

    // ... reconnect handler setup ...

    await this.listener.channel.consume(   // ← line 182: OUTSIDE the try/finally
      this.listener.queue,
      (msg) => { ... }
    );
  }
}
```

If `channel.consume()` throws:
- `this.listener` is truthy (set at line 146)
- `this.listenerStarting` is false (reset by finally)
- Every future `startListener()` call returns at the guard on line 124
- No consumer is ever registered on the channel

## Failure scenario

1. The broker is under load and rejects the `basic.consume` command with a channel-level AMQP exception.
2. `channel.consume()` throws at line 182.
3. `this.listener` is already set to a valid-but-consumer-less channel object.
4. The exception propagates to the caller; no retry path exists.
5. Every subsequent job completion pushes a message to the listen queue; no consumer reads it.
6. All `waitForJob()` callers block on `await done` until they hit their timeout, then throw `"Timed out waiting for job"`.
7. The only recovery is a process restart.

## Proposed fix

Move `channel.consume()` inside the `try` block, and clear `this.listener` in the catch path:

```typescript
try {
  const connection = await amqp.connect(...);
  const channel = await connection.createChannel();
  await channel.prefetch(5);
  const queue = await channel.assertQueue(...);

  await channel.consume(queue.queue, (msg) => { ... });  // ← moved inside try

  this.listener = { type: "rabbitmq", connection, channel, queue: queue.queue };
} catch (err) {
  this.listener = null;   // ensure guard doesn't lock out future attempts
  throw err;
} finally {
  this.listenerStarting = false;
}
```

edenbuilds:
## Summary
- `NuQ.startListener()` set `this.listener` inside `try/finally`, then called `channel.consume()` **outside** that block.
- If `consume()` threw, the early-return guard saw a truthy listener forever and never re-registered a consumer — every `waitForJob()` hung until timeout.
- Move `consume()` into the try, only publish `this.listener` after a successful consume, and clear it on failure.

Fixes #4244

## Test plan
- [ ] Simulate `channel.consume()` throwing after connect/assertQueue; confirm a later `startListener()` retries instead of no-op'ing.
- [ ] Confirm normal RabbitMQ listen path still delivers completion messages to `waitForJob()`.

Made with [Cursor](https://cursor.com)

---
<a href="https://app.blacksmith.sh/firecrawl/codesmith/firecrawl/pr/4263"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788722703&installation_model_id=11126&pr_number=4263&repository=firecrawl%2Ffirecrawl&return_to=https%3A%2F%2Fgithub.com%2Ffirecrawl%2Ffirecrawl%2Fpull%2F4263&signature=75978a7f3a2a88cde44e2b89a9700da2351ef74630e49263ea01b8d5d8825d5f"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Fixes a bug where `NuQ` could set a listener without a consumer, causing `waitForJob()` to hang. We now register the `RabbitMQ` consumer first, only set `this.listener` on success, and clean up with `closeAmqpResource` on failure; fixes #4244.

- **Bug Fixes**
  - Move `channel.consume()` into `startListener()` and set `this.listener` only after a successful consume; `noAck: false` unchanged.
  - On failure, clear the listener and close the channel/connection via `closeAmqpResource` to prevent leaks and allow retries; centralize `onClose` for fast auto-retry; update its expected-close log to "already closing" for restart safety.

<sup>Written for commit e2aef21298727b177f017c9255ae19cbd725d39f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/firecrawl/pull/4263?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/api/src/services/worker/nuq.ts
+++ b/apps/api/src/services/worker/nuq.ts
@@ -220,11 +220,40 @@ class NuQ<JobData = any, JobReturnValue = any> {
           // resources acquired before the failure so retries do not leak
           // connections/channels.
           this.listener = null;
+          // Prefer the shared close helper (expected AMQP close errors are soft).
+          // Keep the original consume failure as the thrown error so retries stay
+          // tied to the real cause; log unexpected close failures instead.
           if (channel) {
-            await channel.close().catch(() => {});
+            try {
+              await closeAmqpResource(
+                () => channel.close(),
+                "listener channel",
+              );
+            } catch (closeErr) {
+              logger.warn(
+                "NuQ failed to close channel after consume failure",
+                {
+                  module: "nuq/rabbitmq",
+                  error: closeErr,
+                },
+              );
+            }
           }
           if (connection) {
-            await connection.close().catch(() => {});
+            try {
+              await closeAmqpResource(
+                () => connection.close(),
+                "listener connection",
+              );
+            } catch (closeErr) {
+              logger.warn(
+                "NuQ failed to close connection after consume failure",
+                {
+                  module: "nuq/rabbitmq",
+                  error: closeErr,
+                },
+              );
+            }
           }
           throw err;
         }
--- a/apps/api/src/services/worker/nuq.ts
+++ b/apps/api/src/services/worker/nuq.ts
@@ -79,7 +79,7 @@ async function closeAmqpResource(
     await close();
   } catch (error) {
     if (isExpectedAmqpCloseError(error)) {
-      logger.info(`NuQ ${resource} already closing during shutdown`, {
+      logger.info(`NuQ ${resource} already closing`, {
         module: "nuq/rabbitmq",
       });
       return;
--- a/apps/api/src/services/worker/nuq.ts
+++ b/apps/api/src/services/worker/nuq.ts
@@ -143,72 +143,80 @@ class NuQ<JobData = any, JobReturnValue = any> {
           },
         );
 
+        // Register the consumer BEFORE publishing this.listener so a
+        // consume() failure cannot leave a consumer-less channel that
+        // permanently trips the early-return guard (#4244).
+        let reconnectTimeout: NodeJS.Timeout | null = null;
+
+        const onClose = function onClose() {
+          logger.info("NuQ listener channel closed", {
+            module: "nuq/rabbitmq",
+          });
+          this.listener = null;
+
+          if (reconnectTimeout) clearTimeout(reconnectTimeout);
+          reconnectTimeout = setTimeout(
+            (() => {
+              this.startListener().catch(err =>
+                logger.error("Error in NuQ listener reconnect", {
+                  err,
+                  module: "nuq/rabbitmq",
+                }),
+              );
+            }).bind(this),
+            250,
+          );
+          return;
+        }.bind(this);
+
+        await channel.consume(
+          queue.queue,
+          (msg => {
+            if (msg === null) {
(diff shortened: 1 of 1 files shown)


edenbuilds:
Addressed cubic P2: on `channel.consume()` failure we now close the half-open channel and connection (in addition to clearing `this.listener`) so retries do not leak broker resources.

edenbuilds:
Addressed cubic P3: consume-failure cleanup now uses `closeAmqpResource` (same helper as shutdown) instead of `.catch(() => {})`. Unexpected close errors are logged; the original consume failure is still what we rethrow so retries stay tied to the real cause.

edenbuilds:
Also took the follow-up cubic P3 on the log wording: `closeAmqpResource` now says \`already closing\` (not \`during shutdown\`), since the same helper is used for consume-failure cleanup as well as process shutdown.

PR #4282, opened 2026-08-09
fix(python-sdk): avoid paginating every batch status poll in wait loops

### LINKED ISSUE #4107: [Bug] Python SDK wait_for_batch_completion unnecessarily paginates and materializes all pages only to discard them
**Describe the Bug**
When using the Firecrawl SDK to perform batch scrapes using [`firecrawl.batch_scrape`](https://github.com/firecrawl/firecrawl/blob/main/apps/python-sdk/firecrawl/v2/methods/batch.py#L333) polling for batch completion unconditionally fetches entire batch set unconditionally, materializing every pending page's response body.

When requesting formats such as `rawHtml` on large HTML pages or using high batch sizes, the materialization performed for a simple poll is unnecessary work and hogs memory.

**To Reproduce**
Steps to reproduce the issue:
1. Set up the Firecrawl SDK
2. Run `firecrawl.batch_scrape` on a large batch. In the example pictures below I ran across 100 pages.
3. Watch the memory grow usage unnecessarily large.

**Expected Behavior**
We should not need to materialize every page in the batch for a batch scrape status check. We do not need to automatically paginate, but the caller of `get_batch_scrape_status` does not pass pagination options, which automatically defaults to `None`.

https://github.com/firecrawl/firecrawl/blob/8627119f39474f7394d7f74a36d7d146954b6fe7/apps/python-sdk/firecrawl/v2/methods/aio/batch.py#L71-L75

Caller does not pass options, meaning it becomes `None`

https://github.com/firecrawl/firecrawl/blob/8627119f39474f7394d7f74a36d7d146954b6fe7/apps/python-sdk/firecrawl/v2/client_async.py#L531-L547

Therefore, the pagination is set to `True`.

https://github.com/firecrawl/firecrawl/blob/8627119f39474f7394d7f74a36d7d146954b6fe7/apps/python-sdk/firecrawl/v2/methods/aio/batch.py#L98-L105

- OS: Linux
- Deployment Type: Cloud, firecrawl.dev
- Firecrawl Version: SDK version 4.32.1, also present on main branch

**Logs**
N/A

**Additional Context**
It is of course possible to "not crawl that large pages", but in our scenario we crawl customer webpages which we do not have control over.

It is also an option to reduce batch size, but that hurts concurrency and increases overall system latency.

Further optimizations are also possible by using a streaming JSON parser to not fully materialize each page body, but I figure that is a different problem.

Diwak4r:
## What

Fixes the Python SDK's batch-scrape wait loops paginating (and materializing) the **entire** batch result set on **every** status poll, only to discard it on the next poll.

Closes #4107

## Root cause

`get_batch_scrape_status()` defaults `auto_paginate=True` when no `PaginationConfig` is supplied:

```python
auto_paginate = pagination_config.auto_paginate if pagination_config else True
```

The wait loops were calling it with no config:

- sync: `wait_for_batch_completion()` in `firecrawl/v2/methods/batch.py`
- async: `wait_batch_scrape()` in `firecrawl/v2/client_async.py`

So a batch of N pages was fetching 1 + N requests every poll interval, for the full duration of the job. For large batches (thousands of URLs) that is a lot of wasted work, and the intermediate results are never even returned to the caller.

## Fix

Poll cheaply: pass `PaginationConfig(auto_paginate=False)` so status checks only fetch the first page. Once the job reaches `completed`, do a single final fetch with `auto_paginate=True` so callers still receive the full result set — the public behavior of `batch_scrape()` / `wait_batch_scrape()` is unchanged.

Both sync and async paths are fixed, with regression tests that assert:

- every status poll used `auto_paginate=False`
- the terminal completed fetch used `auto_paginate=True`
- callers still get the complete data

## Related

The async `wait_crawl()` loop has the same pattern and may deserve the same treatment; keeping this PR scoped to batch scrape as per the issue.

## Tests

```bash
python -m pytest firecrawl/__tests__/unit/v2/methods/test_pagination.py -k "batch or wait"
# 11 passed
```

---
## Summary by cubic
Make batch-scrape wait loops poll cheaply by not auto-paginating during status checks; only paginate once on completion. This reduces request load for large batches without changing SDK behavior.

- **Bug Fixes**
  - Use `PaginationConfig(auto_paginate=False)` for each poll in `wait_for_batch_completion()` and `AsyncFirecrawlClient.wait_batch_scrape()`.
  - After completion, perform one fetch with `auto_paginate=True` to return the full result set.

<sup>Written for commit 1c9637d419940e6f73af576f22819f3002202f8f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/firecrawl/pull/4282?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/python-sdk/firecrawl/__tests__/unit/v2/methods/test_pagination.py
+++ b/apps/python-sdk/firecrawl/__tests__/unit/v2/methods/test_pagination.py
@@ -4,6 +4,7 @@ Unit tests for Firecrawl v2 pagination functionality.
 
 import pytest
 import time
+import asyncio
 from unittest.mock import Mock, patch, AsyncMock
 from typing import Dict, Any, List
 
@@ -15,7 +16,12 @@ from firecrawl.v2.types import (
     DocumentMetadata
 )
 from firecrawl.v2.methods.crawl import get_crawl_status, get_crawl_status_page, _fetch_all_pages
-from firecrawl.v2.methods.batch import get_batch_scrape_status, get_batch_scrape_status_page, _fetch_all_batch_pages
+from firecrawl.v2.methods.batch import (
+    get_batch_scrape_status,
+    get_batch_scrape_status_page,
+    _fetch_all_batch_pages,
+    wait_for_batch_completion,
+)
 from firecrawl.v2.methods.aio.crawl import (
     get_crawl_status as get_crawl_status_async,
     get_crawl_status_page as get_crawl_status_page_async,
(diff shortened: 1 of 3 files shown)


Diwak4r:
Re-requesting review on this fix (open since Aug 9). The only failing check is the Vercel preview deployment, which is a non-blocking deploy gate and unrelated to the batch-wait polling change in this PR. Could a maintainer take another look? Happy to rebase or adjust if needed.

Diwak4r:
Closing this to clear out a backlog I let build up rather than leave it sitting in your queue. It was never picked up and at this age it isn't going to be. If it's still wanted, say so and I'll reopen it against current main.

Issue #451, opened 2024-07-24
Readiness and Liveness Probes [Feat]


JakobStadlhuber:
**Problem Description**
There are no Readiness and Liveness Probes for Health Monitoring or Self Healing processes

**Proposed Feature**
Readiness and Liveness HTTP Routes for every service

**Alternatives Considered**
none

**Implementation Suggestions**
see proposed

**Use Case**
Kubernetes Probes, Status/Health Pages

**Additional Context**
This can be extended with Redis checks or deep health checks

JakobStadlhuber:
If no one has time, I can also have a look into it.

JakobStadlhuber:
#457

nickscamara:
That's great @JakobStadlhuber! We will take a look at it soon!

nickscamara:
Got merged! Closing this :)

PR #4341, opened 2026-08-19
Handle a firebill refusal, and ramp by percentage instead of by allowlist

rakshith48:
## 1. Nothing acted on `false`

I audited every call site while checking the rollout gate: `trackedInRequest` is only ever read as a guard for compensation (`if (!result.success && trackedInRequest)`). **No branch anywhere responds to firebill answering `false`** — so an event firebill refused was dropped silently. Unbilled usage rather than a double charge, which is the less dangerous direction, but invisible either way.

`firebillTrack` now retries once and, if that fails too, counts it and logs at error.

Retrying is safe because the idempotency key is stable across attempts: if the first attempt *did* land (an ambiguous confirm timeout) the second is deduped by Autumn inside its 24h window rather than charged twice. That is not a guess — replaying a key against production Autumn left the balance untouched and produced `outcome="duplicate"`.

Deliberately **not** a throw, and deliberately **no fallback to Autumn**: firebill may already own the event, and the Autumn SDK sends no idempotency key, so the pair could not be deduped.

## 2. Nothing was observable

Three counters on the default prom-client registry, which the existing app and worker scrape configs already export:

| Metric | Use |
| --- | --- |
| `firecrawl_billing_route_total{route}` | firebill vs direct — **the ramp itself** |
| `firecrawl_firebill_track_total{path,outcome}` | `refused` is the one that means money |
| `firecrawl_firebill_retry_total{reason}` | blips that recovered |

The route ratio is what catches a broken gate in seconds. At 5% it should read ~0.05; 1.0 means everyone is routed and 0 means nobody is, and neither shows up in a customer balance for hours.

Note these increment mostly in the **worker**, since `scrape-worker` bills after the job — queries should aggregate across pods rather than pin to the api deployment.

## 3. `FIREBILL_ROLLOUT_PERCENT`

Sticky by org id, layered on top of the allowlist:

```ts
if (firebillOrgIds().has(orgId)) return true;              // always-on
return sampled(orgId, config.FIREBILL_ROLLOUT_PERCENT);    // sticky ramp
```

It reuses the FNV-1a bucketing `HIGHLIGHT_ROLLOUT_PERCENT` already uses, **extracted to `src/lib/rollout.ts`** so the two share one implementation — a second copy of a hash function is how two rollouts silently disagree. `highlights.ts` now imports it; its existing test still passes unchanged.

The property that matters, and that the tests pin: **a ramp only ever adds**. An org in the 5% bucket is still in it at 30%, so `5 → 10 → 30 → 60 → 80 → 100` never moves anyone back, and "is this org on firebill?" has one answer for the whole rollout. `0` is the kill switch and still routes the allowlist.

## Also: stale comments

Three comments still described the intent store — "records it durably in Postgres", "the intent row's primary key", "dedupe against the intent row". firebill has had no database since it moved to the queue; dedupe is Autumn's 24h window over the key firebill forwards. The conclusions were right, the mechanism named was gone.

## Verification

- **12 new tests**, all passing: sampling is off at 0 / on at 100, monotonic across the six ramp stages, deterministic per call, and lands within 5 points of the target share; allowlist beats 0 percent; nothing routes when firebill is unconfigured; retry-then-succeed, retry-then-give-up, retry-on-throw, and refunds hitting `/v1/refund` with the value made positive.
- **64 existing tests** in the three suites I touched (`autumn.service`, `highlights`, `credit_billing`) still pass.
- `tsc --noEmit` clean; prettier clean.

One environmental caveat: `pnpm install` fails here on the `foundationdb` native module, so I installed with `--ignore-scripts`. That leaves `@mendable/firecrawl-rs` unbuilt, and `tsc` reports it missing across ~15 unrelated files. Zero errors in anything I touched, and CI builds it properly.

## Ramp order once this and #451 are in

Alerts and dashboard first, then `FIREBILL_ROLLOUT_PERCENT: 5` and watch the route ratio, refused count, and parked depth before each step up.

---
## Summary by cubic
Routes billing through firebill by a sticky percentage and makes explicit refusals and transport failures visible. Previously a firebill false was ignored; now we retry once with a stable idempotency key (minted when absent), record outcomes, and never fall back to Autumn.

- Routing: `FIREBILL_ROLLOUT_PERCENT` (0–100) samples by org ID on top of the allowlist; 0 is the kill switch. Shared `sampled` lives in `src/lib/rollout.ts` and is used by highlights and firebill. `firecrawl_billing_route_total{route="firebill|direct"}` tracks the split.
- Failure handling: one retry with a short delay using the same idempotency key; we mint a key when callers do not provide one. On final failure, return false, count it, and log. Distinguish `refused` (explicit `success: false`) from `ambiguous` (non-OK or transport failures). Refunds go to `/v1/refund` with a positive value. No direct-Autumn fallback.
- Metrics: `firecrawl_firebill_track_total{operation="track|refund",outcome="accepted|refused|ambiguous"}` and `firecrawl_firebill_retry_total{reason}`; watch the firebill/direct ratio and refused+ambiguous during ramp. Aggregate across API and worker pods.
- Tests/build: adds rollout and firebill tests, including key-minting and retry paths; uses static imports to satisfy the build tsconfig. No other build/runtime changes.
- Required action: to ramp, set `FIREBILL_ROLLOUT_PERCENT` (e.g., 5) and increase gradually; keep 0 to route only the allowlist.

<sup>Written for commit 1efbb412b7d21dd58236a9031d73b553299898b2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/firecrawl/pull/4341?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/api/src/config.ts
+++ b/apps/api/src/config.ts
@@ -405,6 +405,12 @@ const configSchema = z.object({
   FIREBILL_URL: emptyStringAsUndefined(z.string().url()),
   FIREBILL_SECRET: emptyStringAsUndefined(z.string().trim().min(1)),
   FIREBILL_ORG_IDS: delimitedList(",").optional(),
+  // Sticky percentage ramp, on top of the allowlist above. The bucket is a
+  // hash of the org id, so an org that is in at 5 is still in at 30 — a ramp
+  // only ever adds, and never reshuffles who is on which path mid-rollout.
+  // 0 (the default) is also the kill switch: the allowlist still routes, and
+  // nothing else does.
+  FIREBILL_ROLLOUT_PERCENT: z.coerce.number().min(0).max(100).default(0),
 
   // Miscellaneous
   IDMUX_URL: z.string().optional(),
--- a/apps/api/src/lib/__tests__/rollout.test.ts
+++ b/apps/api/src/lib/__tests__/rollout.test.ts
@@ -1,0 +1,40 @@
+import { describe, expect, it } from "vitest";
+import { sampled } from "../rollout";
+
+const orgs = Array.from({ length: 2000 }, (_, i) => `org-${i}`);
+
+describe("sampled", () => {
+  it("is off at 0 and on at 100", () => {
+    expect(orgs.some(o => sampled(o, 0))).toBe(false);
+    expect(orgs.every(o => sampled(o, 100))).toBe(true);
+  });
+
+  it("only ever adds as the percentage rises", () => {
+    // The property the whole ramp rests on: nobody is moved back off firebill
+    // by a later step, so each stage is a superset of the one before it.
+    const stages = [5, 10, 30, 60, 80, 100];
+    for (const org of orgs) {
+      let wasIn = false;
+      for (const percent of stages) {
+        const isIn = sampled(org, percent);
+        expect(!wasIn || isIn).toBe(true);
+        wasIn = isIn;
+      }
+    }
+  });
+
+  it("gives the same answer every call", () => {
+    for (const org of orgs.slice(0, 50)) {
+      const first = sampled(org, 37);
+      for (let i = 0; i < 5; i++) expect(sampled(org, 37)).toBe(first);
+    }
+  });
+
+  it("lands near the requested share", () => {
+    for (const percent of [5, 30, 60]) {
+      const hits = orgs.filter(o => sampled(o, percent)).length;
+      const share = (hits / orgs.length) * 100;
+      expect(Math.abs(share - percent)).toBeLessThan(5);
+    }
+  });
+});
--- a/apps/api/src/lib/rollout.ts
+++ b/apps/api/src/lib/rollout.ts
@@ -1,0 +1,23 @@
+/**
+ * Deterministic percentage sampling for gradual rollouts.
+ *
+ * The bucket is a pure function of the cohort key, so a cohort that is in at
+ * 5% is still in at 30% — a ramp only ever adds, never reshuffles. That is the
+ * property `Math.random() < percent` does not have, and the reason to prefer
+ * this wherever a cohort should stay on one side of a rollout: billing routes,
+ * anything a support question could be asked about, anything you want to
+ * compare before and after.
+ *
+ * FNV-1a over the key, mapped onto [0, 100).
+ */
+export function sampled(cohortKey: string, percent: number): boolean {
+  if (percent <= 0) return false;
+  if (percent >= 100) return true;
+
+  let hash = 2166136261;
+  for (let i = 0; i < cohortKey.length; i++) {
+    hash ^= cohortKey.charCodeAt(i);
+    hash = Math.imul(hash, 16777619);
+  }
+  return ((hash >>> 0) / 0x1_0000_0000) * 100 < percent;
+}
--- a/apps/api/src/search/highlights.ts
+++ b/apps/api/src/search/highlights.ts
@@ -9,6 +9,7 @@ import {
 import type { HighlightFailureReason } from "./highlight-model";
 import { config } from "../config";
 import { logger as rootLogger } from "../lib/logger";
+import { sampled } from "../lib/rollout";
 
 // How far back into the index we're willing to reach for highlight source text.
 const HIGHLIGHTS_INDEX_MAX_AGE_MS = 30 * 24 * 60 * 60 * 1000; // 30 days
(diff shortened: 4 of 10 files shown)


rakshith48:
All four were valid, and the first was a bug this PR introduced. Fixed in e7d82e1.

## P1 — the retry could double-bill a keyless call

The sharpest of the four. My retry is only safe because both attempts are *the same event*, and I had that resting on the caller's idempotency key — but callers without a `chargeId` send none (fireclaw, fresh crawl ids), and firebill then mints one **per request**. So an accepted-but-lost first attempt would have been charged a second time by the retry. I introduced that risk in this PR; without the retry it didn't exist.

`firebillTrack` now mints a key when the caller has none and reuses it across attempts, while separate invocations still get their own. firebill would have minted an equivalent UUID anyway — doing it one level up is what makes it stable across the retry.

## P2 — a transport failure is not a refusal

Also right, and it matters for whoever answers the alert. Recording a timeout as `refused` and logging "it will not be billed" would send someone hunting for a charge that may well have landed.

- `refused` — explicit `success: false`. firebill did not take it; the usage is gone.
- `ambiguous` — timeout or non-OK. firebill may have accepted it and be delivering it right now.

Both still answer the caller `false`. I've updated [infra#451](https://github.com/firecrawl/firecrawl-infra/pull/451)'s dashboard panel to cover both rather than querying a label that would now miss half the cases.

## P2 — the `path` label

Documented as `track|refund`, emitted `/v1/track`. Renamed to `operation` and normalized to `track|refund`, which also matches the vocabulary firebill's own metrics use — so a query written against one works against the other.

## P3 — the unreachable `return false`

Correct. Removed by taking the final attempt out of the loop, which reads better than a fall-through that can't happen.

## Two things that fell out of fixing these

**Two existing tests pinned the old contract** ("omits the field otherwise"). Updated to the new one, with the reasoning in the test. Their sibling assertion — a caller-supplied key is forwarded verbatim — is unchanged.

**My own new tests had a flaw**, which is how I found the refused/ambiguous split wasn't being exercised: `mockResolvedValue(refused())` hands back *one* `Response`, and a body can only be read once, so the second attempt hit a stream error and looked like a transport failure. Each mock now builds a fresh Response.

**79 tests pass**, `tsc` clean (verified with `pnpm build`, the command CI runs, not just `--noEmit`), prettier clean.

rakshith48:
Checked this one rather than assuming, and the specific breakage doesn't apply — but chasing it found something that does.

## The label rename is safe

Every reference to these metrics in [infra#451](https://github.com/firecrawl/firecrawl-infra/pull/451) selects on `route`, `outcome` or `reason` — **none on the renamed label**:

```
firecrawl_billing_route_total{route="firebill"} / clamp_min(...)
sum by (route) (rate(firecrawl_billing_route_total[5m]))
sum(increase(firecrawl_firebill_track_total{outcome=~"refused|ambiguous"}[1h]))
sum by (reason) (rate(firecrawl_firebill_retry_total[5m]))
```

`grep` for a `path` selector across the VMRule and the dashboard returns nothing. And the metric has never been emitted in production — this PR is unmerged — so there is no existing series to orphan. The rename is entirely inside the pre-deploy window your comment identifies as the safe one.

## But you were right that the alert didn't exist

Your phrasing — *"the 'refused is proof usage is dropped' alert"* — described something I hadn't actually built. `refused` was on a **dashboard panel only**. Nothing paged on it, which for the one outcome that means definitively-lost usage is the wrong way round: a dashboard is something you look at, an alert is something that finds you.

Two rules added to #451, on apps/api's series rather than firebill's, since the caller is the only place that knows an event was refused:

| Alert | Severity | Fires on |
| --- | --- | --- |
| `FirebillEventsRefused` | **critical** | any `outcome="refused"` in 15m — usage that will not be billed |
| `FirebillEventsAmbiguous` | warning | >5 `outcome="ambiguous"` in 15m — the two can't talk, but not proof of loss, since the idempotency key means a retry is deduped |

That split only exists because of your previous round, so the two comments compound: separating refused from ambiguous is what made a critical alert on one of them safe to write without paging on every timeout.

#451 is green and mergeable with 8 alerts now. Nothing changed in this PR for this comment.

rakshith48:
Rebased onto main to clear the conflicts with #8fae7cadb (routing locks/finalizes through firebill) and #f69281b95.

Two conflicts, both mechanical:
- `autumn.service.ts` — import list; took the union.
- `firebill.ts` — main extracted `firebillUrl(path)` while this branch inlined the same concatenation into the new `firebillAttempt`. Kept main's helper, dropped the duplicate.

Also removed a `billingRouteTotal` import in `firebill.ts` that was never used (my own dead code from the first commit, not a rebase artifact), and corrected the `firebillTrack` docblock, which still said a `false` meant "not billed yet — firebill keeps retrying". Since this PR, `false` means firebill never took the event and nothing is retrying.

**Worth a reviewer's attention:** main now routes locks and finalizes through `shouldRouteToFirebill`, and this PR changes that function from allowlist-only to allowlist-or-percentage. So the percentage ramp will govern locks and finalizes too, not just track. `sampled()` is deterministic on `orgId`, so an org gets all three or none — a lock never lands in firebill while its track goes straight to Autumn.

Verified locally: `tsc --noEmit` clean (only the pre-existing `@mendable/firecrawl-rs` errors from the unbuilt native module), 76/76 tests pass across `rollout.test.ts`, `firebill.test.ts` and `autumn.service.test.ts` — including main's new lock/finalize routing tests, now running against the percentage gate.

micahstairs (approved):

PR #4350, opened 2026-08-19
feat(api): add rawBase64 scrape format for image URLs

rafaelsideguide:
## What

Scraping a direct image URL (an `image/*` content-type) previously failed with `SCRAPE_UNSUPPORTED_FILE_ERROR`, with no way to get the bytes back. This adds an opt-in **`rawBase64`** scrape format that returns the raw image bytes as a base64 **data URI**.

```jsonc
POST /v2/scrape
{ "url": "https://example.com/logo.png", "formats": ["rawBase64"] }

// -> data.rawBase64 = "data:image/png;base64,iVBORw0KGgo..."
//    data.metadata.contentType = "image/png"
```

## Design

- New `rawBase64` format in the v2 `formats` union + a `Document.rawBase64` field. Modeled as a **format** (like `audio`/`video`/`screenshot`) rather than a `parsers`-style toggle, since a raw image has no text/markdown representation to switch between.
- Routed through a dedicated `image` engine that mirrors the PDF "no parse" path: use the prefetched bytes when a scraping engine already fetched them, otherwise download directly. Download is capped at 50MB (PDF parity) so oversize images fail cleanly.
- Gated behind an `image` feature flag added only when `rawBase64` is requested **and** the resource is actually an image, so **default behavior is unchanged** — a bare scrape of an image still errors.
- `UnsupportedFileError` now appends a hint pointing at `rawBase64` for `image/*` content types.
- Output shape is a data URI so it's self-describing and directly usable; the real mime is also on `metadata.contentType`.

## Behavior notes

- **Non-image URL + `rawBase64`**: the field is simply omitted (no error); other formats work normally.
- **Crawl / batch**: works via the shared `scrapeOptions` — any scraped URL that is an image returns bytes. Crawl still does **not** follow image links; only a scraped image URL yields bytes.
- **Billing**: bills as a normal 1-credit scrape (these requests previously errored and billed nothing).

## SDKs

- TS: `rawBase64` added to the format union + `Document.rawBase64`.
- Python: `rawBase64`/`raw_base64` format literals, `Document.raw_base64`, and camel/snake normalization both directions.

## Tests

E2E snips (`scrape-raw-base64.test.ts`), gated to production engines:
- image URL + `rawBase64` -> data URI + `image/*` contentType
- image URL without the format -> error whose message hints at `rawBase64`
- HTML page + `["markdown", "rawBase64"]` -> markdown present, `rawBase64` absent, no error
- crawl with `rawBase64` in `scrapeOptions` -> base64 for the image

---
## Summary by cubic
Adds an opt-in rawBase64 scrape format that returns image bytes as a base64 data URI for direct image URLs. Previously these errored with SCRAPE_UNSUPPORTED_FILE_ERROR; default behavior is unchanged unless you request rawBase64.

- API: adds 'rawBase64' to the v2 formats union and Document.rawBase64 (data URI; actual mime in metadata.contentType).
- Engine/routing: new 'image' engine behind a feature flag, activated only when formats include 'rawBase64' and content-type is image/* (case-insensitive). Uses prefetch when available with a 50MB cap checked before reading and temp file cleanup; otherwise direct download with the same cap. Honors skipTlsVerification; proxyUsed from prefetch is preserved. No blanket index-cache bypass; image results are never indexed, and cached HTML pages continue to serve from cache with rawBase64 omitted.
- Behavior/billing: image URL without 'rawBase64' still errors (now hints at 'rawBase64'); non-image + 'rawBase64' omits Document.rawBase64 (no error); presence of rawBase64 counts as a waterfall success even without markdown; crawl works via scrapeOptions and still does not follow image links. Bills 1 credit per scrape.
- SDKs/tests: TypeScript and Python add the format and Document fields (Python also supports 'raw_base64'). E2E tests cover image success, image without format (hinted error), HTML + rawBase64 (omitted), and crawl with scrapeOptions.

Required action for clients who need image bytes: add 'rawBase64' to formats.

<sup>Written for commit 880721bcd2a7e25a4ddcdd2dbace523b003c4e86. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/firecrawl/pull/4350?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/api/src/__tests__/snips/v2/scrape-raw-base64.test.ts
+++ b/apps/api/src/__tests__/snips/v2/scrape-raw-base64.test.ts
@@ -1,0 +1,115 @@
+import { describeIf, TEST_PRODUCTION } from "../lib";
+import {
+  scrape,
+  scrapeRaw,
+  scrapeTimeout,
+  crawl,
+  idmux,
+  Identity,
+} from "./lib";
+
+let identity: Identity;
+
+beforeAll(async () => {
+  identity = await idmux({
+    name: "scrape-raw-base64",
+    concurrency: 100,
+    credits: 1000000,
+  });
+}, 10000 + scrapeTimeout);
+
+// A stable, Firecrawl-controlled image asset served with an image/* content-type.
+const IMAGE_URL = "https://firecrawl-test-site.vercel.app/declared-logo.png";
+// A regular HTML page on the same test site.
+const HTML_URL = "https://firecrawl-test-site.vercel.app/";
+
+// Fetching a real image URL depends on production scraping engines.
+describeIf(TEST_PRODUCTION)("Scrape rawBase64 format", () => {
+  it.concurrent(
+    "returns image bytes as a base64 data URI",
+    async () => {
+      const response = await scrape(
+        {
+          url: IMAGE_URL,
+          formats: ["rawBase64"],
+        },
+        identity,
+      );
+
+      expect(response.rawBase64).toBeDefined();
+      expect(typeof response.rawBase64).toBe("string");
+      expect(response.rawBase64!.startsWith("data:image/")).toBe(true);
+      expect(response.rawBase64).toContain(";base64,");
+
+      // The payload after the comma must decode to non-empty bytes.
+      const b64 = response.rawBase64!.split(",")[1];
+      expect(b64.length).toBeGreaterThan(0);
+      const buffer = Buffer.from(b64, "base64");
+      expect(buffer.length).toBeGreaterThan(0);
+
+      expect(response.metadata?.contentType?.startsWith("image/")).toBe(true);
+    },
+    scrapeTimeout,
+  );
+
+  it.concurrent(
+    "errors on an image URL without the format, hinting at rawBase64",
+    async () => {
+      const raw = await scrapeRaw(
+        {
+          url: IMAGE_URL,
+        },
+        identity,
+      );
+
+      expect(raw.statusCode).not.toBe(200);
+      expect(raw.body.success).toBe(false);
+      expect(raw.body.code).toBe("SCRAPE_UNSUPPORTED_FILE_ERROR");
+      expect(raw.body.error).toContain("rawBase64");
+    },
+    scrapeTimeout,
+  );
+
+  it.concurrent(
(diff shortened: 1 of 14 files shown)


rafaelsideguide:
Thanks — addressed all 7 findings in `afe013468`:

- **P1 (prefetch size cap):** the prefetched-bytes branch now enforces the same 50MB limit before encoding.
- **Waterfall success check:** `rawBase64` content is now a success signal, so a `formats: ["rawBase64"]` image no longer gets rejected for lacking markdown. (Good catch — this would have broken the happy path.)
- **Temp-file leak:** the prefetched image file is `unlink`ed in a `finally` after reading.
- **Case-insensitive content-type:** `image/` is matched case-insensitively (RFC 9110) in both the router and the image engine.
- **Index cache bypass:** `rawBase64` requests are excluded from index selection, so the bytes are always fetched fresh.
- **Feature-table mismatch (skipTls/stealth):** the image engine now advertises `skipTlsVerification: true, stealthProxy: false`, matching its actual basic-fetch download. When an upstream engine (incl. stealth) prefetches the image, the real `proxyUsed` still flows through the prefetch.

rafaelsideguide:
Round 2 addressed in `880721bcd`:

- **Prefetch size cap before read:** now `stat()`s the prefetched file and rejects oversize before `readFile()`, so an oversized prefetch never loads into worker memory.
- **Index exclusion narrowed (reverted the blanket rule):** you're right that excluding *all* `rawBase64` requests from the index over-corrected and regressed the common HTML-page + `['markdown','rawBase64']` and crawl (shared `scrapeOptions`) paths. Image-engine results are never indexed anyway — `shouldIndexForSearch` requires ≥200 chars of markdown, which an image (no markdown) never has — so an image URL won't have a stale cached doc lacking `rawBase64`. And if a URL *does* have a cached HTML doc, it's genuinely HTML, so serving that and omitting `rawBase64` is the correct graceful behavior. Reverted the exclusion.

PR #4363, opened 2026-08-20
fix(sdks): add spark-2 to agent model types

rakshith48:
Reported in Slack: the Python library has no option for `spark-2`.

[#4360](https://github.com/firecrawl/firecrawl/pull/4360) added `spark-2` to the `/v2/agent` schema and made it the **default**, but only touched `apps/api`. No SDK was updated.

## The part that wasn't reported

`GET /v2/agent/:id` always returns `model`, and it's now `"spark-2"` for any job that omitted one ([agent-status.ts:31-36,90](https://github.com/firecrawl/firecrawl/blob/main/apps/api/src/controllers/v2/agent-status.ts#L31)). Two SDKs validate that field **at runtime**:

- **Python** — `AgentResponse.model` was a pydantic `Literal`, so parsing raised `ValidationError: Input should be 'spark-1-pro' or 'spark-1-mini' [input_value='spark-2']`. It's parsed in `get_agent_status`, which `wait_agent` and the blocking `agent()` wrapper both call. The entire Python agent flow failed on the **first status poll** of any default-model job.
- **Rust** — same shape via a serde enum with no matching variant.

So this wasn't only a missing option; it was a hard break for the default path, as soon as the spark-2 default started serving.

## Changes

**Request side** — `model` gains `spark-2` in Python (`client`, `client_async`, sync + aio `methods/agent`), JS, Rust (`AgentModel::Spark2`), and the Elixir validator's `:in` list. Java and Go already took plain strings; only their docs were stale.

**Response side** — the server ships models without an SDK release, so an exact union here makes every future model a crash (Python) or a call-site type error (TS). Each SDK gets a catch-all appropriate to its type system:

- Python: `Optional[str]`
- TS: `... | "spark-2" | (string & {})` — keeps autocomplete, accepts unknown names
- Rust: `AgentModel::Unknown`, a `#[serde(other)]` variant. Purely additive — the enum stays `Copy`, every existing variant keeps its name and wire format, request construction is untouched. It serializes to `"unknown"` (which the API rejects), so it's documented as read-only.

## Verification

- Python: 399/399 unit tests pass. The two new response tests fail against the previous `Literal` and pass now — verified by reverting the field and re-running.
- Rust: `cargo build` clean; new `tests/agent_model.rs` covers round-tripping all three known models and degrading an unreleased name.
- JS: `tsc --noEmit` at the same 10-error pre-existing baseline before and after; jest identical (7 failed / 118 passed both ways, all pre-existing and unrelated).
- Elixir/Java not built locally (no `mix` on this machine; the Java change is a comment) — left to CI.

Versions bumped: python 4.37.1, js 4.34.1, rust 2.15.0 (+ CHANGELOG), elixir 1.9.2.

## CI failures on this PR are pre-existing

- **5× Self-hosted environment tests** — all the same `document-converter.test.ts` XLSX assertion. [#4362](https://github.com/firecrawl/firecrawl/pull/4362), which touches no SDK code, fails the identical 5 jobs.
- **Run tests** (JS SDK e2e) — `SdkError: Rate limit exceeded ... Remaining (req/min): 0` against the shared test account, across crawl/extract/scrape/batch. That suite contains no agent test at all, and TS types are erased at runtime, so a type-only widening cannot reach it.

## Follow-up, not fixed here

`AgentResponse.status` is a pydantic `Literal["processing","completed","failed"]`, but `wait_agent` checks for `"cancelled"`. That's dead today — `agent-status.ts` never emits it — but it's the same trap: if a cancelled status ever reaches the status payload, Python polling breaks identically.


rakshith48:
> P2: `AgentModel` is also deserialized in `AgentStatusResponse.model` ... Add a serde safety net on the enum (e.g. a `#[serde(other)]`/unknown catch-all variant ...)

Valid, and fixed in b6452ac.

My justification for leaving it was wrong in a way worth naming: I had believed `#[serde(other)]` was only permitted on internally or adjacently tagged enums, so I framed the choice as "exhaustive enum vs. breaking change to `String`" and picked the former. That premise is false for unit-only enums deserialized from a string, which is exactly this case — I asserted it from memory instead of testing it, and testing it took one `cargo test`.

So `AgentModel::Unknown` is purely additive: the enum stays `Copy`, every existing variant keeps its name and wire format, and request construction is untouched. No consumer breaks, and an unrecognized model no longer fails the whole `AgentStatusResponse` parse.

One correction to the review's reasoning, since it matters for the next case: "the response field is a value consumers only read, so widening/de-risking it would not break anyone" isn't quite right — a consumer doing `match status.model { Some(AgentModel::Spark1Pro) => ... }` would have broken at compile time under the `String` widening. The `#[serde(other)]` route is better precisely because it sidesteps that, not because the `String` route was free.

`Unknown` serializes to `"unknown"`, which the API rejects, so it's documented as read-only. `tests/agent_model.rs` covers round-tripping all three known models and degrading an unreleased name.

erikengervall (approved):

PR #4383, opened 2026-08-21
fix(api): report which field failed validation instead of "Bad Request"

alankritxghosh:
## What

Zod validation failures returned the bare string `"Bad Request"` as the `error`
field for every issue code except a top-level `unrecognized_keys` or a leading
`custom` refinement. The real issues only ever reached `details`, which most
clients never surface. This replaces that fallback with a message that names the
offending field path.

Refs #4054, where a create-monitor call failed with nothing but "Bad Request".

| case | before | after |
|---|---|---|
| monitor target uses `url` not `urls` | `Bad Request` | `targets.0.urls: expected array, received undefined` |
| `retentionDays: "30"` | `Bad Request` | `retentionDays: expected number, received string` |
| `parsers: [{type:"pdf", maxPages:10001}]` | `Bad Request` | `parsers.0.maxPages: Too big: expected number to be <=10000` |

`status`, `code`, and `details` are unchanged. Only the `error` string changes.

## Union branch selection

`targets` is a union of three strict objects, so any malformed target produces a
single top-level `invalid_union` issue with the per-branch failures nested in
`errors: ZodIssue[][]` — which is also why the existing `unrecognized_keys`
check never fires for a bad target.

Naive flattening reports every branch, so a caller who sent `type: "scrape"` is
told the API `expected "crawl"`. The rule used here: discard a branch that fails
on an `invalid_value` with **exactly one** permitted value at depth <= 1. One
permitted value means a literal discriminator (`{ type: "scrape" }`) or a
bare-literal branch (`z.literal("pdf")` beside an object form); an enum such as
`searchWindow` permits several and is an ordinary field the caller got wrong, so
its branch is kept. If every branch is disqualified, the first is reported.

## Contract change beyond v2 (please review this part specifically)

`index.ts:226` is the single app-level `ZodError` handler, and `v0Router`,
`v1Router`, `v2Router`, `labsRouter`, `exchangeRouter` and `adminRouter` all
mount on the same app. **This changes v0 and v1 error messages too, not just v2
monitors.** Controllers that catch `ZodError` locally (`v0/scrape.ts`,
`v0/crawl.ts`, `v1/search.ts`, `v2/search.ts`, ...) are unaffected; everything
that lets the error propagate is not.

The nine tests updated below all live in `snips/v2/`, and the only `v1`
assertion on this string is commented out (`snips/v1/crawl.test.ts:390`). So the
v1 message contract moves here with no test coverage either way. That is a
deliberate call, not an oversight — flagging it for an owner decision rather
than burying it in the diff.

## Tests updated

Nine assertions pinned the literal string `"Bad Request"` — eight in
`snips/v2/parsers.test.ts`, one in `snips/v2/system-prompt-rejection.test.ts`.
They pinned a placeholder rather than a contract, which is why this shipped:
they passed while clients got nothing actionable. Each now asserts the real
field path.

Added: 3 E2E cases in `snips/v2/monitor.test.ts` (wrong target shape, top-level
type mismatch, bad `searchWindow`) and 5 unit cases for the formatter.

## Relationship to #4382

#4382 adds the missing `integration` field to the monitor schema and is the
likely *functional* fix for #4054. This PR is the reason the next such failure
will be legible instead of arriving as a bare "Bad Request". They are
independent and can merge in either order.

---
## Summary by cubic
Returns field-specific Zod validation messages in the `error` string instead of the generic "Bad Request", so clients get actionable 400s. Previously most Zod failures returned `"Bad Request"`; now `error` includes the offending path (for example, `targets.0.urls: expected array, received undefined`) and prefers the deepest union branch to surface the real field.

- Add `formatZodIssues` to flatten issues, recurse into `invalid_union`, discard branches that only fail a literal discriminator at depth ≤ 1, then pick the branch that got furthest into the structure (max path depth).
- Cap output to 2 issues and 512 chars, appending `(and N more validation errors)` when truncated.
- Use it in the app-level `ZodError` handler; affects v0, v1, v2, `labs`, `exchange`, and `admin` routes that let the error propagate. Controllers that catch `ZodError` locally are unchanged.
- Contract: only `error` changes; `status`, `code`, and `details` are unchanged.
- Tests: update 9 assertions that pinned `"Bad Request"`; add 3 E2E monitor cases and 5 unit tests for the formatter.
- Rollout: no schema changes. If any client code relies on the literal `"Bad Request"`, update it. Related to #4054; independent of #4382.

<sup>Written for commit 3ccbb1ec3407463b1afd784d9971b4b093531913. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/firecrawl/pull/4383?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/api/src/__tests__/snips/v2/monitor.test.ts
+++ b/apps/api/src/__tests__/snips/v2/monitor.test.ts
@@ -114,6 +114,83 @@ describeIf(ALLOW_TEST_SUITE_WEBSITE && !TEST_SELF_HOST)("/v2/monitor", () => {
     expect(create.body.success).toBe(false);
   });
 
+  // Issue #4054: every zod code other than a top-level unrecognized key or a
+  // leading custom refinement used to collapse to the bare string "Bad Request",
+  // so a third-party client had nothing to act on. These pin the message, not
+  // just the status.
+  it("names the offending target field rather than returning a bare Bad Request", async () => {
+    const create = await monitorCreateRaw(
+      {
+        name: "wrong target shape",
+        schedule: { cron: "*/30 * * * *", timezone: "UTC" },
+        targets: [
+          {
+            type: "scrape",
+            // `url` where the scrape target wants `urls`
+            url: createTestIdUrl(),
+          },
+        ],
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      } as any,
+      identity,
+    );
+
+    expect(create.statusCode).toBe(400);
+    expect(create.body.success).toBe(false);
+    expect(create.body.error).not.toBe("Bad Request");
+    expect(create.body.error).toContain("targets.0.urls");
+    // The crawl branch only failed on its discriminator; surfacing it would send
+    // the caller after a field they never meant to set.
+    expect(create.body.error).not.toContain('expected "crawl"');
+  });
+
+  it("names the offending field for a top-level type mismatch", async () => {
+    const create = await monitorCreateRaw(
+      {
+        name: "wrong retention type",
+        schedule: { cron: "*/30 * * * *", timezone: "UTC" },
+        targets: [
+          {
+            type: "scrape",
+            urls: [createTestIdUrl()],
+          },
+        ],
+        retentionDays: "30",
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      } as any,
+      identity,
+    );
+
+    expect(create.statusCode).toBe(400);
+    expect(create.body.success).toBe(false);
+    expect(create.body.error).not.toBe("Bad Request");
+    expect(create.body.error).toContain("retentionDays");
+  });
+
+  it("names the offending field inside a search target", async () => {
+    const create = await monitorCreateRaw(
+      {
+        name: "bad search window",
+        schedule: { cron: "*/30 * * * *", timezone: "UTC" },
+        goal: "track launches",
+        targets: [
+          {
+            type: "search",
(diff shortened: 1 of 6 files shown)


alankritxghosh:
Confirmed and fixed — this was a real bug, thanks.

The discriminator rule only disqualifies branches that reject a *literal*, so a union without a discriminator left every branch eligible and `chosen[0]` won by position. Reproduced with `z.union([z.array(z.string()), z.strictObject({ mode: z.enum(["a","b"]), depth: z.number() })])`:

```
input:  { field: { mode: "zzz", depth: 1 } }
before: field: Invalid input: expected array, received object
after:  field.mode: Invalid option: expected one of "a"|"b"
```

Fix (3ccbb1ec): score each surviving branch by how far into the structure it got — the max path depth of its issues — and report the deepest, ties keeping the first. A branch that failed at its own root never matched the shape; one failing at `["mode"]` matched and is complaining about a real field.

Both rules earn their place rather than one replacing the other: for `{ type: "search", searchWindow: "3d" }` the scrape and search branches tie on depth, and only the discriminator rule disqualifies scrape. Added a unit case covering the no-discriminator union; the 16 previously-verified cases are unchanged.

PR #4406, opened 2026-08-25
feat(pdf): separate FirePDF async cohort for crawl/batch-originated scrapes

abimaelmartell:
## Why

A PDF scraped as a child of a crawl or batch scrape has no caller blocked on that specific document — the parent job is already asynchronous from the user's perspective. Treating those children like interactive single scrapes means bulk PDF work competes on the same latency-sensitive path as requests where someone is actually waiting.

## What

Adds a second, independently ramped cohort to the existing FirePDF async routing:

- `FIRE_PDF_ASYNC_BULK_ORIGIN_PERCENT` (default `0`, so this PR is traffic-neutral): percentage of crawl/batch-originated scrapes (any scrape carrying a `crawlId`, which covers both crawl children and batch-scrape children) routed to the FirePDF async jobs lane.
- The bulk cohort hashes a prefixed key (`bulk-origin:<scrapeId>`), so its percentage is independent of the general `FIRE_PDF_ASYNC_PERCENT` cohort.
- All hard exclusions keep precedence: ZDR, short remaining deadline, and `FIRE_PDF_ASYNC_DISABLE_TEAM_IDS` still keep a scrape sync regardless of origin.
- A bulk scrape outside its cohort falls through to the general percentage unchanged.
- The `fire_pdf_async_routed` log line now includes the bulk percentage and `crawl_id` for rollout observability.

## Tests

New routing tests cover: bulk cohort on/off, non-bulk scrapes unaffected, hard-exclusion precedence, cohort independence, and fall-through to the general percentage. Full pdf engine suite: 110 tests green; `tsc --noEmit` clean.

https://claude.ai/code/session_01GVt4uKwW6suAdfQikmYqaZ


abimaelmartell:
@cubic-dev-ai

PR #4410, opened 2026-08-25
feat(api): add PDF file size telemetry

abimaelmartell:
## Summary

Adds structured byte-size telemetry for completed PDF downloads and file-size limit errors. Declared sizes are recorded from Content-Length, while streaming overages record the bytes received when the configured limit is crossed.


abimaelmartell:
Closing in favor of instrumenting the size-limit rejection at the Fire Engine source.

PR #4156, opened 2026-07-28
fix(py-sdk): default get_batch_scrape_status to auto_paginate=False

rkfshakti:
Fixes #4107.

When `wait_for_batch_completion` polls for batch status, it calls `get_batch_scrape_status` without a `pagination_config`. The default for `auto_paginate` was `True`, so every poll cycle fetched and materialized every pending page's response body — even though only the `status` field is needed to decide whether the batch is done.

For large batches with `rawHtml` or high page counts, this caused unnecessary memory pressure during what should be a lightweight status check. The reporter measured significant memory growth polling a 100-page batch.

The fix changes the default from `True` to `False` in both the sync (`batch.py`) and async (`aio/batch.py`) `get_batch_scrape_status` functions. Callers who want auto-pagination can still opt in by passing `PaginationConfig(auto_paginate=True)`. `wait_for_batch_completion` already passes `PaginationConfig(auto_paginate=False)` explicitly, so its behavior is unchanged — only direct callers of `get_batch_scrape_status` without a config are affected, and they now get the sensible default of not materializing everything.

Added regression tests in `test_pagination.py` verifying that `get_batch_scrape_status` (sync and async) without a `pagination_config` makes only a single request and returns the `next` URL instead of following it.

---
[...]

PR #4424, opened 2026-08-27
fix(sdks): reliable batch scrape polling, pagination, and failure reporting (JS + Python)

KrisOei:
## Why

Large batch scrapes felt broken: every status poll downloaded result pages, truncated results looked complete, and failed jobs came back as empty successes (#4107, #4212).

## What

- Status calls only auto-paginate once the job is terminal, so a poll on a running job is a single request.
- Pagination raises on a failed page fetch and keeps the `next` cursor on truncation instead of returning silent partial data. It also stops on a self-referencing cursor instead of looping forever.
- `batchScrape` / `batch_scrape` waiters raise `JobFailedError` (job attached) on failed or cancelled jobs.

Same changes in both SDKs. Overlaps with #4377, #4320, #4156. Follow-up: the Python crawl path still has the old pagination behavior.

## Tests

Unit suites green in both SDKs (497 py, 19 js in touched files) plus the batch e2e suites against a locally running instance (5/5 both SDKs, cancel and pagination paths verified live).

---
## Summary by cubic
Makes large batch scrapes reliable in both SDKs: polling a running job is now a single request per poll, failed result pages raise instead of silently truncating, and waited-on jobs that fail, get cancelled, or fail during kickoff raise instead of coming back as empty successes.

**Breaking changes**

- `batchScrape` / `batch_scrape` waiters now throw a new `JobFailedError` (with the terminal job attached) for failed, cancelled, or kickoff-failed jobs instead of returning them.
- `PaginationConfig.auto_paginate` unset now means "SDK decides" instead of `true`; the SDK paginates batch and crawl status only after the job is terminal, while monitor keeps its old behavior.

**Pagination**

- A failed page fetch or a `success: false` page raises `PAGINATION_FETCH_FAILED` or the non-retryable `PAGINATION_RESPONSE_INVALID` instead of returning silent partial data.
- Truncation by `maxPages` / `maxResults` preserves the unconsumed `next` cursor, overshooting pages are skipped whole, and self-referencing `next` cursors stop pagination instead of looping.
- Versions bumped to `@mendable/firecrawl-js` 4.38.0 and `firecrawl` 4.41.0.

<sup>Written for commit 4556d456f35174bd02187bd4025b2dea1c150901. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/firecrawl/pull/4424?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/js-sdk/firecrawl/package.json
+++ b/apps/js-sdk/firecrawl/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@mendable/firecrawl-js",
-  "version": "4.37.0",
+  "version": "4.38.0",
   "description": "JavaScript SDK for the Firecrawl API: web scraping, crawling, web search, and scientific literature search over a research paper index of PubMed, bioRxiv, medRxiv and arXiv abstracts",
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
--- a/apps/js-sdk/firecrawl/src/__tests__/e2e/v2/batch.test.ts
+++ b/apps/js-sdk/firecrawl/src/__tests__/e2e/v2/batch.test.ts
@@ -2,6 +2,7 @@
  * E2E tests for v2 batch scrape (translated from Python tests)
  */
 import Firecrawl from "../../../index";
+import { JobFailedError } from "../../../v2/types";
 import { config } from "dotenv";
 import { getIdentity, getApiUrl } from "./utils/idmux";
 import { describe, test, expect, beforeAll } from "@jest/globals";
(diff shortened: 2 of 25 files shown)


KrisOei:
Pushed d3c66020e and 93aed3841 to address the review findings: pagination now defaults on only for completed jobs (a failed job raises before fetching pages), a 200 kickoff-failure response raises JobFailedError with the API's error message instead of a generic error, and the crawl path gets the same pagination fixes as batch in both SDKs. Plus the smaller test fixes (cancelled accepted in e2e handlers, no real sleep in the unit test).

PR #4438, opened 2026-08-28
fix(api): treat empty OPENAI_API_KEY/BASE_URL as unset (#4083)

### LINKED ISSUE #4083: Empty OPENAI_API_KEY / OPENAI_BASE_URL bypass zod handling: `json` scrape format fails silently while still consuming credits

SomSamantray:
## Summary

Self-hosted deployments that pass `OPENAI_API_KEY` / `OPENAI_BASE_URL` through docker-compose `${VAR}` passthrough receive empty strings when the variables are unset in `.env`. The config schema accepted `""`, so `createOpenAI` received a literal empty `apiKey`/`baseURL`, and `formats: ["json"]` requests failed with `Failed to parse URL from /responses` — returning `success: true` with no `json` field while still billing credits.

Fixes #4083.

## What changed

- `emptyStringAsUndefined` keeps its exact-`""` semantics. A new `emptyOrWhitespaceStringAsUndefined` helper also coerces whitespace-only strings, and is used only where a whitespace-only value is never meaningful: `OPENAI_API_KEY`, `OPENAI_BASE_URL`, and the `FIREBRAIN_TRACKS_*` pair (which already had inline trim behavior).
- Credential and URL fields that must fail closed at boot (`FIREBILL_URL`, `FIREBILL_SECRET`, `KEYLESS_CONVERSION_HMAC_SECRET`, `MCP_DELEGATED_CREDENTIAL_SECRET`, etc.) are untouched — a whitespace-only `FIREBILL_URL` still fails validation at startup instead of silently disabling billing routing.
- Config-layer coercion alone was not enough: `@ai-sdk/openai` re-reads `process.env` at request time via `loadApiKey` / `loadOptionalSetting` when the `createOpenAI` option is `undefined`. So `config.ts` now deletes `OPENAI_API_KEY` / `OPENAI_BASE_URL` from `process.env` after parsing when they were coerced to `undefined`, and warns at boot. The SDK then raises the intended missing-key error and falls back to the default base URL.
- The regression test moved into `src/__tests__/snips/v2/` so CI's `pnpm harness pnpm test:snips` lane executes it, and now also asserts `process.env` deletion, FIREBRAIN behavior, and whitespace-padded passthrough.

## Verification

- `vitest run src/__tests__/snips/v2/config.test.ts` — 8/8 pass (empty, whitespace-only, real, unset, `process.env` deletion, FIREBRAIN, padded passthrough).
- Empirically verified against the installed `@ai-sdk/openai` 3.0.71: before, env `""` produced request URL `"/responses"` (the `ERR_INVALID_URL` failure); after, the resolved URL uses the default `https://api.openai.com/v1` base.
- `vitest run src/search/highlights.test.ts src/lib/threat-protection/config.test.ts` — pass (45 tests total).
- `tsc --noEmit` — no errors in changed files (remaining repo-wide errors are pre-existing `@mendable/firecrawl-rs` native-module resolution, built by CI).

## Impact

- No migration, no deployment changes, no SDK/API surface changes.
- Operators with a set-but-empty OpenAI key now see a boot warning instead of silent per-request failures that still bill credits.
- Behavior for all other `emptyStringAsUndefined` users is unchanged (exact-`""` semantics preserved).

---
## Summary by cubic
Fixes self-hosted deployments where docker-compose `${VAR}` passthrough exposes empty `OPENAI_API_KEY` and `OPENAI_BASE_URL` strings. Empty and whitespace-only values are now treated as unset, so `@ai-sdk/openai` raises the intended missing-key error and falls back to the default base URL instead of failing per-request and still billing credits.

**Bug Fixes**

- Added `emptyOrWhitespaceStringAsUndefined` for `OPENAI_*` and `FIREBRAIN_TRACKS_*` vars; real values are trimmed so padded keys and URLs reach the SDK normalized.
- Deletes coerced-to-undefined OpenAI keys from `process.env` after config parse, since the SDK re-reads env at request time. Boot warnings differentiate: an empty key makes the OpenAI provider unavailable, while an empty base URL falls back to the default.
- Other fail-closed fields (`FIREBILL_*`, `KEYLESS_*`, `MCP_DELEGATED_*`) keep exact-`""` semantics, so whitespace-only values still fail validation at startup.
- Regression tests moved to `src/__tests__/snips/v2/` so CI's snips lane runs them; coverage includes empty, whitespace-only, real, unset, trimming, and `process.env` deletion cases.

<sup>Written for commit 0fdb5906bc1dd54dceb51b3cce19c71243d03fa0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/firecrawl/pull/4438?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/api/src/__tests__/config.test.ts
+++ b/apps/api/src/__tests__/config.test.ts
@@ -1,0 +1,60 @@
+import { afterAll, describe, expect, it, vi } from "vitest";
+
+// config.ts runs `validatedConfigSchema.parse(process.env)` at import time, so
+// each case sets the env then imports the module fresh via vi.resetModules().
+// Modeled on the env-reload pattern in snips/v2/audio-routing.test.ts.
+describe("config emptyStringAsUndefined (OPENAI_API_KEY / OPENAI_BASE_URL)", () => {
+  const originalOpenAiKey = process.env.OPENAI_API_KEY;
+  const originalOpenAiBaseUrl = process.env.OPENAI_BASE_URL;
+
+  async function loadConfig() {
+    vi.resetModules();
+    const mod = await import("../config.js");
+    return mod.config;
+  }
+
+  afterAll(() => {
+    if (originalOpenAiKey === undefined) {
+      delete process.env.OPENAI_API_KEY;
+    } else {
+      process.env.OPENAI_API_KEY = originalOpenAiKey;
+    }
+    if (originalOpenAiBaseUrl === undefined) {
+      delete process.env.OPENAI_BASE_URL;
+    } else {
+      process.env.OPENAI_BASE_URL = originalOpenAiBaseUrl;
+    }
+  });
+
+  it("treats an empty string as undefined (docker-compose ${VAR} passthrough)", async () => {
+    process.env.OPENAI_API_KEY = "";
+    process.env.OPENAI_BASE_URL = "";
+    const config = await loadConfig();
+    expect(config.OPENAI_API_KEY).toBeUndefined();
+    expect(config.OPENAI_BASE_URL).toBeUndefined();
+  });
+
+  it("treats whitespace-only strings as undefined", async () => {
+    process.env.OPENAI_API_KEY = "   ";
+    process.env.OPENAI_BASE_URL = "\t\n ";
+    const config = await loadConfig();
+    expect(config.OPENAI_API_KEY).toBeUndefined();
+    expect(config.OPENAI_BASE_URL).toBeUndefined();
+  });
+
+  it("passes real values through unchanged", async () => {
+    process.env.OPENAI_API_KEY = "sk-real-key";
+    process.env.OPENAI_BASE_URL = "https://api.openai.com/v1";
+    const config = await loadConfig();
+    expect(config.OPENAI_API_KEY).toBe("sk-real-key");
+    expect(config.OPENAI_BASE_URL).toBe("https://api.openai.com/v1");
+  });
+
+  it("leaves unset vars undefined", async () => {
+    delete process.env.OPENAI_API_KEY;
+    delete process.env.OPENAI_BASE_URL;
+    const config = await loadConfig();
+    expect(config.OPENAI_API_KEY).toBeUndefined();
+    expect(config.OPENAI_BASE_URL).toBeUndefined();
+  });
+});
--- a/apps/api/src/__tests__/snips/v2/config.test.ts
+++ b/apps/api/src/__tests__/snips/v2/config.test.ts
@@ -6,6 +6,8 @@ import { afterAll, describe, expect, it, vi } from "vitest";
 describe("config empty/whitespace OpenAI env vars (#4083)", () => {
   const originalOpenAiKey = process.env.OPENAI_API_KEY;
   const originalOpenAiBaseUrl = process.env.OPENAI_BASE_URL;
+  const originalFirebrainUrl = process.env.FIREBRAIN_TRACKS_URL;
+  const originalFirebrainApiKey = process.env.FIREBRAIN_TRACKS_API_KEY;
 
   async function loadConfig() {
     vi.resetModules();
@@ -23,6 +25,16 @@ describe("config empty/whitespace OpenAI env vars (#4083)", () => {
       delete process.env.OPENAI_BASE_URL;
     } else {
       process.env.OPENAI_BASE_URL = originalOpenAiBaseUrl;
+    }
+    if (originalFirebrainUrl === undefined) {
+      delete process.env.FIREBRAIN_TRACKS_URL;
+    } else {
+      process.env.FIREBRAIN_TRACKS_URL = originalFirebrainUrl;
+    }
+    if (originalFirebrainApiKey === undefined) {
+      delete process.env.FIREBRAIN_TRACKS_API_KEY;
+    } else {
+      process.env.FIREBRAIN_TRACKS_API_KEY = originalFirebrainApiKey;
     }
   });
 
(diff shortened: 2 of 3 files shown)


SomSamantray:
All 3 review findings addressed in 0fdb590:

1. **P2 (config.ts:600)** — Boot warnings now differentiate: an empty `OPENAI_API_KEY` makes the provider unavailable; an empty `OPENAI_BASE_URL` only falls back to the default `https://api.openai.com/v1` (provider still works when a key is present).
2. **P3 (config.test.ts)** — `FIREBRAIN_TRACKS_URL` / `FIREBRAIN_TRACKS_API_KEY` are now captured and restored in `afterAll`, matching the audio-routing pattern; no env leak.
3. **P3 (config.ts)** — `OPENAI_API_KEY` / `OPENAI_BASE_URL` are now `z.string().trim()`-normalized, so whitespace-padded values are trimmed before reaching `createOpenAI` (verified: generic-ai passes the trimmed config value; the SDK uses it directly).

Verified: 8/8 config tests pass (incl. trimmed-padded assertions), 45 tests total, typecheck clean for changed files.

SomSamantray:
## CI status

All code-side checks are green and all review feedback is resolved. The PR is blocked only on **maintainer-side approvals** that an external contributor cannot perform:

1. **Server Test Suite** (test-server.yml) and **Audit NPM Packages** — triggered but waiting for a maintainer to approve the workflow run for this fork PR (`action_required`). Approving requires admin rights (verified 403).
2. **Vercel deployment** — "Authorization required to deploy"; a maintainer must click *Approve and deploy* in Vercel for the fork PR.
3. **Human review** — no maintainer review yet.

Resolved this round:
- cubic AI review: **all 3 findings addressed** (differentiated boot warnings, FIREBRAIN env restore in tests, trim-normalized OpenAI values); cubic confirms "All reported issues were addressed".
- Contributor trust: ✅ · Superagent Security Scan: ✅ · Vercel Agent Review: neutral/skipped.

Once a maintainer approves the workflow runs, the real test suite will execute against `0fdb590`.

PR #4443, opened 2026-08-28
fix: use registrable domain for isSameDomain/isSameSubdomain (multi-part TLDs)

eeshsaxena:
## Problem

`isSameDomain` and `isSameSubdomain` derive the domain from the **last two hostname labels** (`hostname.split(".").slice(-2)`). That is wrong for multi-part public suffixes:

- `foo.co.uk` and `bar.co.uk` both reduce to `co.uk` → reported as the **same domain**
- `a.github.io` and `b.github.io` → same
- `x.vercel.app` / `y.vercel.app`, `*.com.au`, `*.pages.dev`, etc. → same

These two functions gate the `/map` endpoint's same-domain filter (`lib/map-utils.ts` and the v1 map controller filter `mapResults`/`links` with `isSameDomain`/`isSameSubdomain`). So a map of a site hosted on such a suffix (e.g. a `*.github.io` project page, a `.co.uk` site) can include URLs from **other people's sites** sharing the suffix. `isSameSubdomain` has the same defect in both its domain and subdomain computation (`sub.example.co.uk` split its domain as `co.uk` and its subdomain as `sub.example`).

## Fix

Use `parseHostname` (tldts, public-suffix aware) for the registrable domain and subdomain — the helper the rest of the codebase already uses for exactly this (see `lib/url-utils.ts`, which documents the `allowPrivateDomains` option so hosting suffixes like `vercel.app` are handled). A leading `www.` is still ignored in the subdomain comparison.

```ts
isSameDomain("https://foo.co.uk", "https://bar.co.uk")   // before: true   after: false
isSameDomain("https://a.github.io", "https://b.github.io")// before: true   after: false
isSameDomain("https://sub.example.com", "https://example.com") // true (unchanged)
```

## Testing

All existing `isSameDomain`/`isSameSubdomain` cases still pass, and I added tests for the multi-part-suffix cases. I verified the fixed logic against every existing test input plus the suffix cases by running it against `tldts` directly (the resolver the fix uses); the full API test suite runs in CI.

---
## Summary by cubic
Fixes `isSameDomain` and `isSameSubdomain` to use the public-suffix-aware registrable domain instead of the last two hostname labels, so the `/map` endpoint no longer treats `foo.co.uk` and `bar.co.uk` (or `a.github.io` and `b.github.io`) as the same site. Both functions now use `parseHostname` from `lib/url-utils`, and a leading `www.` is still ignored in subdomain comparisons. Adds tests for multi-part suffix cases.

<sup>Written for commit c973250d95e3d4b281fbb1d85e0919f971fd9a69. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/firecrawl/pull/4443?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/api/src/lib/validateUrl.test.ts
+++ b/apps/api/src/lib/validateUrl.test.ts
@@ -17,6 +17,14 @@ describe("isSameDomain", () => {
     expect(result).toBe(false);
   });
 
+  it("should treat different registrable domains under a multi-part suffix as different", () => {
+    // co.uk / github.io are public suffixes, so foo.co.uk and bar.co.uk are
+    // different sites. The last-two-labels split wrongly reported them equal.
+    expect(isSameDomain("https://foo.co.uk", "https://bar.co.uk")).toBe(false);
+    expect(isSameDomain("https://a.github.io", "https://b.github.io")).toBe(false);
+    expect(isSameDomain("https://x.example.co.uk", "https://y.example.co.uk")).toBe(true);
+  });
+
   it("should return true for a subdomain with different protocols", () => {
     const result = isSameDomain(
       "https://sub.example.com",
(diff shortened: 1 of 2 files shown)


eeshsaxena:
The user-facing impact is on /map: because these gate the same-domain filter, mapping a site on a shared suffix (a github.io project page, a co.uk or com.au site, a vercel.app deployment) can surface URLs from other sites under that suffix. The fix reuses parseHostname/tldts, which the codebase already relies on for registrable domains, so it stays consistent with url-utils.

PR #4450, opened 2026-08-29
fix(python-sdk): preserve base URL path prefix in _build_url

### LINKED ISSUE #4339: v2 client _build_url drops path prefix from api_url (urljoin with leading-slash endpoint)
**Title:** v2 client `_build_url` drops the path component of `api_url` (urljoin with leading-slash endpoint)

**Describe the bug**

A common self-hosted deployment exposes Firecrawl behind an **nginx reverse proxy** under a path prefix. For example, nginx routes `/firecrawl/` to the Firecrawl service, so the SDK is configured with:

```
api_url = "http://firecrawl.example.com/firecrawl"
```

When `api_url` includes such a path prefix, the v2 HTTP client builds request URLs that **drop the path prefix**, causing 404s from nginx (which only serves the `/firecrawl` path, and rightly refuses to add short paths like `/v2/*` just for one SDK).

**To Reproduce**

```python
from firecrawl import Firecrawl

client = Firecrawl(api_url="http://firecrawl.example.com/firecrawl")
# api_url has a /firecrawl path prefix (nginx reverse proxy)

client.search(query="test", limit=1)
```

Expected: `POST http://firecrawl.example.com/firecrawl/v2/search`
Actual: `POST http://firecrawl.example.com/v2/search` → 404 from nginx

**Root cause**

`firecrawl/v2/utils/http_client.py` `_build_url` (line 50):

```python
return urljoin(base_str, endpoint)
```

Every V2 method passes an absolute-path endpoint (e.g. `"/v2/search"` in `firecrawl/v2/methods/search.py` line 31). Per `urllib.parse.urljoin` semantics, a leading-slash relative reference replaces the ENTIRE base path, so `urljoin("http://host/firecrawl/", "/v2/search")` returns `http://host/v2/search` — the `/firecrawl` prefix is lost.

The legacy V1 client does this correctly by string concatenation (`f"{api_url}/v1/search"` in `firecrawl/v1/client.py`), which is why V1 works with prefixed base URLs and V2 does not.

**Expected behavior**

`_build_url` should preserve the base URL's path when joining relative endpoints, e.g.:

```python
def _build_url(self, endpoint: str) -> str:
    if endpoint.startswith("http"):
        return urljoin(self.api_url.rstrip("/") + "/", endpoint)
    return f"{self.api_url.rstrip('/')}/{endpoint.lstrip('/')}"
```

**Environment**
- firecrawl-py 4.17.0 (verified same code in 4.35.1, the latest at time of writing)
- Python 3.12, Windows
- Self-hosted Firecrawl behind an nginx reverse proxy with a path prefix

**Impact**
Any self-hosted deployment exposed through nginx (or any reverse proxy) under a non-root path (e.g. `/firecrawl/`) cannot use the V2 API through this SDK. The fix needs to be on the SDK side.

aniruddhaadak80:
## Why this change is needed

When `api_url` includes a path prefix (common in self-hosted deployments behind nginx or other reverse proxies), `_build_url` dropped the prefix because `urljoin` with a leading-slash relative reference replaces the entire base path.

Example: `urljoin("http://host/firecrawl/", "/v2/search")` → `http://host/v2/search` (prefix lost).

This breaks all self-hosted deployments exposed under a non-root path. Reported in #4339.

## What behavior changed

- **`apps/python-sdk/firecrawl/v2/utils/http_client.py`**: Replaced `urljoin(base_str, endpoint)` with `urlunparse`-based path combination for relative endpoints. Now parses `api_url`, appends the endpoint path to the base path via `f"{base_path}/{ep_path}"`, and reconstructs with `urlunparse` preserving `scheme`, `netloc`, `query`, and `fragment`. This fixes both the prefix loss and the raw-concatenation edge case where `api_url` contains query/fragment (cubic P2). Removed unused `urljoin` import.
- **`apps/python-sdk/firecrawl/__tests__/unit/v2/utils/test_http_client.py`**: New test file with 9 regression tests covering path-prefix preservation, trailing slashes, relative endpoints without leading slash, absolute URLs (same/different host), and protocol-relative URLs.

## Exact tests / checks ran

```
py -m pytest apps/python-sdk/firecrawl/__tests__/unit/v2/utils/test_http_client.py -v
# 9 passed

py -m pytest apps/python-sdk/firecrawl/__tests__/unit/v2/utils/ -v
# 111 passed in 1.21s (102 existing + 9 new)
```

No harness needed — SDK-only unit tests.

## Configuration, migration, security, or deployment impact

None. SDK-only change, no environment, migration, or deployment impact. No credentials in commits. Backwards-compatible — default `api_url` (`https://api.firecrawl.dev`) unchanged; only self-hosted prefix deployments gain correct behavior.

## Before / After

| `api_url` | `endpoint` | Before | After |
|---|---|---|---|
| `https://example.com/firecrawl` | `/v2/search` | `https://example.com/v2/search` | `https://example.com/firecrawl/v2/search` |
| `https://example.com/proxy/firecrawl` | `/v2/scrape` | `https://example.com/v2/scrape` | `https://example.com/proxy/firecrawl/v2/scrape` |
| `https://api.firecrawl.dev` | `/v2/scrape` | `https://api.firecrawl.dev/v2/scrape` | `https://api.firecrawl.dev/v2/scrape` (unchanged) |

Fixes #4339

### DIFF
--- a/apps/python-sdk/firecrawl/__tests__/unit/v2/utils/test_http_client.py
+++ b/apps/python-sdk/firecrawl/__tests__/unit/v2/utils/test_http_client.py
@@ -1,0 +1,107 @@
+"""Tests for HttpClient._build_url path-prefix preservation."""
+
+from firecrawl.v2.utils.http_client import HttpClient
+
+
+class TestBuildUrl:
+    """Verify _build_url preserves the base URL's path prefix."""
+
+    def _client(self, api_url: str) -> HttpClient:
+        return HttpClient(api_key=None, api_url=api_url)
+
+    def test_no_prefix_root_endpoint(self):
+        c = self._client("https://api.firecrawl.dev")
+        assert c._build_url("/v2/scrape") == "https://api.firecrawl.dev/v2/scrape"
+
+    def test_no_prefix_trailing_slash(self):
+        c = self._client("https://api.firecrawl.dev/")
+        assert c._build_url("/v2/scrape") == "https://api.firecrawl.dev/v2/scrape"
+
+    def test_preserves_path_prefix(self):
+        c = self._client("https://example.com/firecrawl")
+        assert c._build_url("/v2/scrape") == "https://example.com/firecrawl/v2/scrape"
+
+    def test_preserves_path_prefix_trailing_slash(self):
+        c = self._client("https://example.com/firecrawl/")
+        assert c._build_url("/v2/scrape") == "https://example.com/firecrawl/v2/scrape"
+
+    def test_preserves_deep_path_prefix(self):
+        c = self._client("https://example.com/proxy/firecrawl")
+        assert c._build_url("/v2/search") == "https://example.com/proxy/firecrawl/v2/search"
+
+    def test_relative_endpoint_no_slash(self):
+        c = self._client("https://example.com/firecrawl")
+        assert c._build_url("v2/scrape") == "https://example.com/firecrawl/v2/scrape"
+
+    def test_absolute_url_same_host(self):
+        c = self._client("https://example.com/firecrawl")
+        assert c._build_url("https://example.com/v2/scrape") == "https://example.com/v2/scrape"
+
+    def test_absolute_url_different_host_keeps_base_scheme(self):
+        c = self._client("https://example.com/firecrawl")
+        result = c._build_url("https://other.example.com/v2/scrape")
+        assert result == "https://example.com/v2/scrape"
+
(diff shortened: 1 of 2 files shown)


aniruddhaadak80:
Note for the maintainer: the only red check here is Vercel (fork-deploy authorization required) - all code checks are green on this head. Nothing to fix on my side; merging needs a maintainer-side deploy approval or Vercel exemption for fork PRs.

PR #4451, opened 2026-08-29
fix(python-sdk): skip pagination while polling batch scrape status

### LINKED ISSUE #4107: [Bug] Python SDK wait_for_batch_completion unnecessarily paginates and materializes all pages only to discard them

aniruddhaadak80:
## Why this change is needed

`wait_for_batch_completion` and the async `wait_batch_scrape` poll `get_batch_scrape_status` on every interval while waiting for a batch job to finish. By default that function auto-paginates through **all** result pages, materializing every scraped document in memory — even though the caller only needs the status fields (`status`, `completed`, `total`) to decide whether the job is done.

For large batches (100+ URLs), this turns each poll into an O(n) memory allocation that grows with every check, causing avoidable memory pressure and GC churn during what is otherwise a cheap status probe. Reported in #4107.

## What behavior changed

- **`apps/python-sdk/firecrawl/v2/methods/batch.py`**: `wait_for_batch_completion` now passes `PaginationConfig(auto_paginate=False)` on each intermediate poll, and re-fetches with default pagination (`auto_paginate=True`) once the job reaches a terminal state (`completed`/`failed`/`cancelled`) so callers receive the full document set. Updated docstring to document the optimization.
- **`apps/python-sdk/firecrawl/v2/client_async.py`**: Same fix for the async `wait_batch_scrape` loop (`async_batch.get_batch_scrape_status` with `poll_config`, then re-fetch without override on terminal).
- **`apps/python-sdk/firecrawl/__tests__/unit/v2/methods/test_batch_pagination_fix.py`**: 3 regression tests:
  - `test_poll_uses_pagination_config` — verifies correct endpoint and call count including terminal re-fetch
  - `test_auto_paginate_false_passed` — verifies intermediate polls carry `auto_paginate=False` and terminal re-fetch uses default pagination
  - `test_async_poll_uses_pagination_config` — same verification for the async wait loop

## Exact tests / checks ran

```
py -m pytest apps/python-sdk/firecrawl/__tests__/unit/v2/methods/test_batch_pagination_fix.py -v
# 3 passed in 2.22s

py -m pytest apps/python-sdk/firecrawl/__tests__/unit/v2/utils/ -v
# 111 passed (existing utils suite unaffected)
```

No harness needed — SDK-only unit tests.

## Configuration, migration, security, or deployment impact

None. SDK-only change, no environment, migration, or deployment impact. No credentials in commits. Backwards-compatible — `batch_scrape()` and `wait_for_batch_completion()` still return a fully-paginated `BatchScrapeJob`; intermediate polling is now cheaper but final result is identical (verified by tests).

## Before / After

| Batch size | Polls before completion | Memory per poll (before) | Memory per poll (after) |
|---|---|---|---|
| 1,000 URLs | ~20 | O(1000 docs) | O(10 docs) |
| 10,000 URLs | ~50 | O(10,000 docs) | O(10 docs) |

Terminal result: before and after both return full document set (now via explicit re-fetch; before via per-poll pagination).

Fixes #4107

### DIFF
--- a/apps/python-sdk/firecrawl/__tests__/unit/v2/methods/test_batch_pagination_fix.py
+++ b/apps/python-sdk/firecrawl/__tests__/unit/v2/methods/test_batch_pagination_fix.py
@@ -52,7 +52,7 @@ class TestWaitBatchCompletionPagination:
     """Verify the sync wait loop passes auto_paginate=False."""
 
     def test_poll_uses_pagination_config(self):
-        """Each poll call should include pagination_config with auto_paginate=False."""
+        """Each poll call hits the same endpoint; assertions cover request URL and call count."""
         from firecrawl.v2.methods.batch import wait_for_batch_completion
 
         client = MagicMock()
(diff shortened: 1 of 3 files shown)


aniruddhaadak80:
Note for the maintainer: the only red check here is Vercel (fork-deploy authorization required) - all code checks are green on this head. Nothing to fix on my side; merging needs a maintainer-side deploy approval or Vercel exemption for fork PRs.

PR #4489, opened 2026-09-01
feat(api): OCR raster images as one-page documents through FirePDF

abimaelmartell:
## Why

Scraping a raster image URL (`.jpg`, `.png`, `.tif`, …) fails today with `SCRAPE_UNSUPPORTED_FILE_ERROR`, even though the browser engine has already fetched the bytes and handed them to the API as a `file`. Scanned documents and other image-only content are a real workload, and today every one of them is a hard error that clients tend to retry.

## What

Treat an image handoff as a one-page scanned document and OCR it through FirePDF.

Rollout is per team: the new `imageOcr` team flag (with `FIRE_PDF_BASE_URL` configured) enables it. Single scrapes and parse uploads read the flag from the authenticated request; batch-scrape and crawl jobs carry no team flags, so for those it is resolved once per scrape from the cached team record. For every other team each entry point keeps today's behaviour exactly: image URLs fail as unsupported files and image uploads are rejected.

- **`lib/image-formats.ts`**: PNG/JPEG/TIFF/GIF/BMP content-type and extension maps plus magic-byte sniffing. These are the formats FirePDF opens as a one-page image document; WebP/SVG/AVIF stay unsupported and keep failing fast.
- **`specialtyScrapeCheck`**: when the header says `image/*` (or opaque `application/octet-stream`) and the bytes sniff as a supported format, hand off via `AddFeatureError(["image"], …, imagePrefetch)` instead of throwing. Header-less image payloads are sniffed too. Gated on the team flag; for other teams the old rejection is unchanged.
- **New `image` engine** (`engines/image/index.ts`), modelled on the document engine: reads the prefetch, sniffs the bytes, and posts them through the existing FirePDF client with `mode: "ocr"`. Reuses the result cache, the deadline contract, and the `pages` / `blocks` / `pageMarkers` parser options. `parsers: []` returns the raw base64 like the pdf engine. A FirePDF 400 maps to `UnsupportedFileError`. The inline size ceiling is shared with the PDF path (`FIRE_PDF_INLINE_HARD_MAX_FILE_SIZE`) and enforced here, since the upstream handoff cap is not guaranteed.
- **Routing**: `image` feature flag (priority 100, like pdf/document), set from the URL extension for flagged teams; the file-fetch route in `buildFallbackList` covers it; `index;documents` can serve cached results; the retry loop carries `imagePrefetch` like the other prefetches.
- **`/v2/parse`** uploads accept the same image types.
- Billing is the base scrape cost (one page; no `numPages` is reported, since an image is not a PDF).
- `UnsupportedFileError` message updated to describe what is and isn't supported.

## Verified

- Sample images run through FirePDF as one-page documents: typed and printed scans OCR accurately, tables included; logos return their text; icons and figure thumbnails return a caption; handwriting is poor, as expected. Large TIFF scans process fine.
- Snips (`pnpm harness pnpm test:snips -t "Image OCR"`) against fire-engine and FirePDF for a flagged team: PNG and JPEG URL OCR, raw-base64 passthrough, batch-scrape OCR, WebP still rejected, PNG and JPEG parse upload OCR. For an unflagged team, snips assert image URLs and uploads are still rejected as unsupported. Unit tests cover the format helpers.

## Follow-ups

- FirePDF side: explicit mime hint instead of content sniffing, WebP support, native-resolution rendering for image documents.
- Consider returning `SCRAPE_UNSUPPORTED_FILE_ERROR` as a 4xx so clients stop retrying.

### DIFF
--- a/apps/api/src/__tests__/snips/v2/parse.test.ts
+++ b/apps/api/src/__tests__/snips/v2/parse.test.ts
@@ -693,9 +693,9 @@ describe("/v2/parse", () => {
             formats: ["markdown"],
           },
           file: {
-            content: Buffer.from("not-an-image"),
-            filename: "upload.png",
-            contentType: "image/png",
+            content: Buffer.from("not-a-document"),
+            filename: "upload.exe",
+            contentType: "application/x-msdownload",
           },
         },
         identity,
--- a/apps/api/src/__tests__/snips/v2/scrape-image-ocr.test.ts
+++ b/apps/api/src/__tests__/snips/v2/scrape-image-ocr.test.ts
@@ -1,0 +1,674 @@
+import {
+  ALLOW_TEST_SUITE_WEBSITE,
+  describeIf,
+  TEST_SUITE_WEBSITE,
+} from "../lib";
+import { config } from "../../../config";
+import {
+  batchScrape,
+  idmux,
+  Identity,
+  parse,
+  parseWithFailure,
+  scrape,
+  scrapeTimeout,
+  scrapeWithFailure,
+} from "./lib";
+
+let identity: Identity;
+let ungatedIdentity: Identity;
+
+beforeAll(async () => {
+  // Image OCR is rolled out per team through the imageOcr team flag.
+  identity = await idmux({
+    name: "scrape-image-ocr",
+    concurrency: 100,
+    credits: 1000000,
+    flags: { imageOcr: true },
+  });
+  ungatedIdentity = await idmux({
+    name: "scrape-image-ocr-ungated",
+    concurrency: 100,
+    credits: 1000000,
+  });
+}, 10000 + scrapeTimeout);
+
+// Image OCR rides on fire-engine (the browser hands the image bytes to the
+// API) and FirePDF (which opens the bytes as a one-page scanned document).
+// The flagged identity only exists where idmux is reachable: the self-hosted
+// fallback identity carries no team flags.
+const IMAGE_OCR_AVAILABLE = !!config.FIRE_PDF_BASE_URL && !!config.IDMUX_URL;
+const SHOULD_RUN = !process.env.TEST_SUITE_SELF_HOSTED && IMAGE_OCR_AVAILABLE;
+
+// 760x220 four-colour PNG rendered from three lines of text:
+//   "Firecrawl OCR fixture"
+//   "Invoice number 4821 dated 2026-09-01"
+//   "The quick brown fox jumps over the lazy dog."
+const OCR_FIXTURE_PNG_BASE64 =
+  "iVBORw0KGgoAAAANSUhEUgAAAvgAAADcAgMAAAB47j15AAAADFBMVEUFBQX///9kZGSzs7PWDcSM" +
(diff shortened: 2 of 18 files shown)


abimaelmartell:
@cubic-dev-ai review comments addressed in c7db08ee9: full 8-byte PNG signature, octet-stream media-type normalization, image/x-tiff alias, alias coverage in the unit tests. The external image fixtures stay for now, consistent with the existing PDF snips; moving them to the test site needs a separate deploy. Please re-review.

abimaelmartell:
@cubic-dev-ai addressed in ecd481638: the document and pdf engines now decline when an image prefetch is in hand, and the parse supported-types message only lists image extensions when FirePDF is configured. The self-hosted rejection snip runs and passes in the self-hosted CI matrix, so I resolved that thread as a false positive. Please re-review.

abimaelmartell:
@cubic-dev-ai one more change in e0c8567f8: the whole feature is now behind an IMAGE_OCR_ENABLE config switch (default off, needs FIRE_PDF_BASE_URL) via a single helper, with the snips gated on the same helper. Please re-review.

abimaelmartell:
@cubic-dev-ai rollout changed once more in 35d05277c: instead of a global switch, image OCR is enabled per team via a new imageOcr team flag (with FirePDF configured). Team context flows from the scrape's team flags and from the authenticated request in the parse middlewares; unflagged teams keep the old rejection, asserted by snips with an unflagged identity. Please re-review.

abimaelmartell:
@cubic-dev-ai good catch on batch/crawl: fixed in 81da55d09. The decision is now resolved once per scrape in buildMetaObject, using the job's team flags when present and otherwise the cached team record, and stored on Meta for the handoff and the image engine. Covered by unit tests for the fallback and a batch-scrape snip for a flagged identity. Please re-review.

abimaelmartell:
@cubic-dev-ai addressed the remaining threads in 51eee9931: the raw parsers: [] image path uses the 50 MB inline cap like the pdf engine, and an upload reference whose kind is rejected at resolution now deletes the object and releases its reservation; the parse-upload OCR snip is gated on the flagged identity. The octet-stream sniffing thread I resolved without changes: upload kinds are detected by name or MIME for every type, and the upload-init endpoint has no bytes to sniff, so sniffing only at resolution would make the two disagree. Please re-review.

abimaelmartell:
@cubic-dev-ai both fixed in fd72ba7cc: the handoff file is stat'ed before it is read, and the raw-path cap is checked before the base64 string is built. Please re-review.

abimaelmartell:
@cubic-dev-ai f638d3823 swaps the external JPEG scrape test for an inline JPEG upload fixture and genericizes two unit-test paths; no production code changed. Please re-review.

abimaelmartell:
@cubic-dev-ai ac5261dec restores JPEG coverage on the scrape path with a long-lived public sample image and stable assertions (content type, status, non-empty markdown), so the image/jpeg handoff is exercised again without depending on OCR text. The other thread's first host was already removed in f638d3823; the remaining public sample host is the same accepted dependency pattern the PDF snips use. Please re-review.

PR #4523, opened 2026-09-03
fix(api): FirePDF async live submits fast-fail to the sync path on admission rejection

abimaelmartell:
## What

- New flag `FIRE_PDF_ASYNC_LIVE_FASTFAIL` (default **off**). When on, live (non-crawl) **inline** FirePDF async submits include `admission: "enforce"`, asking fire-pdf to reject up front if its lane queue cannot drain inside the caller's window.
- On `503 admission_rejected`, nothing was accepted server-side, so the document runs on the synchronous FirePDF path immediately (`scrapePDFWithFirePDF`) — the caller gets a result instead of a job that dies unstarted. Surfaced as a new `admission_rejected` fallback reason in the existing async-fallback metric.
- By-reference submits never opt in (they have no inline fallback); other 503s keep the existing `http_503` handling.

## Why

Interactive PDF scrapes have 30–60s windows. When an async lane queues behind a crawl dump or a surge, those jobs die unstarted while the sync path — which serves a 30-page document in ~6s — sits idle. This turns "died in queue" into "served synchronously". Companion to firecrawl/fire-pdf's request-level admission enforcement.

## Verification

- New tests: 503 `admission_rejected` → exactly one request, fallback called with the same base64, result returned; the `admission: "enforce"` field is sent for live inline submits when the flag is on and never for crawl-origin submits
- `vitest` firePDFAsync + routing suites green (47 tests), `tsc` clean, local cubic no issues

---
## Summary by cubic
Lets FirePDF async live submits fast-fail to the synchronous path on admission rejection, so interactive PDF scrapes get a result instead of dying unstarted in the queue.

When `FIRE_PDF_ASYNC_LIVE_FASTFAIL` is enabled (default off), live inline submits send `admission: "enforce"`; a 503 `admission_rejected` — meaning nothing was accepted server-side — runs the document synchronously via FirePDF immediately. Adds an `admission_rejected` reason to the async-fallback metric. By-reference submits never opt in, and other 503s keep the existing `http_503` handling.

<sup>Written for commit 408810452b44529c39f94375f101ff453aae28a3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/firecrawl/pull/4523?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/api/src/config.ts
+++ b/apps/api/src/config.ts
@@ -367,6 +367,13 @@ const configSchema = z.object({
   FIRE_PDF_ASYNC_FORCE_TEAM_IDS: z.string().optional(),
   FIRE_PDF_ASYNC_DISABLE_TEAM_IDS: z.string().optional(),
   FIRE_PDF_ASYNC_ALLOW_REQUEST_OVERRIDE: z.stringbool().default(false),
+  // Live (non-crawl) inline submits ask fire-pdf to ENFORCE lane admission:
+  // when the async queue cannot drain inside the caller's window, fire-pdf
+  // answers 503 admission_rejected and we run the document on the
+  // synchronous FirePDF path immediately instead of letting it die
+  // unstarted in the queue. Off by default; by-reference submits never
+  // opt in (they have no inline fallback).
+  FIRE_PDF_ASYNC_LIVE_FASTFAIL: z.stringbool().default(false),
   // Large-PDF by-reference submits (30-256MB files uploaded to GCS and
   // handed to fire-pdf via `input_gcs_uri`). This is an explicit on/off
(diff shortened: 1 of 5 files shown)


abimaelmartell:
On hold: the fire-pdf live/bulk origin split ships first; this stays a draft (flag off) until we see how the split behaves.

PR #4524, opened 2026-09-03
Nsc/new agent threads fix

nickscamara:
## Summary by cubic
Adds a skill export endpoint for finished agent runs, stops capping pinned exchange toolkits at five, and raises the agent prompt limit so runs can carry a recorded skill blueprint.

**New Features**
- New GET `/agent/:jobId/skill` route proxies the skill endpoint to extract-v3, so the dashboard can export finished runs as reusable skills.
- The response carries one workflow script per language, and the upstream status passes through so "no skill yet" renders as 404/409 instead of 500.
- Agent prompts may now be up to 50,000 characters; recorded skill blueprints do not fit in the old 10,000 limit.

**Bug Fixes**
- Pinned exchange toolkits are no longer capped at five; the list is a filter over the catalog, so longer lists were wrongly rejected.

<sup>Written for commit ddd5e6430d46120da022bb6a490c1692f12fafca. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/firecrawl/pull/4524?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


nickscamara:
@cubic review

PR #4529, opened 2026-09-04
fix(docker): reap zombie worker subprocesses in the api container

### LINKED ISSUE #4527: [Self-Host] Zombie process accumulation under sustained crawl/search load (node as PID 1, no `init` in official compose)
**Describe the Issue**

Self-hosting the official image without an init process: under sustained crawl/search load, the api container accumulates zombies parented to PID 1 (node). In production, ~6 hours of heavy load accumulated **467 zombies, all with `ppid=1`**; the main process then stayed alive but stopped serving HTTP and needed a restart. We can't prove the zombies caused the hang — the host was also under heavy load — but the PID-1 reaping gap is real. We've since redeployed with `--init`; we expect that to resolve the reaping but haven't re-run a sustained-load test.

**To Reproduce**

Not minimized — production observations:

1. Official self-host stack, stock settings, no `init: true` (ours: rootless podman 5.8.6).
2. ~20–25 concurrent jobs incl. `/v1/search`, ~5,800 dispatches/hour, ~6 hours.
3. Zombie count grows over hours:
   ```
   podman exec firecrawl-api sh -c 'ps -eo pid,ppid,state | awk "\$3==\"Z\"" | wc -l'
   ```
4. Under low load the count stays at 0 (~1h observed, still without init) — accumulation is load-dependent.

**Expected Behavior**

Worker subprocesses reaped as jobs complete; PID-1 orphan reaping handled by a proper init.

**Observed Behavior**

After ~6h, from the incident container:

```
$ podman exec firecrawl-api sh -c 'ps -eo pid,ppid,state | awk "\$3==\"Z\"" | wc -l'
467
```

All 467 zombies showed `ppid=1` (node main process). Same-period log, verbatim:

```
worker  2026-09-03 19:13:13 error [queue-worker:workerFun]: WORKER STALLED {"module":"queue-worker","method":"workerFun","cpuUsage":0.8064516129032258,"memoryUsage":0.5077951708594411}
```

The HTTP endpoint eventually stopped responding entirely while the process stayed up.

**Environment**

- OS: NixOS host
- Deployment Type: Self-hosted (rootless podman 5.8.6)
- Firecrawl Version: `ghcr.io/firecrawl/firecrawl:latest`, digest `sha256:bf26733900d199e1fec4c869ab936109bb865684e122bd0f27752f0df77e767b` (built ~3 months ago)
- Node.js Version: v22.22.3
- Database: PostgreSQL + Redis + RabbitMQ (official stack)

**Logs**

See "Observed Behavior".

**Configuration**

Stock official stack; no non-default process-management settings.

**Additional Context**

Verified at commit `1efdf695` (mechanism analysis is our speculation):

- `apps/api/Dockerfile`: `CMD ["node", "dist/src/harness.js", "--start-docker"]` — node directly as PID 1.
- Official `docker-compose.yaml`: no `init: true` on any service.
- `apps/api/src/harness.ts` (`execForward`): spawns workers with `detached: true`, kills via `kill(-pid)` group signals. Grandchildren outliving their parent end up under PID 1, which node doesn't reap. Exact fork site not pinpointed.

Suggested fix (standard for node-as-PID-1 images): add `init: true` to the official compose, or use tini/docker-init as ENTRYPOINT. Docker compose users have the identical PID-1 setup, so the compose fix covers most self-hosters. Happy to help test a PR.

(Searched open/closed issues first — didn't find an existing report of this.)

chelsealong:
## Why

Fixes #4527.

In the official `docker-compose.yaml`, the `api` container runs `node dist/src/harness.js --start-docker` directly as PID 1 (see `apps/api/Dockerfile` `CMD`). `harness.ts`'s `execForward` spawns the API server, worker, and extract-worker as detached child processes and signals them via process-group kills. Node running as PID 1 does not reap orphaned grandchildren the way a real init process does, so under sustained load (reported: ~20–25 concurrent jobs, ~5,800 dispatches/hour over 6 hours) zombie processes parented to PID 1 accumulate — 467 were observed in the reporter's incident, all `ppid=1`.

## What changed

Added `init: true` to the `api` service in `docker-compose.yaml`. This is Docker Compose's built-in flag to run `tini` as PID 1 inside the container, which correctly reaps zombie/orphaned processes. No Dockerfile or application code changes are needed since Compose injects `tini` at container-start time.

## Test plan

Added `apps/api/src/__tests__/docker-compose-init.test.ts`, which reads the repository's `docker-compose.yaml` and asserts the `api` service block sets `init: true`.

Verified the test fails without the fix and passes with it:

```
$ git stash push -- docker-compose.yaml   # temporarily remove the fix
$ pnpm exec vitest run src/__tests__/docker-compose-init.test.ts
 ❯ src/__tests__/docker-compose-init.test.ts:17:29
     expect(apiServiceBlock).toMatch(/^\s{4}init:\s*true\s*$/m);
 Test Files  1 failed (1)
      Tests  1 failed (1)

$ git stash pop   # restore the fix
$ pnpm exec vitest run src/__tests__/docker-compose-init.test.ts
 ✓ src/__tests__/docker-compose-init.test.ts (1 test) 3ms
 Test Files  1 passed (1)
      Tests  1 passed (1)
```

Also validated the compose file resolves correctly and carries the new field through to the final config:

```
$ docker compose -f docker-compose.yaml config | sed -n '/^  api:/,/^  redis:/p' | grep init
    init: true
```

Ran `pnpm run knip` in `apps/api` to confirm no new unused-export findings were introduced (only pre-existing configuration hints, unrelated to this change).

## AI assistance disclosure

This change was written with AI assistance (Claude Code).

---
## Summary by cubic
Prevents zombie worker subprocesses from accumulating in the API container by adding `init: true` to the `api` service in `docker-compose.yaml`. Node previously ran as PID 1 and didn't reap detached workers, causing zombies under load; now Docker Compose runs tini as PID 1 to handle reaping. Fixes #4527.

- Adds a regression test that parses `docker-compose.yaml` structurally to assert `init: true` is set.
- Adds `yaml` as a dev dependency in `apps/api` to support the test's YAML parsing.

<sup>Written for commit d343ce3f6e256cf2cbcc05f6b8fb692e306a84f9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/firecrawl/pull/4529?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -71,7 +71,8 @@
     "tsc-watch": "^7.1.1",
     "tsx": "^4.20.3",
     "undici-types": "^7.16.0",
-    "vitest": "^4.1.9"
+    "vitest": "^4.1.9",
+    "yaml": "^2.9.0"
   },
   "dependencies": {
     "@ai-sdk/anthropic": "^3.0.84",
--- a/apps/api/pnpm-lock.yaml
+++ b/apps/api/pnpm-lock.yaml
@@ -361,6 +361,9 @@ importers:
       vitest:
         specifier: ^4.1.9
         version: 4.1.9(@opentelemetry/api@1.9.1)(@types/node@22.19.1)(jsdom@29.1.1(@noble/hashes@1.8.0))(vite@8.0.16(@types/node@22.19.1)(esbuild@0.28.1)(jiti@2.6.1)(tsx@4.20.3)(yaml@2.9.0))
+      yaml:
+        specifier: ^2.9.0
+        version: 2.9.0
 
   native:
     devDependencies:
@@ -4804,11 +4807,6 @@ packages:
     resolution: {integrity: sha512-0pfFzegeDWJHJIAmTLRP2DwHjdF5s7jo9tuztdQxAhINCdvS+3nGINqPd00AphqJR/0LhANUS6/+7SCb98YOfA==}
     engines: {node: '>=10'}
 
-  yaml@2.8.3:
-    resolution: {integrity: sha512-AvbaCLOO2Otw/lW5bmh9d/WEdcDFdQp2Z2ZUH3pX9U2ihyUY0nvLv7J6TrWowklRGPYbB/IuIMfYgxaCPg5Bpg==}
-    engines: {node: '>= 14.6'}
-    hasBin: true
-
   yaml@2.9.0:
     resolution: {integrity: sha512-2AvhNX3mb8zd6Zy7INTtSpl1F15HW6Wnqj0srWlkKLcpYl/gMIMJiyuGq2KeI2YFxUPjdlB+3Lc10seMLtL4cA==}
     engines: {node: '>= 14.6'}
@@ -7762,7 +7760,7 @@ snapshots:
       nano-spawn: 1.0.2
       pidtree: 0.6.0
       string-argv: 0.3.2
-      yaml: 2.8.3
+      yaml: 2.9.0
     transitivePeerDependencies:
       - supports-color
 
@@ -8957,8 +8955,6 @@ snapshots:
   xtend@4.0.2: {}
 
   y18n@5.0.8: {}
-
-  yaml@2.8.3: {}
 
   yaml@2.9.0: {}
 
--- a/apps/api/src/__tests__/docker-compose-init.test.ts
+++ b/apps/api/src/__tests__/docker-compose-init.test.ts
@@ -1,0 +1,19 @@
+import { readFileSync } from "fs";
+import { join } from "path";
+
+// Regression test for zombie process accumulation under self-hosted Docker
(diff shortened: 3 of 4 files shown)


chelsealong:
Addressed: the test now parses `docker-compose.yaml` with the `yaml` package and asserts `services.api.init === true`, instead of regex-matching the raw file text. Verified this is more robust to reformatting (e.g. re-indenting the block breaks the old regex match but not the new structural parse), and the test still passes against the real compose file.

PR #4541, opened 2026-09-04
fix(api): require entity keys in schema analysis output

mogery:
Schema analysis declared `multiEntityKeys` as optional with a default. The AI SDK consequently omitted it from the JSON schema's `required` list, so strict structured-output providers rejected the request and analysis fell back to single-entity mode.

Require the array and explicitly prompt for `[]` for single-entity results. Keep the existing validation that multi-entity results contain at least one key.

Validation:
- Added four snips covering the actual completion helper and AI SDK serialization, valid single/multi-entity results, and invalid multi-entity output. Only the model boundary is mocked; these tests need no live AI credentials.
- The schema regression fails before the fix and all four tests pass afterward.
- API TypeScript build, Knip, Prettier, and `git diff --check` pass.
- `TEST_SUITE_SELF_HOSTED=true pnpm harness pnpm exec vitest run src/__tests__/snips/v1/extract-schema-analysis.test.ts` built successfully, then stopped during container setup because the local Docker daemon was unavailable. The normal Server Test Suite will run these snips in CI.

---
## Summary by cubic
Makes `multiEntityKeys` required in schema analysis output so strict structured-output providers no longer reject requests and fall back to single-entity mode.

- Prompts now explicitly ask the model to return `[]` for single-entity results.
- Existing validation that multi-entity results contain at least one key is preserved.

<sup>Written for commit d1a1f99e1734a7502bfc899418fd7af895861109. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/firecrawl/pull/4541?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


mogery:
@cubic-dev-ai re-review pls

PR #4542, opened 2026-09-04
fix(api): prevent duplicate monitor finalization

mogery:
Two reconcilers can retain the same running check in separate batches and each complete it. A later worker can confirm an already-consumed credit hold and overwrite confirmed billing with a failed outcome.

Refresh the check from the primary after acquiring its Redis lease, and atomically claim the running-to-terminal transition before confirming or releasing credits. Apply the same claim to stale-check cleanup in the scheduler, prevent delayed reservation/target writes from changing terminal checks, stop scheduling when another finalizer wins the claim, and release Redis leases only when the token still belongs to the caller. Settle the hold returned by the terminal database claim, and release a late, unpersisted reservation only after a primary read excludes ownership by another worker. Record refused or failed credit releases as failed settlement.

Validation: 101 tests pass across the runner, scheduler, and store suites; TypeScript and Knip pass. The new regressions cover stale snapshots, overlapping finalizers after lease expiry, replacement lease ownership, losing confirm/release claims, delayed job delivery, late reservations, reservation persistence failures, and refused or throwing releases. The duplicate-confirmation tests fail against the previous implementation. The server workflow explicitly runs the new regression suites on one matrix entry. Local service-backed tests were blocked by the unavailable Docker daemon; CI covers those configurations.

This change prevents duplicate completion processing. Recovery from a crash between the terminal transition and external credit settlement remains unchanged.

CI: the monitor regression suites, three server configurations without a proxy, FoundationDB core tests, dependency audit, and security scan pass. Cubic reports zero issues on the current commit. The two proxy configurations fail in the unrelated scrape-robots suite introduced on main by #4543; that PR also has both proxy checks failing.

### DIFF
--- a/.github/workflows/test-server.yml
+++ b/.github/workflows/test-server.yml
@@ -218,6 +218,16 @@ jobs:
         env:
           npm_config_ignore_scripts: 'true'
 
+      - name: Run monitor finalization regression tests
+        if: matrix.engine == 'fetch' && matrix.proxy == 'no-proxy' && matrix.nuq == 'postgres'
+        run: >-
+          pnpm exec vitest run
+          src/services/monitoring/runner.finalization.test.ts
+          src/services/monitoring/store.finalization.test.ts
+          src/services/monitoring/scheduler.test.ts
+          --reporter=default
+        working-directory: apps/api
+
       - name: Install test site dependencies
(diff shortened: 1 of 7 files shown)


mogery:
@cubic-dev-ai re-review

PR #4545, opened 2026-09-04
feat(api): isolate Spur cache with a dedicated Redis client

mogery:
When `SPUR_REDIS_URL` is configured, route Spur lookup results, failure markers, and single-flight locks exclusively to a dedicated Redis client. Rate limiting continues using its existing client. With no new env set, existing installations retain their current behavior.

The dedicated client bounds command waits and disables offline queuing. Failures follow Spur's existing fail-open behavior without falling back to stale shared data.

Existing keyless snips seed and clean up through the selected Spur client, and the suspicious VPN fixture marks the tunnel anonymous. Cleanup removes only the two fixture keys.

Validation: all 13 existing Spur tests passed; `git diff --check` and the pre-commit checks passed. The API harness was blocked before test execution by an existing native-module build error: `@mendable/firecrawl-rs` has no exported member `validateRegexes` in `src/lib/crawl-regex.ts`.


markkuehn08-ops:
@copilot

markkuehn08-ops (commented):
Leave playwright out please

PR #4547, opened 2026-09-05
docs(contributing): note pnpm --ignore-scripts for foundationdb build failure

### LINKED ISSUE #4060: Fresh pnpm install fails: foundationdb native build needs fdb_c.h that no setup doc mentions
## Description

Following CONTRIBUTING.md on a fresh macOS machine (node 22 plus pnpm 11.4.0, the pinned `packageManager`), `pnpm install` in `apps/api` fails while building the `foundationdb` dependency:

```
../src/utils.h:13:10: fatal error: 'foundationdb/fdb_c.h' file not found
make: *** [Release/obj.target/fdblib/src/module.o] Error 1
 ELIFECYCLE  Command failed with exit code 1.
```

The install aborts entirely, so `node_modules/.bin` ends up empty and nothing works after that (no tsx, no vitest, the harness cannot start). Root cause: `foundationdb` is a hard dependency and its install script compiles native bindings against the FoundationDB C client library, which a typical dev machine does not have and which is not listed as a prerequisite anywhere. CONTRIBUTING.md lists node, rust, pnpm, redis and postgres only.

CI already works around this in two different ways: `test-server.yml` installs the `foundationdb-clients` deb before `pnpm install`, and the `nuq-fdb-core` job sets `npm_config_ignore_scripts: 'true'` instead. Local contributors get neither hint.

Skipping the build is safe for the default setup: `NUQ_BACKEND` defaults to postgres, and the only value import of the module is lazy (`apps/api/src/services/worker/nuq-fdb/client.ts`, a `require` inside the function; everything else is type-only). I verified `pnpm install --ignore-scripts` on macOS arm64 and the API, workers and the self-hosted snips all run fine on the postgres backend.

## Suggested fix

A few options, in rough order of effort:
1. document the FDB client prereq (or the `--ignore-scripts` escape hatch) in CONTRIBUTING.md
2. move `foundationdb` to `optionalDependencies` so its failed build does not kill the whole install
3. use pnpm's `onlyBuiltDependencies` allowlist so the foundationdb build script does not run unless opted in

Happy to hear which direction you would prefer.

small note from a college freshman trying to contribute for the greater good :) I hit this while setting the repo up locally and worked through the root cause with Claude Code's help. apologies if I misread anything, and I would genuinely love to learn if so.

gaurav-init:
## Summary

`pnpm install` in `apps/api` on a fresh machine aborts while building the `foundationdb` native binding with `fatal error: 'foundationdb/fdb_c.h' file not found`. The FoundationDB C client is not listed as a prerequisite and most dev machines do not have it, so new contributors following `CONTRIBUTING.md` end up with an empty `node_modules/.bin` and no path forward.

The queue defaults to PostgreSQL and the `foundationdb` module is required lazily only when `NUQ_BACKEND=fdb` (`apps/api/src/services/worker/nuq-fdb/client.ts`), so skipping the native build is safe for the default setup. CI already takes the same shape on rows that do not exercise the FDB backend (`.github/workflows/test-server.yml` uses `pnpm install --frozen-lockfile --ignore-scripts` + `npm_config_ignore_scripts: 'true'`).

This adds one short paragraph in the *Set up API development* section pointing new contributors at the same escape hatch when they hit the error. Docs-only change — no code, no scripts, no dependency changes.

## Related

Fixes #4060

## Test plan

- [x] Rendered `CONTRIBUTING.md` — the new paragraph lands right after the existing `pnpm install` / `pnpm start` block; no other content moves.
- [x] `pnpm install --ignore-scripts` on macOS arm64 succeeds when a fresh install without the flag fails on the same box.
- [x] Ran the API on the default PostgreSQL queue backend after the flagged install — nothing lazy-requires `foundationdb`.
- [x] Diff is 6 additions, 0 deletions in a single file.

---
## Summary by cubic
Adds a note to `CONTRIBUTING.md` for new contributors whose `pnpm install` fails building the `foundationdb` native binding because the FoundationDB C client header isn't installed.

- The fix is `pnpm install --ignore-scripts`, which is safe because the queue defaults to PostgreSQL and the `foundationdb` module is only loaded when `NUQ_BACKEND=fdb`.
- CI already skips scripts on non-FDB matrix rows, so this matches existing behavior.

<sup>Written for commit 61df0723873183a59af355ce799cda14b11f3afc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/firecrawl/pull/4547?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -26,6 +26,12 @@ pnpm start
 
 `pnpm start` builds Firecrawl and launches the API, workers, and local dependency containers. Keep Redis running separately as described in the public guide.
 
+If `pnpm install` fails building the `foundationdb` native binding (`fatal error: 'foundationdb/fdb_c.h' file not found`), pass `--ignore-scripts`. The queue defaults to PostgreSQL and the module is required lazily only when `NUQ_BACKEND=fdb`, so skipping the build is safe for the default setup — CI does the same on the non-FDB matrix rows.
+
+```bash
+pnpm install --ignore-scripts
+```
+
 ## Make a focused change
 
 1. Fork the repository and create a branch whose name describes the change.


gaurav-init:
@mogery this docs-only PR fixes a fresh-clone blocker for new contributors — adds one paragraph to CONTRIBUTING.md pointing to `pnpm install --ignore-scripts` when the foundationdb native build fails (same workaround CI uses on non-FDB matrix rows). Zero code changes. Happy to hear if you'd rather handle this differently.

PR #4564, opened 2026-09-07
fix(api): do not interpolate BULL_AUTH_KEY into Express paths

### LINKED ISSUE #1791: [Self-Host] Bull Auth Key Includes Regex Characters
**Describe the Issue**
If the `BULL_AUTH_KEY` env var includes characters that need to be escaped in a regex the firecrawl api process will crash on boot. This is because the `BULL_AUTH_KEY` is being used in the adminRouter's expressjs paths, which ultimately are compiled down to a regex.

**To Reproduce**
Steps to reproduce the issue:
1. Configure the env var `BULL_AUTH_KEY` with a regex character. I hit this with the character `)`. 
2. Boot the firecrawl api
3. Observe the SyntaxError: Invalid regular expression: /^\/admin\/<your-bull-auth-key-here>\/redis-health\/

**Expected Behavior**
Honestly, I'm not sure. I think the code itself might be fine but more documentation around this env var would help. The naming implies it's a secret, so I generated a random password, which happened to include an invalid value. Then the output from booting was very obscure, with the only hint seeing it was the express module and the URL in the stack.

**Screenshots**

- OS: Linux
- Firecrawl Version: v1.14.0
- Node.js Version: 22
- Docker Version (if applicable): n/a
- Database Type and Version: n/a

**Logs**
If applicable, include detailed logs to help understand the self-hosting problem.

**Configuration**
Provide relevant parts of your configuration files (with sensitive information redacted).

**Additional Context**
Add any other context about the self-hosting issue here, such as specific infrastructure details, network setup, or any modifications made to the original Firecrawl setup.

claxman:
Fixes https://github.com/firecrawl/firecrawl/issues/1791.

Express 5.2.1 compiles `/admin/${BULL_AUTH_KEY}/...` as a route pattern. A key containing `)` throws PathError at boot and the process never listens. Same for `(`, `+`, `?`, `{`, `}`, `[`, `]`. `apps/api/.env.example` line 68 sets `BULL_AUTH_KEY=@`, which registers. Both admin.ts and the Bull Board `app.use` in index.ts interpolate the key.

The secret is no longer in the pattern. `bullAuthRoute` emits one `:bullAuthN` per non-empty `key.split("/")` segment, with a literal empty segment for `//` or a trailing `/` (path-to-regexp `:name` needs 1+ chars). Middleware joins the params and `timingSafeEqual`s them against `BULL_AUTH_KEY`. `setBasePath` uses `bullAuthPublicPath`, which `encodeURIComponent`s each segment so `?` is not a query delimiter and a literal `%2F` survives Express decoding. `*bullAuthKey` was rejected because it is greedy in path-to-regexp v8, so `/admin/<key>/queues/api/queues` captures `<key>/queues/api`.

Wrong key with the same segment count returns JSON 404 `{ error: "Not found" }`. No env change. Keys that already worked as a literal path, including `/`, `//`, and trailing `/`, keep the same URLs. S2S bearer compare uses `secretsMatch` from bull-auth.ts. MCP ingest keeps its own `timingSafeSecretEqual`. prettier also rewrapped a few existing long lines in admin.ts.

`cd apps/api && ./node_modules/.bin/vitest run src/lib/__tests__/bull-auth.test.ts` on 03f71585a: 11 passed. Covers GET `/admin/<encoded-key>/queues/api/queues` for `secret)oops`, `abc/def`, `a//b`, `trail/`, `foo?bar`. Encoded GET for `a%2Fb` is 200, raw `/admin/a%2Fb/...` is 404. Same-segment-count misses `x//y` and `wrong/` return JSON 404. With `encodeURIComponent` removed from `bullAuthPublicPath`, 3 of 11 fail (the public-path asserts plus GET 200 for `foo?bar` and `a%2Fb`). `./node_modules/.bin/knip --cache` exit 0.

### DIFF
--- a/apps/api/src/index.ts
+++ b/apps/api/src/index.ts
@@ -122,7 +122,7 @@ const { addQueue, removeQueue, setQueues, replaceQueues } = createBullBoard({
 if (config.BULL_AUTH_KEY) {
   serverAdapter.setBasePath(`/admin/${config.BULL_AUTH_KEY}/queues`);
   app.use(
-    `/admin/:bullAuthKey/queues`,
+    `/admin/*bullAuthKey/queues`,
     createRequireBullAuth(config.BULL_AUTH_KEY),
     serverAdapter.getRouter(),
   );
--- a/apps/api/src/index.ts
+++ b/apps/api/src/index.ts
@@ -16,7 +16,7 @@ import { v0Router } from "./routes/v0";
 import os from "os";
 import { logger } from "./lib/logger";
 import { adminRouter } from "./routes/admin";
-import { createRequireBullAuth } from "./lib/bull-auth";
+import { bullAuthRoute, createRequireBullAuth } from "./lib/bull-auth";
 import http from "node:http";
 import https from "node:https";
 import { v1Router } from "./routes/v1";
@@ -122,7 +122,7 @@ const { addQueue, removeQueue, setQueues, replaceQueues } = createBullBoard({
 if (config.BULL_AUTH_KEY) {
   serverAdapter.setBasePath(`/admin/${config.BULL_AUTH_KEY}/queues`);
   app.use(
-    `/admin/*bullAuthKey/queues`,
+    bullAuthRoute(config.BULL_AUTH_KEY, "/queues"),
     createRequireBullAuth(config.BULL_AUTH_KEY),
     serverAdapter.getRouter(),
   );
--- a/apps/api/src/index.ts
+++ b/apps/api/src/index.ts
@@ -16,6 +16,7 @@ import { v0Router } from "./routes/v0";
 import os from "os";
 import { logger } from "./lib/logger";
 import { adminRouter } from "./routes/admin";
+import { createRequireBullAuth } from "./lib/bull-auth";
 import http from "node:http";
 import https from "node:https";
 import { v1Router } from "./routes/v1";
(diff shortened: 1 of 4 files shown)


claxman:
Addressed the encode and empty-segment miss comments.

`setBasePath` now uses `bullAuthPublicPath`, which percent-encodes each non-empty key segment. GET `/admin/foo%3Fbar/...` and `/admin/a%252Fb/...` return 200. Unencoded `?` is treated as a query string (404). Raw `%2F` is decoded to `/` and 404s.

Empty-segment tests now also GET a same-shape wrong key (`x//y`, `wrong/`) and assert JSON 404.

Left `secretsMatch` in bull-auth.ts. MCP ingest has its own `timingSafeSecretEqual`; moving both into a shared helper would pull MCP routes into this PR.

Left JSON vs HTML 404 for the wrong segment count. Matching-shape misses already return JSON `{ error: "Not found" }`. A catch-all `/admin` handler would also rewrite unrelated unmatched admin paths.

Verified on 03f71585a: `cd apps/api && ./node_modules/.bin/vitest run src/lib/__tests__/bull-auth.test.ts` 11 passed. Stripping `encodeURIComponent` from `bullAuthPublicPath` fails 3 of 11 (public-path asserts, `foo?bar`, `a%2Fb`).

PR #4590, opened 2026-09-10
fix(sdks): forward recordSession in browser-create (JS + Python)

### LINKED ISSUE #4589: JS + Python SDKs cannot disable browser session recording (recordSession missing from browser-create)
The `/v2/browser` API accepts `recordSession` (defaults to `true`, passed through as `record` to the browser service in `apps/api/src/controllers/v2/browser.ts`), so session recording can be switched off per session.

But neither the JS nor the Python SDK exposes it — `browser()` has `ttl`, `activityTtl`, `streamWebView`, `profile` and nothing else, so there is no way to create an unrecorded session from either SDK. I checked for an existing report or PR covering this and found none.

Happy to extend this to the other SDKs as a follow-up if this looks good.

soroush5:
I noticed the API supports `recordSession` on `POST /v2/browser` (wired through to the browser service as `record`), but the JS and Python SDKs had no way to send it — sessions created from either SDK are always recorded.

What I changed:
- JS: added `recordSession` to `browser()` args and the request payload in `src/v2/methods/browser.ts` (left unset by default so the server default still applies).
- Python: added `record_session` to `browser()` in the sync + async method modules and both client wrappers, sent as `recordSession`.

Tests I ran:
- New unit tests on both sides asserting `recordSession: false` is forwarded and that the key is omitted when unset: `src/__tests__/unit/v2/browser-record-session.unit.test.ts` (jest, 2 passed) and `apps/python-sdk/tests/test_browser_record_session.py` (pytest, 3 passed).
- Proved they fail without the fix by stashing the source changes and re-running (both suites go red, then green again after popping).
- Neighboring suites: JS unit dir is 161 passed with 2 pre-existing failures (`threat-protection`, `scrape` — both fail identically on the clean tree); Python `tests/` shows the same 13 pre-existing failures before and after my change, with my 3 new tests passing. `tsc --noEmit` reports only pre-existing errors in files I didn't touch.

No config, migration, or deployment impact — purely additive optional params.

Fixes #4589

---
## Summary by cubic
Fixes `#4589` — adds `recordSession` to browser-create in the JS and Python SDKs. Sessions created from either SDK were always recorded because the API's `recordSession` option was never sent; now the SDKs pass it through, and callers who omit it still get the server's default.

- JS: optional `recordSession` arg on `browser()` is forwarded to the request body.
- Python: optional `record_session` arg added to sync and async `browser()` methods, sent as `recordSession`.
- New unit tests on JS, sync Python, and async Python verify `recordSession: false` is forwarded and the key is omitted when unset.
- Purely additive — no config, migration, or deployment impact.

<sup>Written for commit 3f93acc686e0c2a1786282d16706f5a38a38c7c5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/firecrawl/pull/4590?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/js-sdk/firecrawl/src/__tests__/unit/v2/browser-record-session.unit.test.ts
+++ b/apps/js-sdk/firecrawl/src/__tests__/unit/v2/browser-record-session.unit.test.ts
@@ -1,0 +1,33 @@
+import { describe, expect, jest, test } from "@jest/globals";
+import { browser } from "../../../v2/methods/browser";
+
+function httpMock() {
+  return {
+    post: jest.fn(
+      async () => ({ status: 200, data: { success: true, id: "sess_123" } }),
+    ),
+  } as any;
+}
+
+describe("v2 browser recordSession", () => {
+  test("forwards recordSession: false so session recording can be disabled", async () => {
+    const http = httpMock();
+
+    await browser(http, { recordSession: false });
+
+    expect(http.post).toHaveBeenCalledWith(
+      "/v2/browser",
+      { recordSession: false },
+    );
+  });
+
+  test("omits recordSession when the caller does not set it (server default applies)", async () => {
+    const http = httpMock();
+
+    await browser(http, { ttl: 600 });
+
+    expect(http.post).toHaveBeenCalledWith("/v2/browser", { ttl: 600 });
+    const payload = http.post.mock.calls[0][1] as Record<string, unknown>;
+    expect(payload).not.toHaveProperty("recordSession");
+  });
+});
--- a/apps/js-sdk/firecrawl/src/v2/client.ts
+++ b/apps/js-sdk/firecrawl/src/v2/client.ts
@@ -582,7 +582,7 @@ export class FirecrawlClient {
   // Browser
   /**
    * Create a new browser session.
-   * @param args Session options (ttl, activityTtl, streamWebView, profile).
+   * @param args Session options (ttl, activityTtl, streamWebView, recordSession, profile).
    * @returns Session id, CDP URL, live view URL, and expiration time.
    */
   async browser(
--- a/apps/js-sdk/firecrawl/src/v2/methods/browser.ts
+++ b/apps/js-sdk/firecrawl/src/v2/methods/browser.ts
@@ -16,6 +16,7 @@ export async function browser(
     ttl?: number;
     activityTtl?: number;
     streamWebView?: boolean;
+    recordSession?: boolean;
     profile?: {
       name: string;
       saveChanges?: boolean;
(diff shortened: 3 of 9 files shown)


erekola:
I checked the Python async path at `54e1af3` with `httpx.MockTransport`, starting from `AsyncFirecrawlClient.browser()` and inspecting the serialized POST body. All four cases passed:

| Argument | Request body |
|---|---|
| `record_session=False` | `"recordSession": false` |
| `record_session=True` | `"recordSession": true` |
| `record_session=None` | Key omitted |
| Argument omitted | Key omitted |

The same test also checked that `stream_web_view=False` and `profile.save_changes=False` survived serialization with both TTL values. On main at `bbadeb8`, the three explicit-argument cases failed with `TypeError`. Omitting the argument passed.

The added Python tests currently cover the sync path. Adding this async matrix would cover the changed async facade and helper through the HTTP serialization step. The check used Python 3.13.15 and httpx 0.28.1 with synthetic responses. It did not contact the API or verify recording behavior on the server. I have not run the JS tests or the full SDK suite. AI assisted this check.

<details>
<summary>Async regression test (pytest)</summary>

```python
"""Public async facade -> real HTTP serialization -> in-memory MockTransport."""
import asyncio
import json

import httpx
import pytest
from firecrawl.v2.client_async import AsyncFirecrawlClient


@pytest.mark.parametrize("case", ["false", "true", "none", "omitted"])
def test_async_browser_recording_payload(case):
    async def run():
        requests = []

        def handler(request):
            requests.append(request)
            return httpx.Response(200, json={"success": True, "id": "test-session"})

        client = AsyncFirecrawlClient(api_key="dummy-test-key", max_retries=1)
        await client.async_http_client.close()
        client.async_http_client._client = httpx.AsyncClient(
            base_url="https://example.test", transport=httpx.MockTransport(handler)
        )
        args = dict(ttl=600, activity_ttl=120, stream_web_view=False,
                    profile={"name": "test-profile", "save_changes": False})
        if case != "omitted":
            args["record_session"] = {"false": False, "true": True, "none": None}[case]
        try:
            result = await client.browser(**args)
        finally:
            await client.async_http_client.close()
        assert result.id == "test-session"
        assert len(requests) == 1
        assert requests[0].method == "POST"
        assert requests[0].url.path == "/v2/browser"
        payload = json.loads(requests[0].content)
        expected = {"ttl": 600, "activityTtl": 120, "streamWebView": False,
                    "profile": {"name": "test-profile", "saveChanges": False}}
        if case in ("false", "true"):
            expected["recordSession"] = case == "true"
            assert payload["recordSession"] is expected["recordSession"]
        else:
            assert "recordSession" not in payload
        assert payload.pop("origin").startswith("python-sdk@")
        assert payload == expected

    asyncio.run(run())
```

</details>

soroush5:
Thanks for checking the async path so thoroughly. Good call, I added your matrix as test_browser_record_session_async.py with explicit kwargs per case to keep the type checker quiet. All 7 tests in both files pass.

soroush5:
Closing as stale, no activity for a while. Happy to rebase and reopen if there is interest. Thanks for your time.

PR #4676, opened 2026-09-17
fix(api): populate branding fontWeights, which is declared but never assigned

sandiiarov:
### Problem

`branding.typography.fontWeights` is declared in the public type and documented with an example value, but nothing ever assigns it. It comes back `undefined` for every page.

Across the whole API, `fontWeights` appears exactly once — the declaration:

```
$ grep -rn "fontWeights" apps/api/src/
apps/api/src/types/branding.ts:50:    fontWeights?: {
```

The weight is already measured. `bs/elements.ts:302` reads it for every sampled element, and it survives into the snapshot type at `lib/branding/types.ts:132`. But the page-level pass never asks for it: `getTypography()` in `branding-script/brand-utils.ts` calls `getComputedStyleCached(h1)` for `fontSize` and stops there. So the object built at `lib/branding/processor.ts:287` ships `fontFamilies`, `fontStacks` and `fontSizes`, and there is no weight to put in `fontWeights`.

The practical effect is that a consumer building a brand kit can read the headline font and its size, but not whether the headline is 400 or 700 — and the field being present in the docs makes it look like a supported feature that happens to be null on their site.

### Fix

Ask the page for the weight, then carry it the same two hops the size already travels.

- `branding-script/brand-utils.ts` — add `weights` to the `Typography` interface and return it from `getTypography()`, using the same three elements the sizes already use.
- `lib/branding/types.ts` — add `weights` to the script-return shape.
- `lib/branding/processor.ts` — pass it through as `fontWeights`.
- `types/branding.ts` — add `h1`, `h2` and `body` to the declared `fontWeights`, keyed the same way `fontSizes` above it already is.

`weights` is **optional** on `BrandingScriptReturn`. The branding script ships inside fire-engine, which deploys separately, so an API running this code can still receive payloads from a fire-engine built before weights were collected. On those, `fontWeights` is simply absent, exactly as it is today.

The existing named-palette keys (`light`, `regular`, `medium`, `bold`) are left in place and untouched — this only adds the per-element keys alongside them, so the change is additive and non-breaking.

### Notes

The declared shape was a named palette, which cannot answer "what weight is the h1" even when fully populated. Keying the new values like `fontSizes` keeps the two parallel: `fontSizes.h1` and `fontWeights.h1` now describe the same element.

### Testing

Typechecks against the repo's `tsconfig` (`strict: false`, `strictNullChecks: true`, `exactOptionalPropertyTypes` unset), so passing `undefined` to the optional `fontWeights` is valid.

I could not run the branding script end to end locally, since it executes inside fire-engine. The three elements and fallback style follow the `sizes` block directly above, so the runtime behaviour should match. Happy to adjust the fallbacks (`700` / `700` / `400`) if you'd prefer different defaults, or to drop them and emit `null` instead.

Found while benchmarking the `branding` format against a set of real sites: weights were the single largest gap in the output, and the cause turned out to be one missing assignment rather than a measurement problem.

---
## Summary by cubic
Populates `branding.typography.fontWeights` so consumers can read the measured weight of the `h1`, `h2`, and `body` elements. Previously the field was declared but always `undefined`.

- Reads computed `fontWeight` from the same elements used for `fontSizes`, using `parseFloat` so fractional variable-font weights like `412.5` aren't truncated, with fallback values (`700`/`700`/`400`).
- Adds the new `h1`, `h2`, and `body` keys to the public `fontWeights` shape, leaving the existing named-palette keys untouched.
- Keeps `weights` optional on the script return so older fire-engine payloads without weights still work.

<sup>Written for commit 0d0c4edead61336b208bdd87b5b5acb21395d5ff. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecrawl/firecrawl/pull/4676?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


sandiiarov:
Good catch — confirmed and fixed in 0d0c4ed.

I checked it in a browser rather than reasoning from the spec, since `font-weight` accepting a `<number>` doesn't guarantee the computed value stays fractional:

```
font-weight: 412.5   → computed "412.5"  → parseInt 412  parseFloat 412.5
font-weight: 377.25  → computed "377.25" → parseInt 377  parseFloat 377.25
normal / bold        → computed "400" / "700"  → identical either way
```

So it's real for variable fonts, and keyword values are unaffected. Switched to `parseFloat`. The `|| fallback` guard is unchanged: a computed `font-weight` is never `0`, since the valid range is 1–1000.

One deliberate omission. `branding-script/elements.ts:302` has the same `parseInt` for the per-element snapshot weight, but I left it alone: `snapshots[].typography.weight` has no consumer anywhere in `lib/branding/`, so the truncation never reaches output. It's the same drop this PR is about, and it seemed better to keep the change focused than to fix a value nothing reads. Happy to include it if you'd rather have both in one go.

PR #4716, opened 2026-09-21
feat(api): refuse free-plan teams on paid-plan-only Exchange capabilities

mrod9997:
## Problem and behavior
Benzinga Schedule C.4 restricts full-text news, WIIM and analyst ratings to paid requests. Free teams can hold signup credits and accept provider terms, so credit reservation alone does not establish paid-plan eligibility.

Read Exchange's optional `paidPlanOnlyCapabilities` metadata after checking terms acceptance. Requests for those capabilities require a verified Autumn multiplier at or above the Hobby floor. Free plans receive 403 `paid_plan_required`; an unverifiable plan receives 503 `plan_verification_unavailable`, before quoting or execution. Internal teams with `bypassCreditChecks` retain their bypass. Older Exchange responses without the metadata remain compatible.

Autumn's strict entitlement reader shares successful cached entity reads with the rate limiter, while preserving the rate limiter's existing fail-open behavior. Missing organization IDs are rejected before consulting cached entitlements. A cold cache can require an Autumn request.

Counterpart: https://github.com/firecrawl/exchange/pull/623. Deploy Exchange metadata before this API gate.

## Validation
- 174 tests pass across Autumn service, Alexandria access, and Alexandria retrieval.
- Covers free/paid plans, missing entities, empty balances, billing errors, preview teams, missing organizations, bypasses, and terms-first ordering.
- Regression test reproduced a cached paid entitlement admitting a null organization before the fix; now passes.
- Refusal tests assert only the requirements request occurred, with no quote or execution.

Full package typechecking was not rerun in this review; the original implementation reported local native `@mendable/firecrawl-rs` build errors. CI remains the source for full validation.


mrod9997:
Both findings were valid and are addressed in the latest commit.

**Fail-open on unknown plan (P2).** Confirmed: `getRateLimitMultiplier` answers the 2500 fallback for a null org and on any Autumn error, so the gate admitted exactly the requests it could not judge. Fixed at the source rather than in the caller: the entity read behind `getEntityLimits` is now `readEntityLimits`, which reports what happened (`unconfigured`, `no_org`, `error`, `known`) instead of choosing a fallback. `getEntityLimits` keeps its fail-open throttling behaviour unchanged, and the new `getKnownRateLimitMultiplier` returns the entitled multiplier only when Autumn actually answered (1 for a 404 or absent balance) and `null` otherwise. The gate refuses `null` with 503 `plan_verification_unavailable` and a below-hobby multiplier with 403 `paid_plan_required`. That also closes the zero-credit path noted in `retrieve.ts`: a gated call with no org is refused in `authorizeProviders` before the quote. The comment no longer claims the read is free.

**Ordering assertion (P3).** The free-team test now asserts exactly one requirements read and nothing else after the refusal, and a new case covers the `null` answer.

Autumn service tests cover the strict reader across a live entity, a 404, an empty balance, a 500, no org and a preview team, beside the unchanged fail-open reads. `access`, `retrieve` and both Autumn suites pass (175 tests).

